PR #45, opened 2015-02-19
Fix duplicate dumpapp endpoints for a single process

### LINKED ISSUE #37: dumpapp "Multiple stetho-enabled processes"
Hi Guys,

Here is another issue. When I'm trying to execute 
```
./dumpapp prefs print
```
I see:
```
Multiple stetho-enabled processes available:
	com.mypackage
	com.mypackage
Use -p <process> to select one
```

It just display 2 equals packages, so when I try to specify package with -p com.mypackage it doesn't work.

What I did is open adb shell, and executed:

```
shell@klte:/ $ ps | grep com.mypackage
u0_a216   10774 259   965576 54048 ffffffff 00000000 S com.mypackage
```

So, I'm not really sure what's wrong.

P.S. com.mypackage is just replacement for my real package :)

Please let me know if you will need more details from me. 
I really want to use this project, it should save so lot's of time :)

jasta:
stetho_open.py naively assumes any entry matching @stetho_* represents a
bound server socket, however this is not true.  Connected client sockets
will also appear with different flags and in a different state.  Filter
based on this criteria to only identify server sockets.

Closes #37

### DIFF
--- a/scripts/stetho_open.py
+++ b/scripts/stetho_open.py
@@ -42,12 +42,17 @@ def _find_only_stetho_socket(device):
     process_names = []
     for line in adb.sock.makefile():
       row = line.rstrip().split(' ')
-      if len(row) > 7:
-        socket_name = row[7]
-        if socket_name.startswith('@stetho_'):
-          last_stetho_socket_name = socket_name[1:]
-          process_names.append(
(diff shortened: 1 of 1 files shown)


jasta:
@longinoa merged without review as this is an important bug fix.

Issue #43, opened 2015-02-19
Network inspection not working with OkUrlFactory


bryanstern:
If you are using OkHttp + Volley, you are likely using `OkUrlFactory` via a class like the one below. Unfortunately OkHttp's network interceptors do not support the OkUrlFactory.
https://github.com/square/okhttp/wiki/Interceptors#availability

```java
import com.android.volley.toolbox.HurlStack;
import com.squareup.okhttp.OkHttpClient;
import com.squareup.okhttp.OkUrlFactory;

import java.io.IOException;
import java.net.HttpURLConnection;
import java.net.URL;

/**
 * An {@link com.android.volley.toolbox.HttpStack HttpStack} implementation which
 * uses OkHttp as its transport.
 */
public class OkHttpStack extends HurlStack {
    private final OkUrlFactory mFactory;

    public OkHttpStack() {
        this(new OkHttpClient());
    }

    public OkHttpStack(OkHttpClient client) {
        if (client == null) {
            throw new NullPointerException("Client must not be null.");
        }
        mFactory = new OkUrlFactory(client);
    }

    @Override protected HttpURLConnection createConnection(URL url) throws IOException {
        return mFactory.open(url);
    }
}
```
Any ideas on an OkHttp UrlConnection friendly way of integrating Stetho?

jasta:
@bryanstern you can integrate `HttpURLConnection` using `StethoURLConnectionManager`.  If you depend on `okhttp-urlconnection`, you can likely make this work pretty easily through subclassing `HurlStack`.  Note that you could probably more easily implement a better `OkHttpStack` than the one Google provided in Volley.  Basically all you'd need is:
```java
public HttpResponse performRequest(...) {
  Response okHttpResponse = mOkHttpClient.newCall(convertToCall(request, params)).execute();
  return convertToApacheResponse(okHttpResponse);
}
```

bryanstern:
@jasta Thanks for suggestion. I ended up just implementing Volley's `HttpStack` interface and performing the request directly with OkHttp, doing away with `okhttp-urlconnection`. Once I feel more confident in the code, I'll post that gist up for those interested.

bryanstern:
Here is the gist, just set the network interceptor on the OkHttpClient you pass in. https://gist.github.com/bryanstern/4e8f1cb5a8e14c202750

jasta:
@bryanstern closing now that a solution exists.  I think if we're going to merge something in it's most likely to be from #51 or something like it which more generically supports Volley.

PR #51, opened 2015-02-20
Added Volley stetho network

briangriffey:


jasta:
I think there's a much easier way to do this by subclassing the two HttpStack implementations or maybe even easier by just creating a StethoWrappingHttpStack.

Also, prefer using Volley through Maven coordinates.  We don't currently bundle any jar files.

briangriffey:
Extending HurlStack might be possible. That private static entityFromConnection might prove painful when trying to read the inputstream.

jasta:
As I think about this more, I think wrapping HttpStack is more flexible and simpler.  This can be done by replacing the entity like:
```java
  public HttpResponse performRequest(...) {
    mNetworkEventReporter.requestWillBeSent(new VolleyInspectorRequest(request));
    response = mDelegate.performRequest(request);
    mNetworkEventReporter.receivedResponseHeaders(...);
    InputStream intercepted = mNetworkEventReporter.interpretResponseStream(...);
    response.setEntity(new InputStreamEntity(intercepted));
    return response;
  }
```
This should make it possible to do:
```java
new StethoInterceptingHttpStack(new HurlStack(...));
```

jasta:
Just saw #43 which might be of interest to you as well.  This is a slightly different take on the solution if you use OkHttp.

briangriffey:
Yea, I just think this is the only connection agnostic method to integrating with volley. Still allows you to use OkHttp or UrlConnection or HttpClient under the hood. This sits on the layer right above that.

jasta:
@briangriffey agreed, I think the `HttpStack` wrapper works nicely for that reason too.

settings.gradle
```diff
@@ -1,4 +1,5 @@
 include ':stetho'
```
mannyman1115: Leave a line note

stetho-volley/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,8 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
```
mannyman1115: Leave a line note

jasta:
Any update?  I'd really love to have good Volley support for Stetho...

briangriffey:
Sorry, end of the quarter is happening. I'll circle back to this beginning of April.

gfx:
👀

milek:
Any update on that pull request?

condemoreloer:
brian@briangriffey.com
Ledo un gouache.

vaurdan:
Any news on this one? It is possible to use Stetho with Volley?

jasta:
@vaurdan yes it's always been possible to use Stetho with Volley it just requires you jump through some hoops.  The easiest way is to use OkHttp with Volley and install this implementation into Volley: https://gist.github.com/bryanstern/4e8f1cb5a8e14c202750

This works great actually, it's just not as discoverable/convenient as we'd like.  It also doesn't work if you use Volley with HttpURLConnection :\

vaurdan:
@jasta Oh! Thanks for the tip, I've now Stetho working with Volley using the OkHttp! :+1:

YusukeIwaki:
With stetho-volley, NullPointerException occurs when the response code is 204.

Applying the patch below is required.
https://android.googlesource.com/platform/frameworks/volley/+/e8f5353b17aea6659dc3aa13223317b484d01483%5E%21/src/com/android/volley/toolbox/BasicNetwork.java

shibenli:
Thanks to (http://www.littlerobots.nl/blog/stetho-for-android-debug-builds-only/ )[http://www.littlerobots.nl/blog/stetho-for-android-debug-builds-only/ ].I've now Stetho working with Volley using the OkHttp in my **debug** build, but using Volley alone  in **release**! :+1:

jasta:
Ancient pull request and really isn't everyone just using OkHttp now? :)

PR #53, opened 2015-02-20
Break the bindToSocket retry loop on success

### LINKED ISSUE #49: Error: Could not bind to socket
Upon starting an app, stetho fails to setup a server here:

https://github.com/facebook/stetho/blob/69a9ec51b93dc7eed122900ca5bbb823e6aa4ab9/stetho/src/main/java/com/facebook/stetho/server/LocalSocketHttpServer.java#L287

Not sure what I am doing wrong and whether there is anything particular needed to be configured on my test device (Nexus 5/6)?

jasta:
This caused all invocations to print a stack trace indicating a serious
sounding error when in fact everything was working fine.  Added a log
statement to clarify that everything is fine :)

Closes #49


longinoa:
LGTM. I would probably prefer this to be a static method that returns a `LocalServerSocket` but can go either way.

PR #61, opened 2015-02-21
Avoid caching $HOME/.gradle in Travis-CI

jasta:
This directory changes on every invocation due to a lck file written
into it.  According to http://stackoverflow.com/a/27365925, it might be
better to just take the hit on downloading gradle's distribution each
time.  But we'll keep the Maven dependencies still since those are much
larger.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,4 +14,3 @@ sudo: false
 cache:
   directories:
     - $HOME/.m2
-    - $HOME/.gradle


jasta:
This is a bad idea.  The `~/.m2` directory is apparently not used by Gradle's dependencies, so it's somewhere in `~/.gradle`.  I think it's best to do the naive thing and cache `~/.gradle` as we do now and check back with Travis-CI for a better fix later.

PR #63, opened 2015-02-21
Stop using Travis-CI cache

jasta:
This just doesn't optimize things with Gradle due to the lock file being
changed in ~/.gradle/caches on each build.  Disable for now; hopefully
Travis-CI will fix this.


jasta:
Confirmed, build is actually faster this way.

PR #64, opened 2015-02-22
Add SharedPreferences data to Resources tab

### LINKED ISSUE #50: Implement SharedPreferences Inspection
SharedPreferences could be inspected via the Local Storage area in the Chrome developer tools.

jasta:
inspector-sharedprefs

Closes #50

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/domstorage/DOMStoragePeerManager.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/domstorage/DOMStoragePeerManager.java
@@ -2,14 +2,19 @@ package com.facebook.stetho.inspector.domstorage;
 
 import android.content.Context;
 import android.content.SharedPreferences;
+import com.facebook.stetho.inspector.console.CLog;
 import com.facebook.stetho.inspector.helper.ChromePeerManager;
 import com.facebook.stetho.inspector.helper.PeerRegistrationListener;
 import com.facebook.stetho.inspector.helper.PeersRegisteredListener;
+import com.facebook.stetho.inspector.protocol.module.Console;
 import com.facebook.stetho.inspector.protocol.module.DOMStorage;
 
 import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 public class DOMStoragePeerManager extends ChromePeerManager {
   private final Context mContext;
@@ -17,6 +22,37 @@ public class DOMStoragePeerManager extends ChromePeerManager {
   public DOMStoragePeerManager(Context context) {
     mContext = context;
     setListener(mPeerListener);
+  }
+
+  public void signalItemRemoved(DOMStorage.StorageId storageId, String key) {
+    DOMStorage.DomStorageItemRemovedParams params =
+        new DOMStorage.DomStorageItemRemovedParams();
+    params.storageId = storageId;
+    params.key = key;
+    sendNotificationToPeers("DOMStorage.domStorageItemRemoved", params);
+  }
+
+  public void signalItemAdded(DOMStorage.StorageId storageId, String key, String value) {
+    DOMStorage.DomStorageItemAddedParams params =
+        new DOMStorage.DomStorageItemAddedParams();
+    params.storageId = storageId;
+    params.key = key;
+    params.newValue = value;
+    sendNotificationToPeers("DOMStorage.domStorageItemAdded", params);
+  }
+
+  public void signalItemUpdated(
+      DOMStorage.StorageId storageId,
+      String key,
+      String oldValue,
+      String newValue) {
+    DOMStorage.DomStorageItemUpdatedParams params =
+        new DOMStorage.DomStorageItemUpdatedParams();
+    params.storageId = storageId;
+    params.key = key;
+    params.oldValue = oldValue;
+    params.newValue = newValue;
+    sendNotificationToPeers("DOMStorage.domStorageItemUpdated", params);
   }
 
   private final PeerRegistrationListener mPeerListener = new PeersRegisteredListener() {
@@ -54,11 +90,20 @@ public class DOMStoragePeerManager extends ChromePeerManager {
     private final SharedPreferences mPrefs;
     private final DOMStorage.StorageId mStorageId;
 
+    /**
+     * Maintains a copy of the prefs data structure so that we can invoke
+     * {@code DOMStorage.domStorageItemUpdated}.  This method requires that we know the old
+     * value to perform updates.  Using {@code domStorageItemRemoved}/{@code Added} causes a UI
+     * glitch where the item is moved to the end of the list, unfortunately.
+     */
+    private final Map<String, Object> mCopy;
+
     public DevToolsSharedPreferencesListener(SharedPreferences prefs, String tag) {
       mPrefs = prefs;
       mStorageId = new DOMStorage.StorageId();
       mStorageId.securityOrigin = tag;
       mStorageId.isLocalStorage = true;
+      mCopy = prefsCopy(prefs.getAll());
     }
 
     public void unregister() {
@@ -67,21 +112,51 @@ public class DOMStoragePeerManager extends ChromePeerManager {
 
     @Override
     public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, String key) {
-      DOMStorage.DomStorageItemRemovedParams removedParams =
-          new DOMStorage.DomStorageItemRemovedParams();
-      removedParams.storageId = mStorageId;
-      removedParams.key = key;
-      sendNotificationToPeers("DOMStorage.domStorageItemRemoved", removedParams);
-
       Map<String, ?> entries = sharedPreferences.getAll();
-      if (entries.containsKey(key)) {
-        DOMStorage.DomStorageItemAddedParams addedParams =
-            new DOMStorage.DomStorageItemAddedParams();
-        addedParams.storageId = mStorageId;
-        addedParams.key = key;
-        addedParams.newValue = SharedPreferencesHelper.valueToString(entries.get(key));
-        sendNotificationToPeers("DOMStorage.domStorageItemAdded", addedParams);
+      boolean existedBefore = mCopy.containsKey(key);
+      boolean existsNow = entries.containsKey(key);
+      Object newValue = existsNow ? entries.get(key) : null;
+      if (existedBefore && existsNow) {
+        signalItemUpdated(
+            mStorageId,
+            key,
+            SharedPreferencesHelper.valueToString(mCopy.get(key)),
+            SharedPreferencesHelper.valueToString(newValue));
+        mCopy.put(key, newValue);
+      } else if (existedBefore) {
+        signalItemRemoved(mStorageId, key);
+        mCopy.remove(key);
+      } else if (existsNow) {
+        signalItemAdded(
+            mStorageId,
+            key,
+            SharedPreferencesHelper.valueToString(newValue));
+        mCopy.put(key, newValue);
+      } else {
+        throw new RuntimeException("Prefs change detected for unknown key=" + key);
       }
     }
+  }
+
+  private static Map<String, Object> prefsCopy(Map<String, ?> src) {
+    HashMap<String, Object> dst = new HashMap<String, Object>(src.size());
+    for (Map.Entry<String, ?> entry : src.entrySet()) {
+      String key = entry.getKey();
+      Object value = entry.getValue();
+      if (value instanceof Set) {
+        dst.put(key, shallowCopy((Set<String>)value));
+      } else {
+        dst.put(key, value);
+      }
+    }
+    return dst;
+  }
+
+  private static <T> Set<T> shallowCopy(Set<T> src) {
+    HashSet<T> dst = new HashSet<T>();
+    for (T item : src) {
+      dst.add(item);
+    }
+    return dst;
   }
 }
--- a/stetho/src/main/java/com/facebook/stetho/inspector/domstorage/DOMStoragePeerManager.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/domstorage/DOMStoragePeerManager.java
@@ -1,0 +1,87 @@
+package com.facebook.stetho.inspector.domstorage;
+
+import android.content.Context;
+import android.content.SharedPreferences;
+import com.facebook.stetho.inspector.helper.ChromePeerManager;
+import com.facebook.stetho.inspector.helper.PeerRegistrationListener;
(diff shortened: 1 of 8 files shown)


hzsweers:
This is awesome!

stetho/src/main/java/com/facebook/stetho/inspector/domstorage/DOMStoragePeerManager.java
```diff
@@ -0,0 +1,86 @@
...
+    @Override
+    protected synchronized void onFirstPeerRegistered() {
+      // TODO: We list the tags in Page.getResourceTree as well and those are the real fixed
+      // tags that will be observed by the peer.  We can fix this by making the page frames
+      // dynamically update in response to DOMStorage events.  This would also allow us to
+      // add new SharedPreferences tags as we observe them being created by way of
+      // android.os.FileObserver.
+      List<String> tags = SharedPreferencesHelper.getSharedPreferenceTags(mContext);
+      for (String tag : tags) {
+        SharedPreferences prefs = mContext.getSharedPreferences(tag, 0 /* mode */);
```
longinoa: `Context.MODE_PRIVATE`

stetho/src/main/java/com/facebook/stetho/inspector/domstorage/DOMStoragePeerManager.java
```diff
@@ -0,0 +1,86 @@
...
+    setListener(mPeerListener);
+  }
+
+  private final PeerRegistrationListener mPeerListener = new PeersRegisteredListener() {
+    private final List<DevToolsSharedPreferencesListener> mPrefsListeners =
+        new ArrayList<DevToolsSharedPreferencesListener>();
+
+    @Override
+    protected synchronized void onFirstPeerRegistered() {
+      // TODO: We list the tags in Page.getResourceTree as well and those are the real fixed
```
longinoa: Did we ever figure out how to reference an Issue from code?

stetho/src/main/java/com/facebook/stetho/inspector/domstorage/DOMStoragePeerManager.java
jasta: Ack'd, will update

stetho/src/main/java/com/facebook/stetho/inspector/domstorage/DOMStoragePeerManager.java
jasta: No, but a URL to github seems appropriate.  There is no task filed for this yet, and I actually don't like the idea of enforcing Facebook's internal policy of all TODOs are matched with a task.  I find that incredibly annoying, personally...

stetho/src/main/java/com/facebook/stetho/inspector/domstorage/DOMStoragePeerManager.java
longinoa: It seems incorrect to assume not multi-process.

stetho/src/main/java/com/facebook/stetho/inspector/domstorage/DOMStoragePeerManager.java
```diff
@@ -0,0 +1,86 @@
...
+        mPrefsListeners.add(listener);
+      }
+    }
+
+    @Override
+    protected synchronized void onLastPeerUnregistered() {
+      for (DevToolsSharedPreferencesListener prefsListener : mPrefsListeners) {
+        prefsListener.unregister();
+      }
+    }
```
longinoa: `mPrefsListeners.clear()`

stetho/src/main/java/com/facebook/stetho/inspector/domstorage/DOMStoragePeerManager.java
jasta: Multiprocess support is incomplete/poorly implemented.  It won't help us here.  We'd have to poll for updates to do it properly.  Not worth it, IMO, especially since there is a "refresh" button in the UI.

stetho/src/main/java/com/facebook/stetho/inspector/domstorage/DOMStoragePeerManager.java
```diff
@@ -0,0 +1,86 @@
...
+      mPrefs.unregisterOnSharedPreferenceChangeListener(this);
+    }
+
+    @Override
+    public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, String key) {
+      DOMStorage.DomStorageItemRemovedParams removedParams =
+          new DOMStorage.DomStorageItemRemovedParams();
+      removedParams.storageId = mStorageId;
+      removedParams.key = key;
+      sendNotificationToPeers("DOMStorage.domStorageItemRemoved", removedParams);
```
longinoa: Did you test what happens when you add a completely new item? calling `remove` on an item that didnt exist could be problematic

stetho/src/main/java/com/facebook/stetho/inspector/domstorage/DOMStoragePeerManager.java
jasta: Oops :)

stetho/src/main/java/com/facebook/stetho/inspector/domstorage/SharedPreferencesHelper.java
```diff
@@ -0,0 +1,39 @@
...
+  private SharedPreferencesHelper() {
+  }
+
+  public static List<String> getSharedPreferenceTags(Context context) {
+    ArrayList<String> tags = new ArrayList<String>();
+
+    String rootPath = context.getApplicationInfo().dataDir + "/shared_prefs";
+    File root = new File(rootPath);
+    if (root.exists()) {
+      for (File file : root.listFiles()) {
```
longinoa: :-1: booo that this is the interface we have to use.

stetho/src/main/java/com/facebook/stetho/inspector/domstorage/DOMStoragePeerManager.java
jasta: I did.  It seems to work fine.  I tried using `domStorageItemUpdated` and that's the one that freaked out.  If you don't match the existing value correctly it ignores the update.

stetho/src/main/java/com/facebook/stetho/inspector/domstorage/SharedPreferencesHelper.java
jasta: In the scheme of things we're doing in Stetho, I'm not so sure this is our riskiest assumption.... :)

jasta:
Responded to feedback.

stetho/src/main/java/com/facebook/stetho/inspector/domstorage/DOMStoragePeerManager.java
longinoa: was more a question of "is this possible" than "do this"

stetho/src/main/java/com/facebook/stetho/inspector/domstorage/DOMStoragePeerManager.java
longinoa: :+1: yeah I saw that in the protocol, wanted to make sure we didnt need to cache all values we returned

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOMStorage.java
```diff
@@ -2,20 +2,185 @@
...
+      SharedPreferences prefs = mContext.getSharedPreferences(
+          storage.securityOrigin,
+          Context.MODE_PRIVATE);
+      Object exitingValue = prefs.getAll().get(key);
+      SharedPreferences.Editor editor = prefs.edit();
+      if (!tryAssignByType(editor, key, value, exitingValue)) {
+        editor.putString(key, value);
+      }
+      editor.apply();
+    }
```
longinoa: `else {CLog(...);}`

jasta: Actually no, this is about making sure the data shows up only in the "Local Storage" group not "Session Storage".  I wanted to restrict the data so that it didn't appear to be duplicated.  Maybe we will use "Session Storage" for something else later.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Page.java
```diff
@@ -68,20 +69,55 @@ private void sendWelcomeMessage(JsonRpcPeer peer) {
...
+    // The DOMStorage module expects one key/value store per "security origin" which has a 1:1
+    // relationship with resource tree frames.
+    List<String> prefsTags = SharedPreferencesHelper.getSharedPreferenceTags(mContext);
+    Iterator<String> prefsTagsIter = prefsTags.iterator();
+
+    FrameResourceTree tree = createSimpleFrameResourceTree(
+        "1",
+        null /* parentId */,
+        "Stetho",
+        prefsTagsIter.hasNext() ? prefsTagsIter.next() : "");
```
longinoa: this seems a bit strange looking at it. Why is the first "tag" the security origin of the root of the tree rather than it being stetho or something similar?

jasta: All frames require a security origin.  All security origins show up as subgroups in "Local Storage".  Therefore, the outer frame's origin is basically the first in the list.  If I do the appropriately logical thing for Andriod, a dummy row will appear that doesn't refer to SharedPreferences.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOMStorage.java
```diff
@@ -2,20 +2,185 @@
...
+  }
+
+  private static boolean tryAssignByType(
+      SharedPreferences.Editor editor,
+      String key,
+      String value,
+      @Nullable Object existingValue) {
+    try {
+      if (existingValue instanceof Integer || existingValue instanceof Long) {
+        editor.putLong(key, Long.parseLong(value));
```
longinoa: why are `Integer` and `Long` treated the same?

jasta: Oh, good call.  I assumed `Long` was a strict superset of `Integer` for `SharedPreferences` but this is not the case.  This could cause crashes in caller code.

I'll correct it.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOMStorage.java
```diff
@@ -2,20 +2,185 @@
...
+    String value = params.getString("value");
+
+    if (storage.isLocalStorage) {
+      SharedPreferences prefs = mContext.getSharedPreferences(
+          storage.securityOrigin,
+          Context.MODE_PRIVATE);
+      Object exitingValue = prefs.getAll().get(key);
+      SharedPreferences.Editor editor = prefs.edit();
+      if (!tryAssignByType(editor, key, value, exitingValue)) {
+        editor.putString(key, value);
```
longinoa: This seems pretty fragile, and a good way to break the app. 

I dont have a great way to test this, but when does `setDOMStorageItem` get called on the devtools side?  If it is when you leave the field this has the ability to be pretty harmful to the hosting app.There is no validation, and no telling what the app expects this to be.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOMStorage.java
```diff
@@ -2,20 +2,185 @@
...
+
+    if (storage.isLocalStorage) {
+      SharedPreferences prefs = mContext.getSharedPreferences(
+          storage.securityOrigin,
+          Context.MODE_PRIVATE);
+      prefs.edit().remove(key).apply();
+    }
+  }
+
+  private static boolean tryAssignByType(
```
longinoa: No support for `Set<String>`

longinoa:
The set semantics in this diff seem fragile. Would like to see more on the flow of how that works.

jasta:
Addressed the type mismatch issue:
type-mismatch

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOMStorage.java
```diff
@@ -2,20 +2,197 @@
...
+        StorageId.class);
+    String key = params.getString("key");
+    String value = params.getString("value");
+
+    if (storage.isLocalStorage) {
+      SharedPreferences prefs = mContext.getSharedPreferences(
+          storage.securityOrigin,
+          Context.MODE_PRIVATE);
+      Object existingValue = prefs.getAll().get(key);
+      SharedPreferences.Editor editor = prefs.edit();
```
longinoa: this should be in the `else` block. Why would we call `apply` if we know we arent going to change it.

stetho/src/main/java/com/facebook/stetho/inspector/domstorage/SharedPreferencesHelper.java
```diff
@@ -0,0 +1,92 @@
...
+        throw new IllegalArgumentException(e);
+      }
+    } else {
+      throw new IllegalArgumentException(
+          "Unsupported type: " + existingValue.getClass().getName());
+    }
+  }
+
+  private static Boolean parseBoolean(String s) throws IllegalArgumentException {
+    if ("1".equals(s) || "true".equalsIgnoreCase(s)) {
```
longinoa: I would prefer to stick with `Boolean.parseBoolean` this way there is no confusion around the intent of the value.

longinoa:
You should update the UI on success (IE 1 -> true) and reflect failure ie trying to change true -> fizbuzz should reset the value to true.

stetho/src/main/java/com/facebook/stetho/inspector/domstorage/DOMStoragePeerManager.java
```diff
@@ -0,0 +1,166 @@
...
+        signalItemRemoved(mStorageId, key);
+        mCopy.remove(key);
+      } else if (existsNow) {
+        signalItemAdded(
+            mStorageId,
+            key,
+            SharedPreferencesHelper.valueToString(newValue));
+        mCopy.put(key, newValue);
+      } else {
+        CLog.writeToConsole(
```
longinoa: This should be a throw, or a statement to file a bug. If both `existedBefore` and `existsNow` are `false` that has to imply Android did something it shouldn't do, or more likely our map is out of date.

jasta: Ack'd, will throw.

longinoa:
:+1: Can you squash your last 3 commits

PR #66, opened 2015-02-22
Minor fixes to bindToSocket

jasta:
Make bindToSocket static, mark it non-null and instead throw on error,
and clarify the "retry" semantic to actually indicate the number of
retries not number of tries.

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/server/LocalSocketHttpServer.java
+++ b/stetho/src/main/java/com/facebook/stetho/server/LocalSocketHttpServer.java
@@ -37,6 +37,7 @@ import java.io.InputStreamReader;
 import java.io.InterruptedIOException;
 import java.net.BindException;
 import java.net.SocketException;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import android.annotation.SuppressLint;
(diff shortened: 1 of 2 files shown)


stetho/src/main/java/com/facebook/stetho/server/LocalSocketHttpServer.java
```diff
@@ -266,27 +264,36 @@ public void stop() {
...
+      }
+    } while (retries-- > 0);
+
+    throw firstException;
+  }
+
+  private static void sleepUninterruptibly(long millis) {
+    while (true) {
+      try {
+        Thread.sleep(millis);
```
longinoa: You should save state here. If you keep getting interrupted you will never terminate this loop.

stetho/src/main/java/com/facebook/stetho/server/LocalSocketHttpServer.java
```diff
@@ -266,27 +264,36 @@ public void stop() {
...
+    int retries = MAX_BIND_RETRIES;
+    IOException firstException = null;
+    do {
+      try {
+        if (LogUtil.isLoggable(Log.DEBUG)) {
+          LogUtil.d("Trying to bind to @" + address);
+        }
+        return new LocalServerSocket(address);
+      } catch (BindException be) {
+        LogUtil.w(be, "Binding error, sleep 1 second ...");
```
longinoa: prefer to use the constant `TIME_BETWEEN_RETRIES_MS`

longinoa:
Can you add a description as well.

stetho/src/main/java/com/facebook/stetho/server/LocalSocketHttpServer.java
jasta: Ack'd, wasn't thinking...

jasta:
Updated.  :+1: or :-1: ?

longinoa:
:-1: you still sleep for `millis` rather than `remaining`

longinoa:
:+1:

PR #69, opened 2015-02-23
Use raw process name for UNIX abstract socket name

jasta:
Previously we formatted the process name using underscores due to
primitive tooling that has since been replaced.  We can now use the real
process name which fixes an issue with ./scripts/dumpapp output when
there are multiple targets that include derivative processes
(com.mypackage:myprocess).  Previously this would show up as
com.mypackage.myprocess.

### DIFF
--- a/scripts/stetho_open.py
+++ b/scripts/stetho_open.py
@@ -85,13 +85,12 @@ def _parse_process_from_stetho_socket(socket_name):
   if len(parts) < 2 or parts[0] != '@stetho':
     raise Exception('Unexpected Stetho socket formatting: %s' % (socket_name))
   if parts[-2:] == [ 'devtools', 'remote' ]:
-    return '.'.join(parts[1:-2])
+    return '_'.join(parts[1:-2])
   else:
-    return '.'.join(parts[1:])
+    return '_'.join(parts[1:])
 
 def _format_process_as_stetho_socket(process):
-  filtered = process.replace('.', '_').replace(':', '_')
-  return 'stetho_%s_devtools_remote' % (filtered)
+  return 'stetho_%s_devtools_remote' % (process)
 
 class AdbSmartSocketClient(object):
   """Implements the smartsockets system defined by:
--- a/scripts/stetho_open.py
+++ b/scripts/stetho_open.py
@@ -15,6 +15,7 @@
 
 import socket
 import struct
+import re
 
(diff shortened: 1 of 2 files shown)


stetho/src/main/java/com/facebook/stetho/server/LocalSocketHttpServer.java
```diff
@@ -190,32 +188,28 @@ private static String getProcessName() throws IOException {
...
-    int i = 0;
-    OUTER:
-    for (; i < count; i++) {
-      switch (processName[i]) {
-        case '\\':
-        case '.':
-        case ':':
-          processName[i] = '_';
-          break;
+  private static int findProcessNameEndIndex(byte[] cmdline, int count) {
```
longinoa: this seems like it could conflict with #67

jasta: It will.  Doesn't matter which one goes first, technically, but one will need to be rebased/updated.  Github should block automerging.

longinoa: not really, you are calling `tidyProcessName` before which does the semantic replacement of `processName` characters to `_` 

In this diff you assume that that has been done and you return the index of the terminating character.

jasta: That's what this diff is intending to do.  It no longer does replacements at all.  If it applies first, the one which uses regexps will be rebased and the regexps will be removed.  If this applies last, the regexp will be replaced with nothing...

I developed the diffs in such a way that they were both intentionally based off `master` so neither had to go first strictly.

scripts/stetho_open.py
```diff
@@ -85,13 +85,12 @@ def _parse_process_from_stetho_socket(socket_name):
   if len(parts) < 2 or parts[0] != '@stetho':
     raise Exception('Unexpected Stetho socket formatting: %s' % (socket_name))
   if parts[-2:] == [ 'devtools', 'remote' ]:
-    return '.'.join(parts[1:-2])
+    return '_'.join(parts[1:-2])
```
longinoa: This can now be reflected as a regex

jasta: Done

Issue #59, opened 2015-02-20
Installing 'com.facebook.stetho:stetho-okhttp:1.0.0' breaks okhttp


bbade:
When I included this jar in my project, okhttp would crash when used with what looked like a proguard related issue.

I  get the following error

Caused by: java.lang.NoSuchFieldError: No static field METHODS of type Ljava/util/Set; in class Lcom/squareup/okhttp/internal/http/HttpMethod; or its superclasses (declaration of 'com.squareup.okhttp.internal.http.HttpMethod' appears in /data/app/shiftgig.com.worknow-2/base.apk

i've tried whitelisting that file in proguard to no avail.

jasta:
Were you previously using okhttp 2.2.0 without problems?  This seems like just okhttp itself would fail, independent of `stetho-okhttp`.

bbade:
Yeah, as soon as I comment out that lib in my gradle file, it works again.

jasta:
Can you provide your proguard configuration?  I'm trying to repro locally and am so far unable to.  Not sure what's going on here or why adding Stetho would cause it.

eleventigers:
I would suggest excluding okhttp from stetho dependency, worked for me:

```groovy
compile ('com.facebook.stetho:stetho-okhttp:1.0.0') {
        exclude group: 'com.squareup.okhttp'
}
```

jasta:
Interesting.  The weird thing with the way this works is that I'm really trying to declare that I can use okhttp 2.2.0+ in that module but that I don't really want to depend on it (I want the caller to make the dependency).  Gradle doesn't seem to offer a way to do this that I can see.  Perhaps this is why Square is still using Maven :)

JakeWharton:
You shouldn't use wildcard dependencies for anything. Just declare a dep on 2.2.0. If someone depends on 2.3.0 every dependency resolver (Gradle, Ivy, Aether) knows what to do.

JakeWharton:
Or, optionally, use a `provided` dep which is exactly what you described.

jasta:
@JakeWharton awesome, thanks.

longinoa:
@bbade we just shipped 1.0.1 that has the fix for this can you retry now

jasta:
Err actually I'm not so sure that we fixed this issue. I referenced it only
because of @JakeWharton shaming our usage of wildcard deps :)

I was never able to repro the issue so I can't confirm that we fixed it...

PR #71, opened 2015-02-24
Avoid wildcard dependency in stetho-okhttp

jasta:
According to #59, it should be safe to just depend on 2.2.0 and ask the
dependency resolver to figure it out.


JakeWharton:
LGTM :+1: :shipit:

stetho-okhttp/build.gradle
```diff
@@ -26,7 +26,7 @@ dependencies {
     compile project(':stetho')
     compile 'com.google.code.findbugs:jsr305:2.0.1'
     //noinspection GradleDynamicVersion
```
JakeWharton: Can be removed

longinoa:
:+1:

Issue #44, opened 2015-02-19
Update project files for Java 1.7 syntax support


rickbrew:
I believe there's something we need to declare in gradle.properties and/or elsewhere. So let's figure that out.

jasta:
Diamond syntax FTW!

clemp6r:
You have to add the following in the android section of your build.gradle:

    compileOptions {
        sourceCompatibility JavaVersion.VERSION_1_7
        targetCompatibility JavaVersion.VERSION_1_7
    }

Issue #14, opened 2015-02-12
Implement Elements tab


jasta:
Implement Elements tab to support DOM / view hierarchy inspection

goncalossilva:
Yes, please!

jasta:
We're working on it: https://github.com/rickbrew/stetho/commit/b87a818d6e3b6bf1e68e0e4c8aa5e76b6eabb53e :)

felipecsl:
is this merged in already?

rickbrew:
Woops, nope it's still over in the elements branch.

jasta:
@felipecsl soon..... :)

felipecsl:
cant wait :watch:  :)

PR #73, opened 2015-02-24
Initial check-in for Elements tab

rickbrew:
This is a first stab at providing an Elements tab for Stetho.

The DOM class handles the JSON protocol and element <--> node ID mapping. An element is meant to be a real object in whatever DOM-type tree is being plugged in via a DOM.Provider implementation. We implement this in UIDOMProvider and our "DOM-type tree" is an Android application. We use the android.app.Application as the root of the DOM, its children are the android.app.Activity instances (initial list retrieved via a hack, then updated via Application.ActivityLifecycleCallbacks), then the android.view.Window and finally its android.view.View tree.

We use DOM.NodeDescriptor to describe an element in terms of a node (DOM.Node). In UIDOMProvider we use the element's class to figure out which descriptor(s) to use, but another DOM.Provider implementation could potentially choose another strategy.

For instance, an android.view.ViewGroup is an "element." ViewGroupNodeDescriptor knows how to describe it in DOM.Node terms (name, children), and how to hook into it (with ViewGroup.OnHierarchyChangedListener) in order to report updates (via DOM.childNodeInserted / DOM.childNodeRemoved). Also participating in this are ViewNodeDescriptor (technically anyway, since it's currently empty) and ObjectNodeDescriptor (mostly just knows how to shorten class names to remove all the "android.*.*" package name noise).

In the Elements tab you can mouse-over a node and it will cause the Android device to draw a red highlight over the corresponding View. This is ultimately handled in the WindowNodeDescriptor class because that's where the highlighting View is installed (and uninstalled).

FIrst step on #14 

screen shot 2015-02-24 at 1 27 13 pm

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/Stetho.java
+++ b/stetho/src/main/java/com/facebook/stetho/Stetho.java
@@ -5,6 +5,7 @@ import javax.annotation.Nullable;
 import java.io.IOException;
 import java.util.ArrayList;
 
+import android.app.Application;
 import android.content.Context;
 import android.content.SharedPreferences;
 import android.os.Build;
@@ -19,6 +20,7 @@ import com.facebook.stetho.dumpapp.StreamingDumpappHandler;
 import com.facebook.stetho.dumpapp.plugins.SharedPreferencesDumperPlugin;
 import com.facebook.stetho.inspector.ChromeDevtoolsServer;
 import com.facebook.stetho.inspector.ChromeDiscoveryHandler;
+import com.facebook.stetho.inspector.elements.android.AndroidDOMProviderFactory;
 import com.facebook.stetho.inspector.protocol.ChromeDevtoolsDomain;
 import com.facebook.stetho.inspector.protocol.module.CSS;
 import com.facebook.stetho.inspector.protocol.module.Console;
@@ -105,7 +107,7 @@ public class Stetho {
         modules.add(new Console());
         modules.add(new CSS());
         modules.add(new Debugger());
-        modules.add(new DOM());
+        modules.add(new DOM(new AndroidDOMProviderFactory((Application)context.getApplicationContext())));
         modules.add(new DOMStorage());
         modules.add(new HeapProfiler());
         modules.add(new Inspector());
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/DOMProvider.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/DOMProvider.java
@@ -1,0 +1,49 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+package com.facebook.stetho.inspector.elements;
+
+import javax.annotation.Nullable;
+
+public interface DOMProvider {
+  public void setListener(Listener listener);
+
+  public void dispose();
+
+  @Nullable
+  public Object getRootElement();
+
+  public NodeDescriptor getNodeDescriptor(Object element);
+
+  public void highlightElement(
+      Object element,
+      int contentColor,
+      int paddingColor,
+      int borderColor,
+      int marginColor);
+
+  public void hideHighlight();
+
+  public static interface Factory {
+    DOMProvider create();
+  }
+
+  public static interface Listener {
+    public void onAttributeModified(
+        Object element,
+        String name,
+        String value);
+
+    public void onAttributeRemoved(
+        Object element,
+        String name);
+
+    public void onChildInserted(
+        Object parentElement,
+        @Nullable Object previousElement,
+        Object childElement);
+
+    public void onChildRemoved(
+        Object parentElement,
+        Object childElement);
+  }
+}
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/Descriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/Descriptor.java
@@ -1,0 +1,41 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+package com.facebook.stetho.inspector.elements;
+
+import javax.annotation.Nullable;
+
+public abstract class Descriptor implements NodeDescriptor {
+
+  private Listener mListener;
+
+  protected Descriptor() {
+  }
+
+  void setListener(Listener listener) {
+    mListener = listener;
+  }
+
+  protected final Listener getListener() {
+    return mListener;
+  }
+
+  public interface Listener {
+    public void onAttributeModified(
+        Object element,
+        String name,
+        String value);
+
+    public void onAttributeRemoved(
+        Object element,
+        String name);
+
+    public void onChildInserted(
+        Object parentElement,
+        @Nullable Object previousElement,
+        Object childElement);
+
+    public void onChildRemoved(
+        Object parentElement,
+        Object childElement);
+  }
+}
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/DescriptorMap.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/DescriptorMap.java
@@ -1,0 +1,84 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+package com.facebook.stetho.inspector.elements;
+
+import com.facebook.stetho.common.Util;
+
+import java.util.IdentityHashMap;
+import java.util.Map;
+
+public final class DescriptorMap {
+  private final Map<Class<?>, Descriptor> mMap = new IdentityHashMap<Class<?>, Descriptor>();
+  private boolean mIsInitializing;
+  private Descriptor.Listener mListener;
+
+  public DescriptorMap beginInit() {
+    Util.throwIf(mIsInitializing);
+    mIsInitializing = true;
+    return this;
+  }
+
+  public DescriptorMap register(Class<?> elementClass, Descriptor nodeDescriptor) {
+    Util.throwIfNull(elementClass);
+    Util.throwIfNull(nodeDescriptor);
+    Util.throwIfNot(mIsInitializing);
+
+    if (mMap.containsKey(elementClass)) {
+      throw new UnsupportedOperationException();
+    }
+
+    mMap.put(elementClass, nodeDescriptor);
+    return this;
+  }
+
+  public DescriptorMap setListener(Descriptor.Listener listener) {
+    Util.throwIfNull(listener);
+    Util.throwIfNot(mIsInitializing);
+    Util.throwIfNotNull(mListener);
+
+    mListener = listener;
+
+    return this;
+  }
+
+  public DescriptorMap endInit() {
+    Util.throwIfNot(mIsInitializing);
+    Util.throwIfNull(mListener);
+
+    for (final Class<?> elementClass : mMap.keySet()) {
+      final Descriptor nodeDescriptor = mMap.get(elementClass);
+
+      if (nodeDescriptor instanceof ChainedDescriptor) {
+        final ChainedDescriptor<?> chainedNodeDescriptor = (ChainedDescriptor<?>)nodeDescriptor;
+        Class<?> superClass = elementClass.getSuperclass();
+        Descriptor superNodeDescriptor = getImpl(superClass);
+        chainedNodeDescriptor.setSuper(superNodeDescriptor);
+      }
+
+      nodeDescriptor.setListener(mListener);
+    }
+
+    mIsInitializing = false;
+    return this;
+  }
+
+  public Descriptor get(Class<?> elementClass) {
+    Util.throwIfNull(elementClass);
+    Util.throwIf(mIsInitializing);
+    return getImpl(elementClass);
+  }
+
+  private Descriptor getImpl(final Class<?> elementClass) {
+    Class<?> theClass = elementClass;
+    while (theClass != null) {
+      Descriptor nodeDescriptor = mMap.get(theClass);
+      if (nodeDescriptor != null) {
+        return nodeDescriptor;
+      }
+
+      theClass = theClass.getSuperclass();
+    }
+
+    return null;
+  }
+}
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/ObjectDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/ObjectDescriptor.java
@@ -1,0 +1,53 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+package com.facebook.stetho.inspector.elements;
+
+public final class ObjectDescriptor extends Descriptor {
+  @Override
+  public void hook(Object element) {
+  }
+
+  @Override
+  public void unhook(Object element) {
+  }
+
+  @Override
+  public NodeType getNodeType(Object element) {
+    return NodeType.ELEMENT_NODE;
+  }
+
+  @Override
+  public String getNodeName(Object element) {
+    return element.getClass().getName();
+  }
+
+  @Override
+  public String getLocalName(Object element) {
+    return getNodeName(element);
+  }
+
+  @Override
+  public String getNodeValue(Object element) {
+    return null;
+  }
+
+  @Override
+  public int getChildCount(Object element) {
+    return 0;
+  }
+
+  @Override
+  public Object getChildAt(Object element, int index) {
+    throw new IndexOutOfBoundsException();
+  }
+
+  @Override
+  public int getAttributeCount(Object element) {
+    return 0;
+  }
+
+  @Override
+  public void copyAttributeAt(Object element, int index, NodeAttribute outAttribute) {
+    throw new IndexOutOfBoundsException();
+  }
+}
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ApplicationDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ApplicationDescriptor.java
@@ -1,0 +1,166 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+package com.facebook.stetho.inspector.elements.android;
+
+import android.annotation.TargetApi;
+import android.app.Activity;
+import android.app.Application;
+import android.os.Build;
+import android.os.Bundle;
+
+import com.facebook.stetho.common.LogUtil;
+import com.facebook.stetho.common.Util;
+import com.facebook.stetho.common.android.ApplicationUtil;
+import com.facebook.stetho.inspector.elements.ChainedDescriptor;
+import com.facebook.stetho.inspector.elements.NodeType;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.IdentityHashMap;
+import java.util.List;
+import java.util.Map;
+
+final class ApplicationDescriptor extends ChainedDescriptor<Application> {
+  private static final String TAG = "ApplicationDescriptor";
+
+  private final Map<Application, ElementContext> mElementToContextMap =
+      Collections.synchronizedMap(new IdentityHashMap<Application, ElementContext>());
+
+  private ElementContext getContext(Application element) {
+    return mElementToContextMap.get(element);
+  }
+
+  @Override
+  protected void onHook(Application element) {
+    ElementContext context = newElementContext();
+    context.hook(element);
+    mElementToContextMap.put(element, context);
+  }
+
+  @Override
+  protected void onUnhook(Application element) {
+    ElementContext context = mElementToContextMap.remove(element);
+    context.unhook();
+  }
+
+  @Override
+  protected NodeType onGetNodeType(Application element) {
+    return NodeType.DOCUMENT_NODE;
+  }
+
+  @Override
+  protected int onGetChildCount(Application element) {
+    ElementContext context = getContext(element);
+    return context.getActivitiesList().size();
+  }
+
+  @Override
+  protected Object onGetChildAt(Application element, int index) {
+    ElementContext context = getContext(element);
+    return context.getActivitiesList().get(index);
+  }
+
+  private ElementContext newElementContext() {
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.ICE_CREAM_SANDWICH) {
+      return new ElementContextICS();
+    } else {
+      LogUtil.w(TAG, "Running on pre-ICS: must manually reload inspector when Activity changes");
+      return new ElementContextPreICS();
+    }
+  }
+
+  private abstract class ElementContext {
+    public abstract void hook(Application element);
+
+    public abstract void unhook();
+
+    public abstract List<Activity> getActivitiesList();
+  }
+
+  private final class ElementContextPreICS extends ElementContext {
+    @Override
+    public void hook(Application element) {
+    }
+
+    @Override
+    public void unhook() {
+    }
+
+    @Override
+    public List<Activity> getActivitiesList() {
+      return ApplicationUtil.getAllActivities();
+    }
+  }
+
+  @TargetApi(Build.VERSION_CODES.ICE_CREAM_SANDWICH)
+  private final class ElementContextICS extends ElementContext {
+    private Application mElement;
+    private Application.ActivityLifecycleCallbacks mCallbacks;
+    private List<Activity> mActivities;
+
+    @Override
+    public void hook(Application element) {
+      mElement = Util.throwIfNull(element);
+      mActivities = new ArrayList<Activity>();
+
+      // TODO: tree diffing will remove the need to even worry about installing this callback,
+      //       which will then allow ~realtime updates for pre-ICS. for now, pre-ICS will just
+      //       have to close and re-open the inspector whenever the activity changes.
+      //       (this does assume that the hack below for getAllActivitiesHack() works on pre-ICS)
+      mCallbacks = new Application.ActivityLifecycleCallbacks() {
+        @Override
+        public void onActivityCreated(Activity activity, Bundle savedInstanceState) {
+          mActivities.add(0, activity);
+          getListener().onChildInserted(mElement, null, activity);
+        }
+
+        @Override
+        public void onActivityStarted(Activity activity) {
+        }
+
+        @Override
+        public void onActivityResumed(Activity activity) {
+        }
+
+        @Override
+        public void onActivityPaused(Activity activity) {
+        }
+
+        @Override
+        public void onActivityStopped(Activity activity) {
+        }
+
+        @Override
+        public void onActivitySaveInstanceState(Activity activity, Bundle outState) {
+        }
+
+        @Override
+        public void onActivityDestroyed(Activity activity) {
+          mActivities.remove(activity);
+          getListener().onChildRemoved(mElement, activity);
+        }
+      };
+
+      mElement.registerActivityLifecycleCallbacks(mCallbacks);
+
+      mActivities = ApplicationUtil.getAllActivities();
+    }
+
+    @Override
+    public void unhook() {
+      if (mElement != null) {
+        if (mCallbacks != null) {
+          mElement.unregisterActivityLifecycleCallbacks(mCallbacks);
+          mCallbacks = null;
+        }
+
+        mElement = null;
+      }
+    }
+
+    @Override
+    public List<Activity> getActivitiesList() {
+      return mActivities;
+    }
+  }
+}
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/TextViewDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/TextViewDescriptor.java
@@ -1,0 +1,87 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+package com.facebook.stetho.inspector.elements.android;
+
+import android.text.Editable;
+import android.text.TextWatcher;
+import android.widget.TextView;
+
+import com.facebook.stetho.common.Util;
+import com.facebook.stetho.inspector.elements.ChainedDescriptor;
+import com.facebook.stetho.inspector.elements.NodeAttribute;
+
+import java.util.Collections;
+import java.util.IdentityHashMap;
+import java.util.Map;
+
+final class TextViewDescriptor extends ChainedDescriptor<TextView> {
+  private static final String TEXT_ATTRIBUTE_NAME = "text";
+
+  private final Map<TextView, ElementContext> mElementToContextMap =
+      Collections.synchronizedMap(new IdentityHashMap<TextView, ElementContext>());
+
+  @Override
+  protected void onHook(final TextView element) {
+    ElementContext context = new ElementContext();
+    context.hook(element);
+    mElementToContextMap.put(element, context);
+  }
+
+  protected void onUnhook(TextView element) {
+    ElementContext context = mElementToContextMap.remove(element);
+    context.unhook();
+  }
+
+  @Override
+  protected int onGetAttributeCount(TextView element) {
+    return (element.getText().length() == 0) ? 0 : 1;
+  }
+
+  @Override
+  protected void onCopyAttributeAt(TextView element, int index, NodeAttribute outAttribute) {
+    if (index != 0) {
+      throw new IndexOutOfBoundsException();
+    }
+
+    CharSequence text = element.getText();
+    if (text.length() == 0) {
+      throw new IndexOutOfBoundsException();
+    } else {
+      outAttribute.name = TEXT_ATTRIBUTE_NAME;
+      outAttribute.value = text.toString();
+    }
+  }
+
+  private final class ElementContext implements TextWatcher {
+    private TextView mElement;
+
+    public void hook(TextView element) {
+      mElement = Util.throwIfNull(element);
+      mElement.addTextChangedListener(this);
+    }
+
+    public void unhook() {
+      if (mElement != null) {
+        mElement.removeTextChangedListener(this);
+        mElement = null;
+      }
+    }
+
+    @Override
+    public void beforeTextChanged(CharSequence s, int start, int count, int after) {
+    }
+
+    @Override
+    public void onTextChanged(CharSequence s, int start, int before, int count) {
+    }
+
+    @Override
+    public void afterTextChanged(Editable s) {
+      if (s.length() == 0) {
+        getListener().onAttributeRemoved(mElement, TEXT_ATTRIBUTE_NAME);
+      } else {
+        getListener().onAttributeModified(mElement, TEXT_ATTRIBUTE_NAME, s.toString());
+      }
+    }
+  }
+}
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
@@ -1,0 +1,20 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+package com.facebook.stetho.inspector.elements.android;
+
+import android.view.View;
+
+import com.facebook.stetho.common.StringUtil;
+import com.facebook.stetho.inspector.elements.ChainedDescriptor;
+
+final class ViewDescriptor extends ChainedDescriptor<View> {
+  @Override
+  protected String onGetNodeName(View element) {
+    String className = element.getClass().getName();
+
+    return
+        StringUtil.removePrefix(className, "android.view.",
+        StringUtil.removePrefix(className, "android.widget."));
+  }
+}
+
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewGroupDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewGroupDescriptor.java
@@ -1,0 +1,164 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+package com.facebook.stetho.inspector.elements.android;
+
+import android.view.View;
+import android.view.ViewGroup;
+
+import com.facebook.stetho.common.Util;
+import com.facebook.stetho.inspector.elements.ChainedDescriptor;
+
+import java.lang.reflect.Field;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+final class ViewGroupDescriptor extends ChainedDescriptor<ViewGroup> {
+  // TODO: We're probably going to switch to another way of determining structural
+  //       changes in the View tree. So this gizmo with chaining OnHierarchyChangeListener
+  //       via reflection should go away soon.
+
+  private static final Field sOnHierarchyChangeListenerField;
+  static {
+    Field field;
+    try {
+      field = ViewGroup.class.getDeclaredField("mOnHierarchyChangeListener");
+    } catch (NoSuchFieldException ex) {
+      field = null;
+    }
+
+    sOnHierarchyChangeListenerField = field;
+
+    if (sOnHierarchyChangeListenerField != null) {
+      sOnHierarchyChangeListenerField.setAccessible(true);
+    }
+  }
+
+  private final Map<ViewGroup, ElementContext> mElementToContextMap =
+      Collections.synchronizedMap(new HashMap<ViewGroup, ElementContext>());
+
+  public ViewGroupDescriptor() {
+  }
+
+  @Override
+  protected void onHook(ViewGroup element) {
+    ElementContext context = new ElementContext();
+    context.hook(element);
+    mElementToContextMap.put(element, context);
+  }
+
+  @Override
+  protected void onUnhook(ViewGroup element) {
+    ElementContext context = mElementToContextMap.remove(element);
+    context.unhook();
+  }
+
+  @Override
+  protected int onGetChildCount(ViewGroup element) {
+    return element.getChildCount();
+  }
+
+  @Override
+  protected Object onGetChildAt(ViewGroup element, int index) {
+    if (index < 0 || index >= element.getChildCount()) {
+      throw new IndexOutOfBoundsException();
+    }
+
+    return element.getChildAt(index);
+  }
+
+  private static ViewGroup.OnHierarchyChangeListener getOnHierarchyChangeListenerHack(
+      ViewGroup viewGroup) throws NoSuchFieldException {
+    if (sOnHierarchyChangeListenerField == null) {
+      throw new NoSuchFieldException();
+    }
+
+    try {
+      return (ViewGroup.OnHierarchyChangeListener) sOnHierarchyChangeListenerField.get(viewGroup);
+    } catch (IllegalAccessException ex) {
+      // should not happen since we called setAccessible(true)
+      throw new IllegalAccessError(ex.getMessage());
+    }
+  }
+
+  private static int findChildIndex(ViewGroup parent, View child) {
+    int count = parent.getChildCount();
+    for (int i = 0; i < count; ++i) {
+      if (parent.getChildAt(i) == child) {
+        return i;
+      }
+    }
+    return -1;
+  }
+
+  private final class ElementContext implements ViewGroup.OnHierarchyChangeListener {
+
+    private ViewGroup mElement;
+    private ViewGroup.OnHierarchyChangeListener mInnerListener;
+
+    public void hook(ViewGroup element) {
+      mElement = Util.throwIfNull(element);
+
+      ViewGroup.OnHierarchyChangeListener innerListener;
+      try {
+        innerListener = getOnHierarchyChangeListenerHack(mElement);
+      } catch (NoSuchFieldException ex) {
+        innerListener = null;
+      }
+      mInnerListener = innerListener;
+
+      mElement.setOnHierarchyChangeListener(this);
+    }
+
+    public void unhook() {
+      if (mElement != null) {
+        ViewGroup.OnHierarchyChangeListener currentListener;
+        try {
+          currentListener = getOnHierarchyChangeListenerHack(mElement);
+        } catch (NoSuchFieldException ex) {
+          currentListener = null;
+        }
+
+        if (currentListener == this) {
+          mElement.setOnHierarchyChangeListener(mInnerListener);
+        } else {
+          mElement.setOnHierarchyChangeListener(null);
+        }
+
+        mInnerListener = null;
+        mElement = null;
+      }
+    }
+
+    @Override
+    public void onChildViewAdded(View parent, View child) {
+      if (mElement == null) {
+        return;
+      }
+
+      if (mInnerListener != null) {
+        mInnerListener.onChildViewAdded(parent, child);
+      }
+
+      if (parent instanceof ViewGroup) {
+        final ViewGroup parentGroup = (ViewGroup)parent;
+        int index = findChildIndex(parentGroup, child);
+        View previousChild = (index == 0) ? null : parentGroup.getChildAt(index - 1);
+        getListener().onChildInserted(parent, previousChild, child);
+      }
+    }
+
+    @Override
+    public void onChildViewRemoved(View parent, View child) {
+      if (mElement == null) {
+        return;
+      }
+
+      if (mInnerListener != null) {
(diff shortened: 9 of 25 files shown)


jasta:
`id="..."` is probably one of the first important features to add next.  It's super hard to make sense of the hierarchy otherwise :)

stetho-sample/src/main/AndroidManifest.xml
```diff
@@ -4,6 +4,7 @@
     package="com.facebook.stetho.sample">
 
   <uses-permission android:name="android.permission.INTERNET" />
+  <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW" />
```
jasta: As discussed, this won't fly :)

stetho/src/main/java/com/facebook/stetho/Stetho.java
```diff
@@ -105,7 +106,7 @@ public static InspectorModulesProvider defaultInspectorModulesProvider(final Con
         modules.add(new Console());
         modules.add(new CSS());
         modules.add(new Debugger());
-        modules.add(new DOM());
+        modules.add(new DOM(new UIDOMProvider(context)));
```
jasta: This should be internalized to `DOM`.  Just accept a `Context`.  This is a public API for Stetho and folks can subclass `DOM`.

stetho/src/main/java/com/facebook/stetho/inspector/elements/ActivityNodeDescriptor.java
```diff
@@ -0,0 +1,32 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+package com.facebook.stetho.inspector.elements;
+
+import android.app.Activity;
+import android.view.Window;
+
+final class ActivityNodeDescriptor extends ChainedNodeDescriptor<Activity> {
+  public ActivityNodeDescriptor() {
```
jasta: Empty, public constructors are implied in Java

stetho/src/main/java/com/facebook/stetho/inspector/elements/ActivityNodeDescriptor.java
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  // TODO: do we have to worry about listening for when the Activity's Window transitions
```
jasta: What is this comment referring to?  When what exactly goes null?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -2,30 +2,315 @@
...
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.GuardedBy;
+
 public class DOM implements ChromeDevtoolsDomain {
-  public DOM() {
+  private final Provider mDOMProvider;
+  private final DOMPeerManager mPeerManager;
+  private final ObjectMapper mObjectMapper;
+
+  private final Object mNodeMapSync = new Object();
```
jasta: This can be better expressed by a separate class which handles `mNextNodeId`, `mElementToNodeIdMap`, etc.  Easier to reason about and can be re-used later by `Console`.  Something like `ObjectIdMapper`.  Could be put into `inspector/helper`.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -2,30 +2,315 @@
...
+  @GuardedBy("mNodeMapSync")
+  private int mNextNodeId = 1;
+
+  @GuardedBy("mNodeMapSync")
+  private final Map<Object, Integer> mElementToNodeIdMap = new HashMap<>();
+
+  @GuardedBy("mNodeMapSync")
+  private final SparseArray<Object> mNodeIdToElementMap = new SparseArray<>();
+
+  public DOM(Provider provider) {
```
jasta: Oh, I think I misspoke previously on my advice about accepting only `Context` here.  This probably is best expressed as this generic interface :)

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -2,30 +2,315 @@
...
+    mDOMProvider = Util.throwIfNull(provider);
+    mDOMProvider.setListener(new ProviderListener());
+
+    mPeerManager = new DOMPeerManager();
+    mObjectMapper = new ObjectMapper();
+  }
+
+  private static final class DOMPeerManager extends ChromePeerManager {
```
jasta: Separate class.

Also, I'm thinking you should invoke `mDOMProvider.setListener` during `onFirstPeerRegistered`.  See `DatabasePeerManager` for reference.  This could be a good opportunity to upgrade the `setListener` API to `addListener`/`removeListener` as well.

stetho/src/main/java/com/facebook/stetho/inspector/elements/UIDOMProvider.java
```diff
@@ -0,0 +1,101 @@
...
+import android.widget.TextView;
+
+import com.facebook.stetho.inspector.protocol.module.DOM;
+
+public final class UIDOMProvider implements DOM.Provider {
+  private final NodeDescriptorMap mNodeDescriptorMap;
+  private final Application mApplication;
+  private Listener mListener;
+
+  public UIDOMProvider(Context context) {
```
jasta: This should be renamed `AndroidDOMProvider`, no?

stetho/src/main/java/com/facebook/stetho/inspector/elements/UIDOMProvider.java
```diff
@@ -0,0 +1,101 @@
...
+    mApplication = (Application)context.getApplicationContext();
+
+    mNodeDescriptorMap = new NodeDescriptorMap()
```
jasta: This is generally an anti-pattern on Android.  Pre-initializing a mapping data structure adds to the heap and adds init overhead.  Strongly recommend either lazily initializing on first access or using the imperative pattern (that is, a function that translates Class/Object instances to `*NodeDescriptor`)

stetho/src/main/java/com/facebook/stetho/inspector/elements/UIDOMProvider.java
```diff
@@ -0,0 +1,101 @@
...
+      mListener.onChildRemoved(parentElement, childElement);
+    }
+  }
+
+  @Override
+  public void setListener(Listener listener) {
+    mListener = listener;
+  }
+
+  /*private static boolean canEnableHighlight(Activity activity) {
```
jasta: Remove dead code

stetho/src/main/java/com/facebook/stetho/inspector/elements/UIDOMProvider.java
```diff
@@ -0,0 +1,101 @@
...
+
+import com.facebook.stetho.inspector.protocol.module.DOM;
+
+public final class UIDOMProvider implements DOM.Provider {
+  private final NodeDescriptorMap mNodeDescriptorMap;
+  private final Application mApplication;
+  private Listener mListener;
+
+  public UIDOMProvider(Context context) {
+    mApplication = (Application)context.getApplicationContext();
```
jasta: This is a pretty dangerous assumption to make at such a low level.  Strongly prefer requiring the caller pass in `Application` (which moves the assumption higher up the stack).  This will make it easier for us to correct it by actually requiring the `Application` object at some point.

stetho/src/main/java/com/facebook/stetho/inspector/elements/UIDOMProvider.java
```diff
@@ -0,0 +1,101 @@
...
+    return mApplication;
+  }
+
+  @Override
+  public DOM.NodeDescriptor getNodeDescriptor(Object element) {
+    return (element == null) ? null : mNodeDescriptorMap.get(element.getClass());
+  }
+
+  @Override
+  public void highlightElement(Object element) {
```
jasta: Per our discussion, hide this behind another abstraction which can work with `ViewOverlay` when available.

stetho/src/main/java/com/facebook/stetho/inspector/elements/UIDOMProvider.java
```diff
@@ -0,0 +1,101 @@
...
+    if (element instanceof View) {
```
jasta: That seems weird.  Why can't we just have `highlight` be part of the `NodeDescriptor` API?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Runtime.java
```diff
@@ -9,6 +9,8 @@
 
 import com.facebook.stetho.json.annotation.JsonProperty;
 import com.facebook.stetho.json.annotation.JsonValue;
+
+import org.json.JSONException;
```
jasta: Remove

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -2,30 +2,315 @@
...
+    public DOMPeerManager() {
+    }
+  }
+
+  private final class ProviderListener implements Provider.Listener {
```
jasta: Very strangely positioned here.  Add a method to `DOMPeerManager` so that it can internally delegate to `sendNotificationToPeers`.

jasta: As I think a little more.  Maybe you can even shove the listener itself into `DOMPeerManager` and just have it take over ownership.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -2,30 +2,315 @@
...
+      mPeerManager.sendNotificationToPeers("DOM.attributeRemoved", message);
+    }
+
+    @Override
+    public void onChildInserted(Object parentElement, Object previousElement, Object childElement) {
+      ChildNodeInsertedEvent message = new ChildNodeInsertedEvent();
+      message.parentNodeId = getOrCreateNodeIdForElement(parentElement);
+
+      message.previousNodeId = (previousElement == null)
+          ? -1  // using -1 was just a guess, and it seemed to work. this is for the case where we
```
jasta: style nit: avoid hanging comments; use:
```java
// Comment here
weirdThingHere();
```

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -2,30 +2,315 @@
...
+  private static class ChildNodeRemovedEvent {
+    @JsonProperty(required = true)
+    public int parentNodeId;
+
+    @JsonProperty(required = true)
+    public int nodeId;
+  }
+
+  @ChromeDevtoolsMethod
+  public void enable(JsonRpcPeer peer, JSONObject params) {
```
jasta: style nit: inner classes are never to be defined before methods.  Push them to the end.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -2,30 +2,315 @@
...
   @ChromeDevtoolsMethod
   public JsonRpcResult getDocument(JsonRpcPeer peer, JSONObject params) {
-    Node node = new Node();
-    node.nodeId = 1;
-    node.nodeType = 1;
-    node.nodeName = "";
-    node.nodeValue = "";
-    return node;
+    Object rootElement = mDOMProvider.getRootElement();
+    if (rootElement == null) {
```
jasta: Are you sure this works?  What does DevTools do when you return null here?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -2,30 +2,315 @@
...
+      return null;
+    }
+
+    Node rootNode = createNodeForElement(rootElement);
```
jasta: Why not `getOrCreate...` here?  What if `getDocument` is called twice?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -2,30 +2,315 @@
...
+    @JsonProperty(required = false)
+    public Integer nodeId;
+
+    @JsonProperty(required = false)
+    public String objectId;
+  }
+
+  public static class HighlightConfig {
+    @JsonProperty(required = false)
+    public Boolean showInfo;
```
jasta: What is this?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -2,30 +2,315 @@
...
   @ChromeDevtoolsMethod
   public void hideHighlight(JsonRpcPeer peer, JSONObject params) {
+    mDOMProvider.hideHighlight();
+  }
+
+  public static class HighlightNodeRequest {
+    @JsonProperty(required = true)
+    public HighlightConfig highlightConfig;
+
+    @JsonProperty(required = false)
```
jasta: Technically `required = false` is implied.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -2,30 +2,315 @@
...
+
+    GetDocumentResponse result = new GetDocumentResponse();
+    result.root = rootNode;
+    return result;
+  }
+
+  @ChromeDevtoolsMethod
+  public void highlightNode(JsonRpcPeer peer, JSONObject params) {
+    HighlightNodeRequest request = mObjectMapper.convertValue(params, HighlightNodeRequest.class);
+    if (request.nodeId != null) {
```
jasta: Given the complexities of debugging network protocols, strongly prefer that you have an else block that sends to logcat (via `LogUtil`)

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -2,30 +2,315 @@
...
+    public int g;
+
+    @JsonProperty(required = true)
+    public int b;
+
+    @JsonProperty(required = false)
+    public Double a;
+  }
+
+  private void unhookElement(Object element) {
```
jasta: I think this is fixed by moving the object id mapping into another class, but `unhook` isn't really a consistent verb here.  You're really mapping and unmapping objects.  Or adding and removing.  Not hooking and unhooking.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -2,30 +2,315 @@
...
+    node.nodeName = nodeDescriptor.getNodeName(element);
+    node.localName = nodeDescriptor.getLocalName(element);
+    node.nodeValue = nodeDescriptor.getNodeValue(element);
+
+    node.children = getChildNodesForElement(element);
+    node.childNodeCount = node.children.size();
+
+    int attributeCount = nodeDescriptor.getAttributeCount(element);
+    if (attributeCount > 0) {
+      node.attributes = new ArrayList<>(attributeCount * 2);
```
jasta: Danger.  We haven't yet resolved #44.  I can tackle this if you want, but it needs to be a precursor.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -2,30 +2,315 @@
...
+      NodeAttribute attribute = new NodeAttribute();
+
+      for (int i = 0; i < attributeCount; ++i) {
+        nodeDescriptor.getAttributeAt(element, i, attribute);
```
jasta: This is highly unusual nomenclature for Java.  A getter would always be expected to return something.  Instead, use: `copyAttributeAt`.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -2,30 +2,315 @@
...
+        if (attribute.name == null || attribute.value == null) {
+          throw new IllegalArgumentException();
+        }
+
+        node.attributes.add(attribute.name);
+        node.attributes.add(attribute.value);
+
+        attribute.name = null;
```
jasta: This has a bad design smell.  If you need to do this, then you also need `copyAttributeAt` above to return a boolean which you must test for.  Generally, you should be able to predictably depend on the attribute being fully specified after `copyAttributeAt` successfully returns.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -2,30 +2,315 @@
...
+        attribute.value = null;
+      }
+    }
+
+    return node;
+  }
+
+  private List<Node> getChildNodesForElement(Object element) {
```
jasta: nit: this method is probably better expressed being marked `static` and accepting the `DOMProvider`.  It clarifies that it has no real stateful side effects and could be applied generically.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -43,5 +328,95 @@ public void hideHighlight(JsonRpcPeer peer, JSONObject params) {
...
+    DOCUMENT_TYPE_NODE(10),
+    DOCUMENT_FRAGMENT_NODE(11);
+
+    private int mValue;
+
+    private NodeType(int value) {
+      mValue = value;
+    }
+
+    int getValue() {
```
jasta: Err.  What is this method hooked up to?  Isn't `@JsonMethod` or whatever required?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -43,5 +328,95 @@ public void hideHighlight(JsonRpcPeer peer, JSONObject params) {
...
+      return mValue;
+    }
+  }
+
+  public static interface Provider {
```
jasta: Separate file, in `inspector/elements`

jasta: javadocs would be nice :)

jasta:
Overall I really love this.  Awesome job.

Please do look into Travis CI build failures though :)

stetho/src/main/java/com/facebook/stetho/inspector/elements/ActivityNodeDescriptor.java
rickbrew: It goes from null to non-null ... when an Activity is first created, it isn't yet plugged into a Window, right? And yet there's nothing here hooking in to listen to when the Window is attached. I think we'll be able to rely on the tree diffing stuff though for that.

stetho/src/main/java/com/facebook/stetho/inspector/elements/UIDOMProvider.java
rickbrew: I scratched my head on the right way to do this for a bit, and finally decided on mirroring the DOM JSON API as much as possible in the Java code; highlightNode and hideHighlight are methods on the JSON DOM protocol, therefore they're methods on our DOM.Provider. This should make it easier to navigate and understand the two if they are mirrors of each other. Plus, from a practical standpoint, this would require the NodeDescriptor to coordinate with the DOM.Provider. There isn't currently a channel of communication in that direction and I wasn't sure this was the right justification for adding it. It's very easy for a DOM.Provider implementation to delegate this call to NodeDescriptor if that's the strategy it chooses to use (just use getNodeDescriptor!), but the reverse isn't as easy (for NodeDescriptor to delegate to DOM.Provider).

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: Well it didn't crash ... :) Should I throw an exception?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: The class or the field? HighlightConfig is part of the HighlightNode payload, and showInfo is part of it. We're not using it though ... can probably remove it? The old docs I was using has far fewer fields than the newer and more official docs we found

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: Except that it is stateful because it calls into createNodeForElement, which is also stateful because it calls into getOrCreateNodeIdForElement

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: Yes, I like that name better too. I was mimicking some UI call, getBounds(Rect outRect), or something.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: This is meant as a defensive play to help catch bugs in the NodeDescriptor. Otherwise there's no way to detect if copyAttributeAt accidentally forgot to set name or value. If the index is out of bounds then I'd expect an IndexOutOfBoundsException, otherwise .node and .value must be non-null.

stetho/src/main/java/com/facebook/stetho/inspector/elements/ActivityNodeDescriptor.java
jasta: When an Activity's onCreate is called, it is attached to a window.    This is the point where you are inserting it into your structure.

stetho/src/main/java/com/facebook/stetho/inspector/elements/ApplicationNodeDescriptor.java
```diff
@@ -0,0 +1,161 @@
...
+import java.lang.reflect.InvocationTargetException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+final class ApplicationNodeDescriptor extends ChainedNodeDescriptor<Application> {
+  private ConcurrentMap<Application, ElementContext> mElementToContextMap =
+      new ConcurrentHashMap<>();
```
jasta: Never use this on Android.  It is designed for high throughput in highly concurrent environments at the expensive of RAM.  Use `Collections.synchronizedMap(new HashMap...` instead.

As an aside, you don't need a map.  `Application` is a singleton in Android.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -2,30 +2,315 @@
...
+      mPeerManager.sendNotificationToPeers("DOM.childNodeInserted", message);
+    }
+
+    @Override
+    public void onChildRemoved(Object parentElement, Object childElement) {
+      ChildNodeRemovedEvent message = new ChildNodeRemovedEvent();
+      message.parentNodeId = getOrCreateNodeIdForElement(parentElement);
+      message.nodeId = getOrCreateNodeIdForElement(childElement);
+      mPeerManager.sendNotificationToPeers("DOM.childNodeRemoved", message);
+      unhookElement(childElement);
```
jasta: I just realized there is a serious bug here in that you do not run any logic in `onLastPeerUnregistered`.  This should be leaking, no?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
jasta: You can't throw because this is a reasonable case.  Start the sample app, hit back to dismiss all UI, open chrome://inspect.  This case should occur.  But what actually happens in the DevTools UI?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
jasta: I meant what is `showInfo`, the field.  Not important, was just curious...

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
jasta: I'd give Android framework engineers the same criticism :)  Though they sometimes get it right: `Cursor#copyStringToBuffer(...)`

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
jasta: Defensive programming is anything but here.  It would lead to subtle bugs going undetected because new fields might not be copied but the caller would expect that they were.  Document that all fields must be set so it's clear that a bug exists if this happens in practice.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
jasta: Oops, missed that.  Nevermind :)

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: Yeah I was thinking of that too

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: IIRC, it had an empty Elements tab

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: I've switched all instances of the diamond operator, here and elsewhere, to not use it

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: I was using an int in the Node for nodeType, but now I see the @JsonValue patter and have switched to it

stetho-sample/src/main/AndroidManifest.xml
rickbrew: Yup, next update has this fixed up!

stetho/src/main/java/com/facebook/stetho/Stetho.java
```diff
@@ -105,7 +108,14 @@ public static InspectorModulesProvider defaultInspectorModulesProvider(final Con
         modules.add(new Console());
         modules.add(new CSS());
         modules.add(new Debugger());
-        modules.add(new DOM());
+
+        modules.add(new DOM(new DOMProvider.Factory() {
```
jasta: This code should continue to read as one-liners.  It's intended as an extensibility point and should be terse.  Prefer:
```java
modules.add(new DOM(new AndroidDOMProviderFactory()));
```

stetho/src/main/java/com/facebook/stetho/inspector/elements/DOMProvider.java
```diff
@@ -0,0 +1,46 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+package com.facebook.stetho.inspector.elements;
+
+import javax.annotation.Nullable;
+
+// TODO: javadocs
```
jasta: Prefer to remove this comment and leave a note for yourself out of band to follow-up.  It's unprofessional IMO to accidentally leave in a TODO like this.

stetho/src/main/java/com/facebook/stetho/inspector/elements/DOMProvider.java
```diff
@@ -0,0 +1,46 @@
...
+
+public interface DOMProvider {
+  public static interface Factory {
+    DOMProvider create();
+  }
+
+  public static interface Listener {
```
jasta: style nit: inner classes should be defined after methods.

stetho/src/main/java/com/facebook/stetho/inspector/elements/ObjectDescriptor.java
```diff
@@ -0,0 +1,92 @@
...
+
+  private String getClassFriendlyName(Class<?> elementClass) {
+    if (mClassToFriendlyNameCache.containsKey(elementClass)) {
+      return mClassToFriendlyNameCache.get(elementClass);
+    }
+
+    String className = elementClass.getName();
+
+    String friendlyClassName =
+        removePrefix(className, "android.app.",
```
jasta: This is at the wrong layer of inheritance.  Only `View` subclasses can be in `android.{view,widget}.*`, and only `Activity` can be in `android.app.*`.  The substitutions should be happening in `ViewDescriptor` and `ActivityDescriptor` as a result.

stetho/src/main/java/com/facebook/stetho/inspector/elements/ObjectDescriptor.java
```diff
@@ -0,0 +1,92 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+package com.facebook.stetho.inspector.elements;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public final class ObjectDescriptor extends Descriptor {
+  private final Map<Class<?>, String> mClassToFriendlyNameCache =
```
jasta: You probably don't want or need this.  Running indexOf/substring is almost certainly faster than maintaining and looking up in a hash map.  Remember, in order to lookup in a hash map, you must call String.equals on the key which will iterate through each character of the string.  indexOf and substring will iterate through only part of the string.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -2,30 +2,233 @@
...
+  private DOMObjectIdMapper mObjectIdMapper;
+
+  public DOM(DOMProvider.Factory providerFactory) {
+    mDOMProviderFactory = Util.throwIfNull(providerFactory);
+
+    mPeerManager = new ChromePeerManager();
+    mPeerManager.setListener(new PeerManagerListener());
+  }
+
+  private final class PeerManagerListener extends PeersRegisteredListener {
```
jasta: style nit again: please move class definition below methods :)

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -2,30 +2,233 @@
...
+    @Override
+    protected void onFirstPeerRegistered() {
+      assert(mDOMProvider == null);
```
jasta: Do not use `assert`.  It's tragically broken in Java.  Use `Util.throwIfNotNull`.  Also, you have a race here if you don't synchronize these two methods.  The base class doesn't do it because I was trying not to be heavy handed (maybe I should've though)

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -2,30 +2,233 @@
...
+
+      mObjectMapper = new ObjectMapper();
```
jasta: This is unnecessary FWIW.  `ObjectMapper` is basically just static functions (we should have made them static functions actually).  It has no initialization and accounts for just 1 object.  Likewise, `mObjectIdMapper` is also very small if you clear the data structure.  It's only `mDOMProvider` that's considered large and is special cased using this pattern.

It's your choice as to whether you adjust things, but you must use `@Nullable` for all fields that you deinitialization in this way which is where I think you'll find that the choice is a little uglier than I think you'd prefer (IntelliJ will start warning you all over about possible null access).

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -2,30 +2,233 @@
...
+
+    @Override
+    protected void onLastPeerUnregistered() {
+      Object rootElement = mDOMProvider.getRootElement();
+      removeElementTree(rootElement);
+
+      mDOMProvider.dispose();
+      mDOMProvider = null;
+
+      mObjectIdMapper = null;
```
jasta: This is sloppy IMO.  There should be a `clear` method that you explicitly invoke.  This should let you continue to hold the reference safely with minimal memory usage.  See previous comment for more context.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewHighlighter.java
```diff
@@ -0,0 +1,112 @@
...
+
+import java.util.concurrent.atomic.AtomicReference;
+
+abstract class ViewHighlighter {
+  public static ViewHighlighter newInstance() {
+    // TODO: find ways to do highlighting on older versions too
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR2) {
+      return new OverlayHighlighter();
+    } else {
+      return new NullHighlighter();
```
jasta: This represents a poor API.  If you are not going to provide a working implementation for all versions, the `newInstance` method should be marked `@Nullable` and you should actually return null.  This way callers explicitly have to write:
```java
if (mHighlighter != null) {
  mHighlighter.highlight(...)
}
```

An alternative strategy is to add an `isSupported` method that `NullHighlighter` (prefer `NoopHighlighter` as well since null has no semantic context here) implements and returns false, then throws on all other methods, forcing callers to write:
```java
if (mHighlighter.isSupported()) {
  mHighlighter.highlight(...);
}
```

The reason for being explicit is so that readers of the highlighting code can see plainly that it might not be supported and that would explain why it's not working for them on a naive/quick read of the source code.  Forcing folks to dig all the way down into the implementation to see that possibility is frustrating.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewHighlighter.java
```diff
@@ -0,0 +1,112 @@
...
+        hideHighlightOnUiThread();
+      }
+    };
+
+    public OverlayHighlighter() {
+      mHighlightDrawable = new ColorDrawable(0xffff0000);
+      mHighlightDrawable.setAlpha(128);
+    }
+
+    public void highlightView(final View view) {
```
jasta: nit: `final` is not needed

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewHighlighter.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    private void highlightViewOnUiThread() {
+      hideHighlightOnUiThread();
+
+      // TODO: this weird scheme is meant to coalesce a flurry of highlight requests so that we
+      //       don't get a visible lag whereby the highlight keeps moving through elements 1 through
+      //       N even though the request for N came in a long time ago. it doesn't seem to work for
+      //       some reason that I can't yet figure out
+      while (true) {
```
jasta: There's a much simpler solution here:
```java
mHandler = new Handler(Looper.getMainLooper());

synchronized void highlightView(...) {
  mHandle.removeCallbacks(mHighlightOnUiThreadRunnable);
  mHandle.postDelayed(mHighlight..., 50 /* millis */);
}
```

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ApplicationDescriptor.java
```diff
@@ -0,0 +1,155 @@
...
+
+    public void hook(Application element) {
+      mElement = Util.throwIfNull(element);
+      mActivities = new ArrayList<>();
+
+      // TODO: tree diffing will remove the need to even worry about installing this callback,
+      //       which will then allow ~realtime updates for pre-ICS. for now, pre-ICS will just
+      //       have to close and re-open the inspector whenever the activity changes.
+      //       (this does assume that the hack below for getAllActivitiesHack() works on pre-ICS)
+      if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.ICE_CREAM_SANDWICH) {
```
jasta: Never write this imperatively without a class guarding the implementation.  Doing so causes significant performance problems when the verifier runs on this code -- it sprays logcat with warnings about having to rewrite opcodes on the fly.

Always produce a compat class using the pattern you have for `ViewHighlighter` which is designed to isolate the verifier from the illegal code on older versions.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ApplicationDescriptor.java
```diff
@@ -0,0 +1,155 @@
...
+            getListener().onChildRemoved(mElement, activity);
+          }
+        };
+
+        mElement.registerActivityLifecycleCallbacks(mCallbacks);
+      }
+
+      try {
+        mActivities = getAllActivitiesHack();
+      } catch (Exception ex) {
```
jasta: This should be encapsulated in the `...Hack` method.  Asking callers to deal with this is crazy.  For instance, what if there's an NPE in `getAllActivitiesHack`.  Do you really want the caller to have to figure that out?

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -0,0 +1,12 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+package com.facebook.stetho.inspector.elements.android;
+
+import android.view.View;
+
+import com.facebook.stetho.inspector.elements.ChainedDescriptor;
+
+final class ViewDescriptor extends ChainedDescriptor<View> {
+  // TODO: all sorts of cool stuff
```
jasta: ...? :)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewGroupDescriptor.java
```diff
@@ -0,0 +1,164 @@
...
+final class ViewGroupDescriptor extends ChainedDescriptor<ViewGroup> {
+  // TODO: We're probably going to switch to another way of determining structural
+  //       changes in the View tree. So this gizmo with chaining OnHierarchyChangeListener
+  //       via reflection should go away soon.
+
+  private static final Field sOnHierarchyChangeListenerField;
+  static {
+    Field field;
+    try {
+      field = ViewGroup.class.getDeclaredField("mOnHierarchyChangeListener");
```
jasta: Ohh I don't like this at all.... :(  Strongly prefer using `ViewTreeObserver` and doing diffing per our previous discussion.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewGroupDescriptor.java
```diff
@@ -0,0 +1,164 @@
...
+    private ViewGroup mElement;
+    private ViewGroup.OnHierarchyChangeListener mInnerListener;
+
+    public void hook(ViewGroup element) {
+      mElement = Util.throwIfNull(element);
+
+      ViewGroup.OnHierarchyChangeListener innerListener;
+      try {
+        innerListener = getOnHierarchyChangeListenerHack(mElement);
+      } catch (NoSuchFieldException ex) {
```
jasta: again, you don't want your hack to throw stuff like this out.  Instead have a two stage method call structure:
```java
innerListener = HierarchyChangeListenerUtil.isSupported()  HierarchyChangeListenerUtil.getBlaBla(mElement) : null;
```

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/WindowDescriptor.java
```diff
@@ -0,0 +1,220 @@
...
+  protected Object onGetChildAt(Window element, int index) {
+    View decorView = element.peekDecorView();
+    if (decorView == null) {
+      throw new IndexOutOfBoundsException();
+    } else {
+      return decorView;
+    }
+  }
+
+  // TODO: We're probably going to switch to another way of determining structural
```
jasta: Glossing over this.... :)

stetho/src/main/java/com/facebook/stetho/inspector/helper/ObjectIdMapper.java
```diff
@@ -0,0 +1,79 @@
...
+package com.facebook.stetho.inspector.helper;
+
+import android.util.SparseArray;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import javax.annotation.concurrent.GuardedBy;
+
+public class ObjectIdMapper {
```
jasta: nit: `ObjectIdMapper<T>` can improve convenience and generalization.

jasta: This is a good candidate for unit testing...

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Runtime.java
```diff
@@ -9,6 +9,7 @@
 
 import com.facebook.stetho.json.annotation.JsonProperty;
 import com.facebook.stetho.json.annotation.JsonValue;
+
```
jasta: nit: remove

jasta:
Can you also take a look at why Travis CI thinks this is failing?  Once we're passing the build, I think we'll be able to start merging into the elements branch.  My thought is that once we're happy with the end state we can rebase the elements branch a little and then merge the result into master.

stetho/src/main/java/com/facebook/stetho/inspector/helper/ObjectIdMapper.java
```diff
@@ -0,0 +1,79 @@
...
+  protected final Object mSync = new Object();
+
+  @GuardedBy("mSync")
+  private int mNextId = 1;
+
+  @GuardedBy("mSync")
+  private final Map<Object, Integer> mObjectToIdMap = new HashMap<Object, Integer>();
```
jasta: This should be an `IdentityHashMap`.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewGroupDescriptor.java
rickbrew: I'm quite loathe to do this as well (the reflection hax0ring, that is). How about we leave as-is for now and yank it once diffing is in place? (which I'm definitely not treating as a low priority thing btw)

stetho/src/main/java/com/facebook/stetho/inspector/elements/ObjectDescriptor.java
rickbrew: Ahh, that makes sense. I must've mentally slipped back into C#-mode where every string is its own allocation without any sharing (I can explain why it's done that way if you ever want to know, btw).

stetho/src/main/java/com/facebook/stetho/inspector/helper/ObjectIdMapper.java
rickbrew: Do we want it to be generic? I didn't need it to be generic right now, so seemed kind of pointless for now, and trivial cost to change later ...

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewGroupDescriptor.java
jasta: I can live with that :)

stetho/src/main/java/com/facebook/stetho/inspector/helper/ObjectIdMapper.java
jasta: Perhaps you're right.  Even the Console tab use case will treat it as `Object`.  Nevermind :)

stetho/src/main/java/com/facebook/stetho/inspector/elements/Descriptor.java
```diff
@@ -0,0 +1,45 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+package com.facebook.stetho.inspector.elements;
+
+import javax.annotation.Nullable;
+
+public abstract class Descriptor implements NodeDescriptor {
+
+  public interface Listener {
```
jasta: here too re order

stetho/src/main/java/com/facebook/stetho/inspector/elements/Descriptor.java
```diff
@@ -0,0 +1,45 @@
...
+
+  void setListener(Listener listener) {
+    mListener = listener;
+  }
+
+  protected final Listener getListener() {
+    return mListener;
+  }
+
+  // TODO: which place to put these?
```
jasta: this is not relevant

stetho/src/main/java/com/facebook/stetho/inspector/elements/DescriptorMap.java
```diff
@@ -0,0 +1,84 @@
...
+
+package com.facebook.stetho.inspector.elements;
+
+import com.facebook.stetho.common.Util;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public final class DescriptorMap {
+  private final Map<Class<?>, Descriptor> mMap = new HashMap<Class<?>, Descriptor>(); // TODO: ArrayMap
```
jasta: This isn't a TODO really.  We shouldn't use the support library so it's not really an option.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ApplicationDescriptor.java
```diff
@@ -0,0 +1,198 @@
...
+      Field activityField = activityRecordClass.getDeclaredField("activity");
+      activityField.setAccessible(true);
+      Activity activity = (Activity)activityField.get(activityRecord);
+      activitiesList.add(activity);
+    }
+
+    return activitiesList;
+  }
+
+  private ElementContext newElementContext() {
```
jasta: Separate class please.  Try to be discrete with class responsibility, especially when hacks or backwards compatibility is involved.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/TextViewDescriptor.java
```diff
@@ -0,0 +1,87 @@
...
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+final class TextViewDescriptor extends ChainedDescriptor<TextView> {
+  private static final String TEXT_ATTRIBUTE_NAME = "text";
+
+  private final Map<TextView, ElementContext> mElementToContextMap =
+      Collections.synchronizedMap(new HashMap<TextView, ElementContext>());
```
jasta: A lot of these maps you're using actually should be `IdentityHashMap`.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewHighlighter.java
```diff
@@ -0,0 +1,137 @@
...
+          Util.throwIfNull(view),
+          contentColor,
+          paddingColor,
+          borderColor,
+          marginColor);
+    }
+
+    private void setHighlightedViewImpl(
+        @Nullable View view,
+        int contentColor,
```
jasta: Probably want a container class for this.

jasta:
This is looking great.  I think it's ready to be merged, but I'd like to do some rebasing first to clean up the diff stack.  Ideally this first check-in should expressed as a single commit.

jasta:
Looks good!

PR #85, opened 2015-03-10
Fix to avoid NPE in DOM.ProviderListener.onChildRemoved

### LINKED ISSUE #83: Elements tab crashes with NPE in DOM.ProviderListener.onChildRemoved()
In onChildRemoved, we try to get the nodeId for an element that isn't mapped. We end up with a null Integer and then blindly try to store it into Node's nodeId field (an int), and immediately NPE.

This happened several times, but the stack trace has long since scrolled off into oblivion and I can't repro it anymore. For now I've got a fix that logs a warning and skips sending the DOM.childNodeRemoved event over the wire.

rickbrew:
This should fix the NPE. I also fixed up my LogUtil usage in nearby code.

Closes #83

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/common/android/ApplicationUtil.java
+++ b/stetho/src/main/java/com/facebook/stetho/common/android/ApplicationUtil.java
@@ -14,8 +14,6 @@ import java.util.List;
 import java.util.Map;
 
 public class ApplicationUtil {
-  private static final String TAG = "ApplicationUtil";
-
   private ApplicationUtil() {
   }
 
@@ -23,7 +21,7 @@ public class ApplicationUtil {
     try {
       return getAllActivitiesHack();
     } catch (Exception e) {
-      LogUtil.w(TAG, e, "Cannot retrieve list of Activity instances. UI inspection may not work!");
+      LogUtil.w(e, "Cannot retrieve list of Activity instances. UI inspection may not work!");
       return Collections.emptyList();
     }
   }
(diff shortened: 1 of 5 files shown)


stetho/src/main/java/com/facebook/stetho/common/android/ApplicationUtil.java
```diff
@@ -23,7 +21,7 @@ private ApplicationUtil() {
     try {
       return getAllActivitiesHack();
     } catch (Exception e) {
-      LogUtil.w(TAG, e, "Cannot retrieve list of Activity instances. UI inspection may not work!");
+      LogUtil.w(e, "Cannot retrieve list of Activity instances. UI inspection may not work!");
```
jasta: Oops :)

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -221,10 +221,23 @@ public void onChildInserted(Object parentElement, Object previousElement, Object
...
-      ChildNodeRemovedEvent message = new ChildNodeRemovedEvent();
-      message.parentNodeId = mObjectIdMapper.getIdForObject(parentElement);
-      message.nodeId = mObjectIdMapper.getIdForObject(childElement);
-      mPeerManager.sendNotificationToPeers("DOM.childNodeRemoved", message);
+      Integer parentNodeId = mObjectIdMapper.getIdForObject(parentElement);
+      Integer childNodeId = mObjectIdMapper.getIdForObject(childElement);
+
+      if (parentNodeId == null || childNodeId == null) {
+        LogUtil.w(
+            "DOM.onChildRemoved() called for a non-mapped node: "
```
jasta: Does this spam in practice?  I want to avoid confusing folks with internal error messages.

Specifically, if this is expected and/or fires often, consider using debug or info level.

jasta:
Looks good other than these minor issues.  Please address, rebase, and merge.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: No spam. Also I think I found a reliable repro

PR #89, opened 2015-03-11
Adding id attribute to Views

### LINKED ISSUE #86: Add 'id' attribute to ViewDescriptor
If the layout XML looks like this:

> `<Button android:id="@+id/apod_btn"` ...

Then we should show it similarly in the elements tab, e.g.

> `<button id="@+id/apod_btn"` ... 

View.toString() already does this, so adapting from that should be straightforward.

rickbrew:
The 'id' attribute will now show up in a way that's recognizable from XML layouts, e.g. `id="@id/ima_button"`

I also refactored the protocol for retrieving attributes via NodeDescriptor to reduce allocations and CPU usage. I found that I was having to determine the contents of the attributes twice, once for getAttributeCount() to determine its existence, and then a second time in copyAttributeAt(). Since we currently only need to retrieve all the attributes at once, it makes sense to switch to an accumulator which can pack the name/value pairs directly into the interleaved String array that the JSON protocol uses. And the code is a lot simpler.

screen shot 2015-03-11 at 4 34 58 pm

Closes #86

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
@@ -2,12 +2,19 @@
 
 package com.facebook.stetho.inspector.elements.android;
 
+import android.content.res.Resources;
 import android.view.View;
 
+import com.facebook.stetho.common.LogUtil;
 import com.facebook.stetho.common.StringUtil;
+import com.facebook.stetho.inspector.elements.AttributeAccumulator;
 import com.facebook.stetho.inspector.elements.ChainedDescriptor;
 
+import javax.annotation.Nullable;
+
 final class ViewDescriptor extends ChainedDescriptor<View> {
+  private static final String ID_ATTRIBUTE_NAME = "id";
+
   @Override
   protected String onGetNodeName(View element) {
     String className = element.getClass().getName();
@@ -16,5 +23,69 @@ final class ViewDescriptor extends ChainedDescriptor<View> {
         StringUtil.removePrefix(className, "android.view.",
         StringUtil.removePrefix(className, "android.widget."));
   }
-}
 
+  @Override
+  protected void onCopyAttributes(View element, AttributeAccumulator attributes) {
+    String id = getIdAttribute(element);
+    if (id != null) {
+      attributes.add(ID_ATTRIBUTE_NAME, id);
+    }
+  }
+
+  private static int getResourcePackageId(int id) {
+    return (id >>> 24) & 0xff;
+  }
+
+  @Nullable
+  private static String getIdAttribute(View element) {
+    // Adapted from View.toString()
+
+    int id = element.getId();
+    if (id == View.NO_ID) {
+      return null;
+    }
+
+    String idString = null;
+    Resources r = element.getResources();
+    if (r != null) {
+      try {
+        String prefix;
+        String prefixSeparator;
+        switch (getResourcePackageId(id)) {
+          case 0x7f:
+            prefix = "";
+            prefixSeparator = "";
+            break;
+          default:
+            prefix = r.getResourcePackageName(id);
+            prefixSeparator = ":";
+            break;
+        }
+
+        String typeName = r.getResourceTypeName(id);
+        String entryName = r.getResourceEntryName(id);
+
+        StringBuilder sb = new StringBuilder(
+            1 + prefix.length() + prefixSeparator.length() +
+            typeName.length() + 1 + entryName.length());
+        sb.append("@");
+        sb.append(prefix);
+        sb.append(prefixSeparator);
+        sb.append(typeName);
+        sb.append("/");
+        sb.append(entryName);
+
+        idString = sb.toString();
+
+      } catch (Resources.NotFoundException e) {
+        LogUtil.w(e, "Could not retrieve resource info for element: %s", element);
+      }
+    }
+
+    if (idString == null) {
+      idString = "#" + Integer.toHexString(id);
+    }
+
+    return idString;
+  }
+}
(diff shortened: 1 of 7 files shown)


jasta:
Please attach a screenshot to the pull request.  You can edit the pull request summary and drag-n-drop an image into it.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -16,5 +22,76 @@ protected String onGetNodeName(View element) {
...
+          sb.append(prefix);
+          sb.append(prefixSeparator);
+          sb.append(typeName);
+          sb.append("/");
+          sb.append(entryName);
+
+          idString = sb.toString();
+
+        } catch (Resources.NotFoundException e) {
+        }
```
jasta: `LogUtil` -- I don't know what this means in practice; if it ever happens, I want more info :)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -16,5 +22,76 @@ protected String onGetNodeName(View element) {
...
+    }
+
+    String idString = null;
+
+    if (resourceHasPackage(id)) {
+      Resources r = element.getResources();
+      if (r != null) {
+        try {
+          String prefix;
+          String prefixSeparator;
```
jasta: Seems like this can be simplified as just prefix, with the concatenation happening in each case statement, no?

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -16,5 +22,76 @@ protected String onGetNodeName(View element) {
...
+  @Override
+  protected void onCopyAttributes(View element, AttributeAccumulator attributes) {
+    String id = getIdAttribute(element);
+    if (id != null) {
+      attributes.add(ID_ATTRIBUTE_NAME, id);
+    }
+  }
+
+  // Copied from Resources.resourceHasPackage because it's marked as @hide
+  private static boolean resourceHasPackage(int resid) {
```
jasta: I really have no idea why this check exists in `View` itself.  This looks a bit like unnecessary legacy.  Specifically, even if the package id is 0x00, `Resource#getResourcePackageName` should still work.  I'd prefer to drop it and take a crash/bad result if I'm wrong :)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -16,5 +22,76 @@ protected String onGetNodeName(View element) {
...
+            default:
+              prefix = r.getResourcePackageName(id);
+              prefixSeparator = ":";
+              break;
+          }
+
+          String typeName = r.getResourceTypeName(id);
+          String entryName = r.getResourceEntryName(id);
+
+          // "@" + prefix + prefixSeparator + typeName + "/" + entryName
```
jasta: This comment seems a bit unnecessary.  It's easy code to read :)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -16,5 +22,76 @@ protected String onGetNodeName(View element) {
...
+        try {
+          String prefix;
+          String prefixSeparator;
+          switch (id & 0xff000000) {
+            case 0x7f000000:
+              prefix = "";
+              prefixSeparator = "";
+              break;
+            case 0x01000000:
+              prefix = "android";
```
jasta: This special case statement is probably not needed.  I bet `r.getResourcePackageName` will yield android.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -16,5 +22,76 @@ protected String onGetNodeName(View element) {
...
+
+    String idString = null;
+
+    if (resourceHasPackage(id)) {
+      Resources r = element.getResources();
+      if (r != null) {
+        try {
+          String prefix;
+          String prefixSeparator;
+          switch (id & 0xff000000) {
```
jasta: This reads better as:
```java
switch (resourceGetPackageId(id)) {
  case 0x7f: ...
}

private static int resourceGetPackageId(int id) {
  return (id >>> 24) & 0xff;
}
```

As an extra bonus, `resourceHasPackage` can be inlined as: `if (resourceGetPackageId(id) != 0) {`...

jasta:
Minor cleanup requested, then feel free to merge.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
rickbrew: Trying to be prudent with string allocations 8-) If I do the concatenation within the case, then I can't know precisely how large to allocate the StringBuilder a few lines down.

jasta: I try to be pragmatic about this kind of thing.  For instance, the relative cost of this allocation is dwarfed by the relatively high cost of traversing the `Resources` system looking for string identifiers.  Your call though :)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
rickbrew: Tried skipping the check and running on a few different apps ... seems to work, why not?

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
rickbrew: Why the name "resourceGetPackageId" ? (versus "getResourcePackageId")

jasta: No reason, your `resourceHasPackage` method scrambled my better judgment :)

Issue #82, opened 2015-03-10
Feedback on non SELECT database queries.


Tolriq:
Currently there's no feedback when doing update or delete queries.

It would be great to have some to know that it worked as even the query vanish from the interface.

Maybe use the current error system 

```
delete from albums where tt = false;
no such column: tt (code 1): , while compiling: delete from albums where tt = false;
```

With the error line containing : Success : Last id = XX or affected rows = YY

Or return a dummy cursor with only 2 columns last id , affected rows and 1 line with data.

jasta:
I just looked into this briefly and noticed that of course you can indeed return results as if it were a SQL query.  I think that's probably the right thing to do.

Patches welcome? :)  You'd be looking at `Database` and `DatabasePeerManager`.

Tolriq:
If no one look into that before I have time, I'll try :)

But I must admit I'm in the middle of a big rewrite and try to gain time with tools, so hard to find time soon to learn new product to contribute.

jasta:
@Tolriq I'll see what I can do :)

Tolriq:
No hurry the tool is already wonderful right now and make me win lot's of time for tests and debug.

So I can check that when I have time after the database / sync rewrite :)

longinoa:
I looked into this a bit today. We are using `SqliteDatabase#rawQuery(...)` which returns a `Cursor`. As far as I can tell we cant extract the data of how many rows were modified from a `Cursor` and would need to find out if the query is `UPDATE`/`INSERT`/`DELETE`/`SELECT`/... by parsing the raw sql, and then calling structured methods on `SqlliteDatabase`. 

I know presto has parsing for sqlite but do we want to go down that path?

PR #90, opened 2015-03-11
support sending back rows changed

longinoa:
This solves #82. We use the prefix of the first 3 characters to determine what type of request it is. Since we only have access to "columns" and "error" text use the column to display the type and the value to display the count.
scalar-return

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/database/DatabasePeerManager.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/database/DatabasePeerManager.java
@@ -11,16 +11,19 @@ import java.io.File;
 import java.util.ArrayList;
 import java.util.List;
 
+import android.annotation.TargetApi;
 import android.content.Context;
 import android.database.Cursor;
 import android.database.sqlite.SQLiteDatabase;
 import android.database.sqlite.SQLiteException;
+import android.database.sqlite.SQLiteStatement;
 
 import com.facebook.stetho.common.Util;
 import com.facebook.stetho.inspector.helper.ChromePeerManager;
 import com.facebook.stetho.inspector.helper.PeerRegistrationListener;
 import com.facebook.stetho.inspector.jsonrpc.JsonRpcPeer;
 import com.facebook.stetho.inspector.protocol.module.Database;
+import com.facebook.stetho.inspector.protocol.module.DatabaseConstants;
 
 @ThreadSafe
 public class DatabasePeerManager extends ChromePeerManager {
@@ -129,17 +132,67 @@ public class DatabasePeerManager extends ChromePeerManager {
     Util.throwIfNull(handler);
     SQLiteDatabase database = openDatabase(databaseName);
     try {
-      Cursor cursor = database.rawQuery(query, null);
-      try {
-        return handler.handleResult(cursor);
-      } finally {
-        cursor.close();
+      String firstWord = getFirstWord(query);
+
+      if (firstWord.equalsIgnoreCase("UPDATE") || firstWord.equalsIgnoreCase("DELETE")) {
+        return executeUpdateDelete(database, query, handler);
+      } else if (firstWord.equalsIgnoreCase("INSERT")) {
+        return executeInsert(database, query, handler);
+      } else if (firstWord.equalsIgnoreCase("SELECT")) {
+        return executeSelect(database, query, handler);
+      } else {
+        return executeRawQuery(database, query, handler);
       }
+
     } finally {
       database.close();
     }
   }
 
+  private static String getFirstWord(String s) {
+    s.trim();
+    int firstSpace = s.indexOf(' ');
+    return firstSpace >= 0 ? s.substring(0, firstSpace) : s;
+  }
+
+  @TargetApi(DatabaseConstants.MIN_API_LEVEL)
+  private <T> T executeUpdateDelete(
+      SQLiteDatabase database,
+      String query,
+      ExecuteResultHandler<T> handler) {
+    SQLiteStatement statement = database.compileStatement(query);
+    int count = statement.executeUpdateDelete();
+    return handler.handleUpdateDelete(count);
+  }
+
+  private <T> T executeInsert(
+      SQLiteDatabase database,
+      String query,
+      ExecuteResultHandler<T> handler) {
+    SQLiteStatement statement = database.compileStatement(query);
+    long count = statement.executeInsert();
+    return handler.handleInsert(count);
+  }
+
+  private <T> T executeSelect(
+      SQLiteDatabase database,
+      String query,
+      ExecuteResultHandler<T> handler) {
+    Cursor cursor = database.rawQuery(query, null);
+    try {
+      return handler.handleSelect(cursor);
+    } finally {
+      cursor.close();
+    }
+  }
+
+  private <T> T executeRawQuery(
+      SQLiteDatabase database,
+      String query,
+      ExecuteResultHandler<T> handler) {
+    database.execSQL(query);
+    return handler.handleRawQuery();
+  }
   private SQLiteDatabase openDatabase(String databaseName) throws SQLiteException {
     Util.throwIfNull(databaseName);
     File databaseFile = mContext.getDatabasePath(databaseName);
@@ -151,7 +204,13 @@ public class DatabasePeerManager extends ChromePeerManager {
   }
 
   public interface ExecuteResultHandler<T> {
-    public T handleResult(Cursor result) throws SQLiteException;
+    public T handleRawQuery() throws SQLiteException;
+
+    public T handleSelect(Cursor result) throws SQLiteException;
+
+    public T handleInsert(long insertedId) throws SQLiteException;
+
+    public T handleUpdateDelete(int count) throws SQLiteException;
   }
 
   private final PeerRegistrationListener mPeerRegistrationListener =
--- a/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DatabaseConstants.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DatabaseConstants.java
@@ -1,0 +1,13 @@
+// Copyright 2015-present Facebook. All Rights Reserved.
+
+package com.facebook.stetho.inspector.protocol.module;
+
+import android.os.Build;
+
+public interface DatabaseConstants {
+
+  /**
+   * Minimum API version required to use the {@link Database}.
+   */
+  public static final int MIN_API_LEVEL = Build.VERSION_CODES.HONEYCOMB;
+}
--- a/stetho/src/main/java/com/facebook/stetho/Stetho.java
+++ b/stetho/src/main/java/com/facebook/stetho/Stetho.java
@@ -1,3 +1,5 @@
+// Copyright 2015-present Facebook. All Rights Reserved.
+
 package com.facebook.stetho;
 
 import com.facebook.stetho.inspector.database.DefaultDatabaseFilesProvider;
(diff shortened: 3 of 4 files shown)


stetho/src/main/java/com/facebook/stetho/inspector/database/DatabasePeerManager.java
```diff
@@ -86,7 +87,7 @@ private static String removeSuffix(String str, String[] suffixesToRemove) {
     SQLiteDatabase database = openDatabase(databaseName);
     try {
       Cursor cursor = database.rawQuery("SELECT name FROM sqlite_master WHERE type=?",
-          new String[] { "table" });
+          new String[]{"table"});
```
jasta: Please don't :)

stetho/src/main/java/com/facebook/stetho/inspector/database/DatabasePeerManager.java
```diff
@@ -107,17 +108,54 @@ private static String removeSuffix(String str, String[] suffixesToRemove) {
...
-      try {
-        return handler.handleResult(cursor);
-      } finally {
-        cursor.close();
+      String trimmedSql = query.trim();
+      String prefixSql = "";
+      if (trimmedSql.length() >= 3) {
+        prefixSql = trimmedSql.substring(0, 3).toUpperCase();
       }
+      if (prefixSql.equals("UPD") || prefixSql.equals("DEL")) {
```
jasta: This must be case insensitive.

stetho/src/main/java/com/facebook/stetho/inspector/database/DatabasePeerManager.java
```diff
@@ -107,17 +108,54 @@ private static String removeSuffix(String str, String[] suffixesToRemove) {
...
+    return handler.handleResult(count, "ID Of last inserted row");
+  }
+
+  private <T> T executeRawQuery(
+      SQLiteDatabase database,
+      String query,
+      ExecuteResultHandler<T> handler) {
+    Cursor cursor = database.rawQuery(query, null);
+    try {
+      return handler.handleResult(cursor);
```
jasta: Out of curiosity, what was this actually doing before when `UPDATE` et al were passed?  What was the cursor?

stetho/src/main/java/com/facebook/stetho/inspector/database/DatabasePeerManager.java
```diff
@@ -130,17 +168,19 @@ private SQLiteDatabase openDatabase(String databaseName) throws SQLiteException
...
+    public T handleResult(Long value, String name) throws SQLiteException;
   }
 
   private final PeerRegistrationListener mPeerRegistrationListener =
       new PeerRegistrationListener() {
-    @Override
-    public void onPeerRegistered(JsonRpcPeer peer) {
-      bootstrapNewPeer(peer);
-    }
+        @Override
```
jasta: Revert.  Tell IntelliJ to stop auto formatting :)

stetho/src/main/java/com/facebook/stetho/inspector/database/DatabasePeerManager.java
```diff
@@ -130,17 +168,19 @@ private SQLiteDatabase openDatabase(String databaseName) throws SQLiteException
 
   public interface ExecuteResultHandler<T> {
     public T handleResult(Cursor result) throws SQLiteException;
+
+    public T handleResult(Long value, String name) throws SQLiteException;
```
jasta: Rename and clarify:
```java
handleSelect(Cursor result);
handleInsert(long lastInsertedId);
handleUpdateDelete(int rowsAffected);
```

The mapping to the DevTools response and is the job of the implementer of this interface.

stetho/src/main/java/com/facebook/stetho/inspector/database/DatabasePeerManager.java
jasta: Also, curious why we don't spell out UPDATE, DELETE, etc?  Can you short-hand these in some way otherwise?

stetho/src/main/java/com/facebook/stetho/inspector/database/DatabasePeerManager.java
```diff
@@ -107,17 +108,54 @@ private static String removeSuffix(String str, String[] suffixesToRemove) {
     Util.throwIfNull(handler);
     SQLiteDatabase database = openDatabase(databaseName);
     try {
-      Cursor cursor = database.rawQuery(query, null);
-      try {
-        return handler.handleResult(cursor);
-      } finally {
-        cursor.close();
+      String trimmedSql = query.trim();
```
jasta: I think this can be done in a clearer way:
```java
String command = getFirstWord(query);
if ("UPDATE".equalsIgnoreCase(command)) { ... }

private static String getFirstWord(String s) {
  s.trim();
  int firstSpace = s.indexOf(' ');
  return firstSpace >= 0 ? s.substring(0, firstSpace) : s;
}
```

jasta:
Check the Travis build errors.  I think they're lint related, which I don't have a great way to extract from Travis-CI at the moment.  I just re-run` gradle check` and check out the XML file it generates.

stetho/src/main/java/com/facebook/stetho/inspector/database/DatabasePeerManager.java
longinoa: This is a copy from how `DatabaseUtils` does categorization.

stetho/src/main/java/com/facebook/stetho/inspector/database/DatabasePeerManager.java
longinoa: the cursor was basically null. the count was always -1, there were no columns, and no values. Plus "rows updated" always = 0

stetho/src/main/java/com/facebook/stetho/inspector/database/DatabasePeerManager.java
longinoa: This is case insensitive, note the call to `toUpperCase()`

stetho/src/main/java/com/facebook/stetho/inspector/database/DatabasePeerManager.java
longinoa: Maybe we should make it so that IntelliJ just does the right thing!

stetho/src/main/java/com/facebook/stetho/inspector/database/DatabasePeerManager.java
longinoa: BTW The reason I'm not directly using `DatabaseUtils#getSqlStatementType(...)` is that it doesnt give us quite enough information.

I agree that creating the trimmed string is probably a bit of overkill here, but i preferred the simpler approach to the most performant.

jasta: Re `toUpperCase`, I missed this but prefer `equalsIgnoreCase` on the full word.  It's better performing and IMO, clearer.  Also, not applicable in this case but `a.equalsIgnoreCase(b)` is not strictly equal to `a.toUpperCase().equals(b.toUpperCase())`.  Casing is defined very differently in a lot of languages.

These are just nits tho, no big deal.

stetho/src/main/java/com/facebook/stetho/inspector/database/DatabasePeerManager.java
jasta: I'm going to.  We have to for license reasons anyway.

stetho/src/main/java/com/facebook/stetho/inspector/database/DatabasePeerManager.java
jasta: Oops, I meant `s = s.trim()` :)

Tolriq:
Seems my answer on original post was lost or I forget to push the button:(

Why don't you just check for no colums or count -1 in resulting cursor ? (if cursor null then it's an error)
Since the first query should be run asynchronous there's no problem starting another one after in this thread.

If no column then start new simple query 'select changes(), last_insert_rowid()' and return a MaxtrixCursor with 1 line / 1 row containing the data. (Can handle one last case when no change no row id for create database for example).

This sounds cleaner and can handle more things like create XXX, drop, ... 
Even if less used those commands are used and should return something too.

longinoa:
@Tolriq  yeah I thought about doing that. 
You have to parse the type of query to find out if you should do `select_changes` or `last_insert_rowId`.
How do you dont know that it is a select w/ no rows returned?
What would you expect create/drop to return?

stetho/src/main/java/com/facebook/stetho/inspector/database/DatabasePeerManager.java
longinoa: done

Tolriq:
A select without rows have the columns filled, so does not trigger this.

From the other problem about database mode change, I suppose that the database is opened / closed on each query / access.

So changes and last insert are the values corresponding to previous query so rule would be in order :
- Cursor is null : Error
- Cursor have colums : return cursor
- Cursor have no colums : Do a "select changes(), last_insert_rowid()"
- If last insert  != 0 then return matrixCursor lastinsert and the id
- If changes != 0 then return matrixCursor affectedRows and the count
- If nothing then it's another command and since cursor is not null return a matrixCursor result and true.

This covers all cases and is simple to follow and maintain.

Of course this does not work if open / close assumption is wrong.

longinoa:
what would be returned for last_insert_id or select_changes for statements other than update/insert/delete? IE what is the return type for `create table` for instance? 

Im just trying to figure out the value in the return from them to do an additional round trip.
From what I have seen it looks like the inspector ui does bizarre things when you dont return a "valid" result with the sql query disappearing. I think we can tackle that by returning a dummy result but if there is value in the return for the other calls I am more than happy to add it in.

Tolriq:
They return 0 if no query have made changes or insert, that why my previous post talked about != 0 to handle all known cases.

And yes that why I suggested using MatrixCursor that are valid cursors so will be correctly handled by Stetho and Inspector UI.

longinoa:
I think if they are congruent changes doing the smart thing and handling the return in the correct way rather than a second round trip to the database.

stetho/src/main/java/com/facebook/stetho/inspector/database/DatabasePeerManager.java
```diff
@@ -129,7 +187,13 @@ private SQLiteDatabase openDatabase(String databaseName) throws SQLiteException
   }
 
   public interface ExecuteResultHandler<T> {
-    public T handleResult(Cursor result) throws SQLiteException;
+    public T handleRawQuery(Cursor result) throws SQLiteException;
```
jasta: This shouldn't be `Cursor`; it implies that the cursor came from `database.rawQuery`.  Instead, prefer to parse `changes(), last_insert_row_id()` into two longs (or whatever type they are?)

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DatabaseConstants.java
```diff
@@ -0,0 +1,13 @@
+// Copyright 2015-present Facebook. All Rights Reserved.
+
+package com.facebook.stetho.inspector.protocol.module;
+
+import android.os.Build;
+
+public class DatabaseConstants {
```
jasta: nit: `interface`

jasta:
Looks good except for nits.  Please update the commit message of the rebased diff to include the summary in the pull request and "Closes #..." at the bottom.  I recently learned that the pull request's metadata is not used for this feature, it must be in the commit itself.  In general it's better to have the meta data in the commit anyway so that it is visible in the git history later.

longinoa:
I was planning on squashing and was going to update the commit message to have `Closes #` at the end

Issue #88, opened 2015-03-11
Out of memory when doing a lot of request


fourlastor:
Hello, I was trying to inspect my network request while downloading a huge (150+) set of images for a grid in the app I work for. After scrolling fast enough Stetho threw an OOM.

I'm using Picasso to download the image and intercepting the network requests with StethoInterceptor:

```java
// standard Dagger module
    @Provides @Singleton OkHttpClient provideOkHttpClient() {
        return new OkHttpClient();
    }

    @Provides @Singleton Picasso providePicasso(@ForApplication Context context, OkHttpClient client) {
        return new Picasso.Builder(context).downloader(new OkHttpDownloader(client)).build();
    }


// overridden module for debug

    @Module(overrides = true, addsTo = AndroidModule.class, injects = PauseDebugApplication.class)
    static class DebugModule {
        @Provides @Singleton OkHttpClient provideDebugClient() {
            OkHttpClient client = new OkHttpClient();
            client.networkInterceptors().add(new StethoInterceptor());
            return client;
        }
    }
```

Here is the stacktrace:

```
03-11 12:49:48.675    4685-4906/fm.pause.debug E/AndroidRuntime﹕ FATAL EXCEPTION: StethoWorker30
    Process: fm.pause.debug, PID: 4685
    java.lang.OutOfMemoryError
            at java.lang.String.<init>(String.java:255)
            at java.lang.String.<init>(String.java:213)
            at java.io.ByteArrayOutputStream.toString(ByteArrayOutputStream.java:175)
            at com.facebook.stetho.inspector.network.ResponseBodyFileManager.readContentsAsUTF8(ResponseBodyFileManager.java:62)
            at com.facebook.stetho.inspector.network.ResponseBodyFileManager.readFile(ResponseBodyFileManager.java:52)
            at com.facebook.stetho.inspector.protocol.module.Network.readResponseBody(Network.java:69)
            at com.facebook.stetho.inspector.protocol.module.Network.getResponseBody(Network.java:54)
            at java.lang.reflect.Method.invokeNative(Native Method)
            at java.lang.reflect.Method.invoke(Method.java:515)
            at com.facebook.stetho.inspector.MethodDispatcher$MethodDispatchHelper.invoke(MethodDispatcher.java:89)
            at com.facebook.stetho.inspector.MethodDispatcher.dispatch(MethodDispatcher.java:60)
[...]

PR #93, opened 2015-03-14
Catch out of memory exceptions while inspecting download of large files

fourlastor:
This solves #88.

@jasta that should do (solution 1), right?

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/ChromeDevtoolsServer.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/ChromeDevtoolsServer.java
@@ -132,7 +132,16 @@ public class ChromeDevtoolsServer implements SimpleEndpoint {
       response.result = result;
       response.error = error;
       JSONObject jsonObject = mObjectMapper.convertValue(response, JSONObject.class);
-      String responseString = jsonObject.toString();
+      String responseString;
+      try {
+        responseString = jsonObject.toString();
+      } catch (OutOfMemoryError e) {
+        // JSONStringer can cause an OOM when the Json to handle is too big.
+        response.result = null;
+        response.error = mObjectMapper.convertValue(e.getMessage(), JSONObject.class);
+        jsonObject = mObjectMapper.convertValue(response, JSONObject.class);
+        responseString = jsonObject.toString();
+      }
       peer.getWebSocket().sendText(responseString);
     }
   }
--- a/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Network.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Network.java
@@ -64,9 +64,16 @@ public class Network implements ChromeDevtoolsDomain {
   }
 
   private GetResponseBodyResponse readResponseBody(String requestId)
-      throws IOException {
+      throws IOException, JsonRpcException {
     GetResponseBodyResponse response = new GetResponseBodyResponse();
-    ResponseBodyData bodyData = mResponseBodyFileManager.readFile(requestId);
+    ResponseBodyData bodyData;
+    try {
+      bodyData = mResponseBodyFileManager.readFile(requestId);
+    } catch (OutOfMemoryError e) {
+      throw new JsonRpcException(new JsonRpcError(JsonRpcError.ErrorCode.INTERNAL_ERROR,
+          e.toString(),
+          null /* data */));
+    }
     response.body = bodyData.data;
     response.base64Encoded = bodyData.base64Encoded;
     return response;
--- a/stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkEventReporterImpl.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkEventReporterImpl.java
@@ -110,7 +110,7 @@ public class NetworkEventReporterImpl implements NetworkEventReporter {
       if (body != null) {
         return new String(body, Utf8Charset.INSTANCE);
       }
-    } catch (IOException e) {
+    } catch (IOException | OutOfMemoryError e) {
       CLog.writeToConsole(
           peerManager,
           Console.MessageLevel.WARNING,


stetho/src/main/java/com/facebook/stetho/server/LocalSocketHttpServer.java
```diff
@@ -260,6 +260,8 @@ public void run() {
         LogUtil.w("I/O error: %s", ex);
       } catch (HttpException ex) {
         LogUtil.w("Unrecoverable HTTP protocol violation: %s", ex);
+      } catch (OutOfMemoryError ex) {
```
jasta: Way too high level of a catch.  Strongly recommend putting this in two critical places, `Network#readResponseBody` (specifically around the `readFile` call) and also `NetworkEventReporterImpl#readBodyAsString`.

I prefer this strategy because it is very specific about suppressing this serious error only in call sites we expect could accidentally allocate much larger segments than we normally expect.  No other code should be doing this.  Also, catching it appropriate lower levels allows us to respond using the actual DevTools protocol so that the UI can represent the error in a discoverable fashion.

jasta:
Also per https://help.github.com/articles/closing-issues-via-commit-messages/, please write "Closes #88" in the commit message, ideally on the last line by convention.

stetho/src/main/java/com/facebook/stetho/server/LocalSocketHttpServer.java
fourlastor: True, but I noticed that this is not always the case.
For example, to work on this, I was using the Stetho sample:

- Use Picasso to download an oversized image (I used http://upload.wikimedia.org/wikipedia/commons/2/24/Willaerts_Adam_The_Embarkation_of_the_Elector_Palantine_Oil_Canvas-huge.jpg), so it won't fail due to the image size in the ImageView, only for the network request.
- Use OkHttpClient with network interceptors to intercept Picasso's network request.

This is the code in `APODPostsAdapter#bindView`
```java
      OkHttpClient okclient = new OkHttpClient();
      okclient.networkInterceptors().add(new StethoInterceptor());

      Picasso picasso = new Picasso.Builder(APODActivity.this)
        .downloader(new OkHttpDownloader(okclient))
        .build();

      picasso.load("http://upload.wikimedia.org/wikipedia/commons/2/24/Willaerts_Adam_The_Embarkation_of_the_Elector_Palantine_Oil_Canvas-huge.jpg?" + cursor.getPosition())
          .fit()
          .centerCrop()
          .into(holder.image);
```

After doing that, this is the stack trace:
```
03-15 11:11:59.140    6755-7032/com.facebook.stetho.sample E/AndroidRuntime﹕ FATAL EXCEPTION: StethoWorker12
    Process: com.facebook.stetho.sample, PID: 6755
    java.lang.OutOfMemoryError
            at java.lang.AbstractStringBuilder.enlargeBuffer(AbstractStringBuilder.java:94)
            at java.lang.AbstractStringBuilder.append0(AbstractStringBuilder.java:132)
            at java.lang.StringBuilder.append(StringBuilder.java:124)
            at org.json.JSONStringer.string(JSONStringer.java:344)
            at org.json.JSONStringer.value(JSONStringer.java:252)
            at org.json.JSONObject.writeTo(JSONObject.java:672)
            at org.json.JSONStringer.value(JSONStringer.java:237)
            at org.json.JSONObject.writeTo(JSONObject.java:672)
            at org.json.JSONObject.toString(JSONObject.java:641)
            at com.facebook.stetho.inspector.ChromeDevtoolsServer.handleRemoteRequest(ChromeDevtoolsServer.java:135)
            at com.facebook.stetho.inspector.ChromeDevtoolsServer.handleRemoteMessage(ChromeDevtoolsServer.java:104)
            at com.facebook.stetho.inspector.ChromeDevtoolsServer.onMessage(ChromeDevtoolsServer.java:80)
            at com.facebook.stetho.websocket.WebSocketSession$1.handleTextFrame(WebSocketSession.java:169)
            at com.facebook.stetho.websocket.WebSocketSession$1.onCompleteFrame(WebSocketSession.java:129)
            at com.facebook.stetho.websocket.ReadHandler.readLoop(ReadHandler.java:39)
            at com.facebook.stetho.websocket.WebSocketSession.handle(WebSocketSession.java:38)
            at com.facebook.stetho.websocket.WebSocketHandler.doUpgrade(WebSocketHandler.java:147)
            at com.facebook.stetho.websocket.WebSocketHandler.handleSecured(WebSocketHandler.java:99)
            at com.facebook.stetho.server.SecureHttpRequestHandler.handle(SecureHttpRequestHandler.java:45)
            at org.apache.http.protocol.HttpService.doService(HttpService.java:243)
            at org.apache.http.protocol.HttpService.handleRequest(HttpService.java:187)
            at com.facebook.stetho.server.LocalSocketHttpServer$WorkerThread.run(LocalSocketHttpServer.java:255)
```

So maybe in `ChromeDevtoolsServer#handleRemoteRequest`?

jasta: That is unfortunate.  Given this, let's catch it generically as well as specially with a comment in the general path indicating that `JSONStringer` can cause the issue as well.  That said, `JSONStringer` sucks and we really should be using a custom `JsonWriter/Reader`-based approach.

fourlastor:
@jasta good for `ChromeDevToolsServer` and `NetworkEventReporterImpl`.

I'm not sure on what to do with `Network`, is there a standard way to communicate the failure with chrome dev tools? I guess throwing a `JsonRpcException` would do the trick but I'm not sure.

stetho/src/main/java/com/facebook/stetho/inspector/ChromeDevtoolsServer.java
```diff
@@ -132,7 +132,13 @@ private void handleRemoteRequest(JsonRpcPeer peer, JSONObject requestNode)
       response.result = result;
       response.error = error;
       JSONObject jsonObject = mObjectMapper.convertValue(response, JSONObject.class);
-      String responseString = jsonObject.toString();
+      String responseString;
+      try {
+        responseString = jsonObject.toString();
+      } catch (OutOfMemoryError e) {
+        // JSONStringer can cause an OOM when the Json to handle is too big.
+        throw new MessageHandlingException(e);
```
jasta: This will cause us to close the socket, dismissing the inspector window.  Instead, convert the response yourself to an error (similar to how the `catch (JsonRpcException)` clause a few lines up does it)

jasta:
Yes I think `Network` should use a `JsonRpcException`.  This will cause it to show up in the Console as an error.

fourlastor:
Great, done. Thank you a lot for being so patient and guiding me through it :)

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Network.java
```diff
@@ -64,9 +64,16 @@ public JsonRpcResult getResponseBody(JsonRpcPeer peer, JSONObject params)
...
   private GetResponseBodyResponse readResponseBody(String requestId)
-      throws IOException {
+      throws IOException, JsonRpcException {
     GetResponseBodyResponse response = new GetResponseBodyResponse();
-    ResponseBodyData bodyData = mResponseBodyFileManager.readFile(requestId);
+    ResponseBodyData bodyData;
+    try {
+      bodyData = mResponseBodyFileManager.readFile(requestId);
+    } catch (OutOfMemoryError e) {
+      throw new JsonRpcException(new JsonRpcError(JsonRpcError.ErrorCode.PARSER_ERROR,
```
jasta: Use `ErrorCode.INTERNAL_ERROR`

jasta:
Looks good except the one minor issue about which `ErrorCode` you're using.  Feel free to squash/rebase into a single commit, then I can merge.

jasta:
Awesome, thanks!

PR #95, opened 2015-03-17
Upgrade latest gradle and sdk versions. Fix build error because of:   classpath 'com.github.jcandksolutions.gradle:android-unit-test:2.1.1'

IgorGanapolsky:
Upgrade latest gradle and sdk versions.
Fix build error because of:
  classpath 'com.github.jcandksolutions.gradle:android-unit-test:2.1.1'

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 *.iml
 .DS_Store
 /build
+/.project
--- a/build.gradle
+++ b/build.gradle
@@ -3,8 +3,8 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.0.0'
-        classpath 'com.github.jcandksolutions.gradle:android-unit-test:2.1.1'
+        classpath 'com.android.tools.build:gradle:1.1.3'
+//        classpath 'com.github.jcandksolutions.gradle:android-unit-test:2.1.1'
     }
 }
 
--- a/stetho-okhttp/build.gradle
+++ b/stetho-okhttp/build.gradle
@@ -1,12 +1,12 @@
 apply plugin: 'com.android.library'
 
 android {
-    compileSdkVersion 21
-    buildToolsVersion "21.1.2"
+    compileSdkVersion 22
+    buildToolsVersion "22.0.0"
 
     defaultConfig {
         minSdkVersion 9
-        targetSdkVersion 21
+        targetSdkVersion 22
         versionCode 1
         versionName "1.0"
         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
@@ -20,7 +20,7 @@ android {
     }
 }
 
-apply plugin: 'android-unit-test'
+//apply plugin: 'android-unit-test'
 
 dependencies {
     compile project(':stetho')
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,8 @@ language: android
 
 android:
   components:
-    - build-tools 21.1.2
-    - android-21
+    - build-tools 22.0.0
+    - android-22
 
 branches:
   except:
--- a/stetho-sample/build.gradle
+++ b/stetho-sample/build.gradle
@@ -1,13 +1,13 @@
 apply plugin: 'com.android.application'
 
 android {
-    compileSdkVersion 21
-    buildToolsVersion "21.1.2"
+    compileSdkVersion 22
+    buildToolsVersion "22.00"
 
     defaultConfig {
         applicationId "com.facebook.stetho.sample"
         minSdkVersion 11
-        targetSdkVersion 21
+        targetSdkVersion 22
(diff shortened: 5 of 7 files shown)


jasta:
Why do we want this patch?  It upgrades to the latest build tools, but for what reason?  Also, it breaks our automated tests, which of course is a non-starter...

IgorGanapolsky:
Latest build tools and latest SDK is what many developers have installed on
their machines.  If you resist such changes and such progress, you are
setting your project up for friction with dev community.

Igor

jasta:
@IgorGanapolsky of course, that makes sense.  But we still need our tests working, of course :)  Any chance you can follow up as to why Robolectric testing is not working properly with the latest build tools?  AFAICT, they recommend using the plugin that you had to disable.

IgorGanapolsky:
Well, jcandksolutions shouldn't be used anymore since Gradle now supports
jvm testing in projects natively.  In fact, if you check jcandksolutions
say on his github repo that his project is deprecated.

Igor

.gitignore
```diff
@@ -4,3 +4,4 @@
 *.iml
 .DS_Store
 /build
+/.project
```
jasta: What is this for?

jasta:
Oh I didn't realize that was tested to work with Robolectric as well.  Looks like the Robolectric sample was updated accordingly as well: https://github.com/robolectric/deckard-gradle/blob/master/build.gradle

Though they don't use gradle 1.3.0 yet.  Can you please test that `gradle check connectedCheck` still runs all the tests?  Also note that Travis CI doesn't appear to have the latest build tools available for some reason.  We won't be able to merge without that being resolved.

build.gradle
```diff
@@ -3,8 +3,8 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.0.0'
-        classpath 'com.github.jcandksolutions.gradle:android-unit-test:2.1.1'
+        classpath 'com.android.tools.build:gradle:1.1.3'
+//        classpath 'com.github.jcandksolutions.gradle:android-unit-test:2.1.1'
```
jasta: If this is truly no longer needed, please just delete the line.

stetho-okhttp/build.gradle
```diff
@@ -20,7 +20,7 @@ android {
     }
 }
 
-apply plugin: 'android-unit-test'
+//apply plugin: 'android-unit-test'
```
jasta: Remove these as well.

.gitignore
IgorGanapolsky: As far as .gitignore file, that was added by my GitEye client
automatically.  I'll look into it.

fourlastor:
It might help to change the script entry in the travis config to `./gradlew test --continue`.
This runs the unit tests and it worked for my company to run CI on travis and the last gradle plugin.

More info here: http://tools.android.com/tech-docs/unit-testing-support

jasta:
@fourlastor without tests working in Travis I'm inclined to just wait for them to fix this upstream and then integrate your patch.  As of now it doesn't seem the value is worth the cost of a loss of tests in CI.

fourlastor:
What I'm saying is that you don't need to run the `connectedCheck` task to run the tests: stetho has only unit tests so the `test` task will do just fine.

The only test that doesn't pass is `ObjectMapperTest#testNestedProperty` because the `JSONObject#toString` method returns the keys in a different order, but the problem there is asserting equality on `JSONObject`s with the toString method.

All the suite of test (14 in the stetho project, 2 in the stetho-okhttp project) passed for me with this setup:

- Upgrade gradle plugin to 1.1.0
- Remove ` 'com.github.jcandksolutions.gradle:android-unit-test:2.1.1'` from the main build.gradle
- Remove `apply plugin: 'android-unit-test'` from the projects build.gradle

For the `JSONObject#toString` problem I suggest using an assertion library meant for it (such as https://github.com/skyscreamer/JSONassert).

jasta:
@fourlastor Ack'd.  Curious though why the removal of `android-unit-test` would break a flaky test that was previously working.  Doesn't matter though, we can just fix the test.

As for Travis CI, there remains the more serious issue of build tools 22.0.0 being apparently missing (check the Travis CI build output).  Until that passes, we cannot merge your change.

fourlastor:
It might be that it's not working on *my* machine, due to the JDK I'm using or whatever.

It's not my changes :P I'm not the PR author.

I would suggest to split this in 2 separate PR tho:
- unit test plugin removal (and bump of the Gradle plugin version to 1.1.x), which works and it's great to have
- bump of build tools to 22

fourlastor:
Travis CI is going to update their environments tomorrow, including Android SDK 22:

http://docs.travis-ci.com/user/build-environment-updates/2015-04-09/

jasta:
This is being divided and worked on as separate PRs.

PR #98, opened 2015-03-20
Support for Fragments as parents of their associated View

### LINKED ISSUE #97: Add Fragments support to elements tab
It'll be very useful if Fragments can show up in the elements tree. They should be emitted as the parent of their View.

rickbrew:
This adds support for Fragments in the element tree. Fragments will show up as the parent of the View that they contain, and therefore as the child of the ViewGroup which contains the Fragment's View. 

`ViewGroup -> Fragment -> View`

ViewGroupDescriptor does the high-level work of substituting Fragments for Views, and maintains a cache for mapping from Views to Fragments. Looking up the Fragment for a View is a bit expensive since it requires searching through all of the Fragments (see `FragmentApiUtil.findViewForFragment()`).

Closes #97

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/common/android/FragmentApi.java
+++ b/stetho/src/main/java/com/facebook/stetho/common/android/FragmentApi.java
@@ -1,0 +1,239 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+package com.facebook.stetho.common.android;
+
+import android.app.Activity;
+import android.view.View;
+
+import com.facebook.stetho.common.ReflectionUtil;
+import com.facebook.stetho.common.LogUtil;
+import com.facebook.stetho.common.Util;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.util.List;
+
+import javax.annotation.Nullable;
+
+// TODO: Don't use reflection to implement the accessors. We cannot express an optional dependency
+//       on the v4 support library with our current maven plugin for Gradle (need to upgrade to
+//       maven-publish?).
+//       Having 2 accessor implementations, one for the support API and another for the native API,
+//       will yield better performance and much cleaner code.
+//       (We will still need to use reflection to access fields, e.g. FragmentManagerImpl.mAdded)
+
+public final class FragmentApi {
+  private static final Class<?> sSupportFragmentClass;
+  private static final Class<?> sSupportFragmentActivityClass;
+  private static final Class<?> sSupportFragmentManagerClass;
+
+  private static final FragmentAccessor sSupportFragmentAccessor;
+  private static final FragmentActivityAccessor sSupportFragmentActivityAccessor;
+  private static final FragmentManagerAccessor sSupportFragmentManagerAccessor;
+
+  private static final Class<?> sFragmentClass;
+  private static final Class<?> sActivityClass;
+  private static final Class<?> sFragmentManagerClass;
+
+  private static final FragmentAccessor sFragmentAccessor;
+  private static final FragmentActivityAccessor sActivityAccessor;
+  private static final FragmentManagerAccessor sFragmentManagerAccessor;
+
+  static {
+    sSupportFragmentClass = ReflectionUtil.tryGetClassForName("android.support.v4.app.Fragment");
+    sSupportFragmentAccessor = (sSupportFragmentClass != null)
+        ? new ReflectionFragmentAccessor(sSupportFragmentClass)
+        : null;
+
+    sSupportFragmentActivityClass = ReflectionUtil.tryGetClassForName(
+        "android.support.v4.app.FragmentActivity");
+    sSupportFragmentActivityAccessor = (sSupportFragmentActivityClass != null)
+        ? new ReflectionFragmentActivityAccessor(sSupportFragmentActivityClass)
+        : null;
+
+    sSupportFragmentManagerClass = ReflectionUtil.tryGetClassForName(
+        "android.support.v4.app.FragmentManagerImpl");
+    sSupportFragmentManagerAccessor = (sSupportFragmentManagerClass != null)
+        ? new ReflectionFragmentManagerAccessor(sSupportFragmentManagerClass)
+        : null;
+
+    sFragmentClass = ReflectionUtil.tryGetClassForName("android.app.Fragment");
+    sFragmentAccessor = (sFragmentClass != null)
+        ? new ReflectionFragmentAccessor(sFragmentClass)
+        : null;
+
+    sActivityClass = ReflectionUtil.tryGetClassForName("android.app.Activity");
+    sActivityAccessor = (sActivityClass != null)
+        ? new ReflectionFragmentActivityAccessor(sActivityClass)
+        : null;
+
+    sFragmentManagerClass = ReflectionUtil.tryGetClassForName("android.app.FragmentManagerImpl");
+    sFragmentManagerAccessor = (sFragmentManagerClass != null)
+        ? new ReflectionFragmentManagerAccessor(sFragmentManagerClass)
+        : null;
+  }
+
+  @Nullable
+  public static Class<?> tryGetFragmentClass() {
+    return sFragmentClass;
+  }
+
+  @Nullable
+  public static Class<?> tryGetSupportFragmentClass() {
+    return sSupportFragmentClass;
+  }
+
+  public static FragmentAccessor getFragmentAccessorFor(Object fragment) {
+    Util.throwIfNull(fragment);
+
+    if (sSupportFragmentClass != null &&
+        sSupportFragmentClass.isAssignableFrom(fragment.getClass())) {
+      return sSupportFragmentAccessor;
+    }
+
+    if (sFragmentClass != null &&
+        sFragmentClass.isAssignableFrom(fragment.getClass())) {
+      return sFragmentAccessor;
+    }
+
+    throw new IllegalArgumentException();
+  }
+
+  @Nullable
+  public static FragmentActivityAccessor tryGetFragmentActivityAccessorFor(Object fragmentActivity) {
+    Util.throwIfNull(fragmentActivity);
+
+    if (sSupportFragmentActivityClass != null &&
+        sSupportFragmentActivityClass.isAssignableFrom(fragmentActivity.getClass())) {
+      return sSupportFragmentActivityAccessor;
+    }
+
+    if (sActivityClass != null &&
+        sActivityClass.isAssignableFrom(fragmentActivity.getClass())) {
+      return sActivityAccessor;
+    }
+
+    return null;
+  }
+
+  public static FragmentManagerAccessor getFragmentManagerAccessorFor(Object fragmentManager) {
+    if (sSupportFragmentManagerClass != null &&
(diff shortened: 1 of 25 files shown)


stetho-okhttp/src/main/java/com/facebook/stetho/okhttp/StethoInterceptor.java
```diff
@@ -262,7 +263,13 @@ public MediaType contentType() {
 
     @Override
     public long contentLength() {
-      return mBody.contentLength();
+      // TODO: apparently there was a change in okhttp
```
jasta: Rebase, this will not be needed anymore.

stetho/src/main/java/com/facebook/stetho/common/android/FragmentApi.java
```diff
@@ -0,0 +1,459 @@
...
+import com.facebook.stetho.common.Util;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.List;
+
+import javax.annotation.Nullable;
+
+// TODO: don't use reflection to implement the accessors
```
jasta: Please provide a little more context.  Specifically, there are two issues at play here:

1. The support library is needed as an optional/provided dependency, but we cannot express that today with the `maven` plugin in Gradle (I believe we need to upgrade to `maven-publish`.
2. The class needs to be divided significantly so that the Honeycomb+ APIs can be separated in terms of the class verifier from the Support library APIs.  Once separated, we can use booleans that use `Class.forName` to detect the presence of the classes we're interested in and adjust behaviour accordingly.

That said, I'm fine proceeding as-is per our conversation offline.

jasta:
Can you provide a screenshot as well?  Visual changes like this are very useful to see in context.

stetho/src/main/java/com/facebook/stetho/common/android/FragmentApi.java
```diff
@@ -0,0 +1,459 @@
...
+            fragmentClass);
+
+        methodGetFragmentManager = null;
+      }
+
+      Method methodGetId;
+      try {
+        methodGetId = fragmentClass.getMethod("getId");
+      } catch (Exception e) {
+        LogUtil.w(
```
jasta: Per our conversation, strongly prefer converting these to multicatches that catch all checked exceptions and rethrow using `throw new RuntimeException(ex);`.  But be careful, this advice only applies to methods that you expect exist (publicly documented, stable APIs).

stetho/src/main/java/com/facebook/stetho/common/android/FragmentApi.java
```diff
@@ -0,0 +1,459 @@
...
+    @Override
+    public Object getFragmentManager(Object fragment) {
+      if (mMethodGetFragmentManager == null) {
+        return null;
+      }
+
+      try {
+        return mMethodGetFragmentManager.invoke(fragment);
+      } catch (Exception e) {
+        LogUtil.w(e, "Could not call Fragment.getFragmentManager()");
```
jasta: The advice applies here as well on account of the fact that you've already proven the method exists.  It must be invokable else something very unusual has occurred that we want to know about.

stetho/src/main/java/com/facebook/stetho/common/android/ViewGroupUtil.java
```diff
@@ -0,0 +1,57 @@
...
+    try {
+      field = ViewGroup.class.getDeclaredField("mOnHierarchyChangeListener");
+    } catch (NoSuchFieldException ex) {
+      field = null;
+    }
+
+    sOnHierarchyChangeListenerField = field;
+
+    if (sOnHierarchyChangeListenerField != null) {
+      sOnHierarchyChangeListenerField.setAccessible(true);
```
jasta: To me, this reads more clearly as:

  try {
    Field field = ...;
    field.setAccessible(true);
    sOnHierarchyChangeListenerField = field;
  } catch (...) {
    sOnHierarchyChangeListenerField = null;
  }

stetho/src/main/java/com/facebook/stetho/common/android/ViewGroupUtil.java
```diff
@@ -0,0 +1,57 @@
...
+  public static ViewGroup.OnHierarchyChangeListener tryGetOnHierarchyChangeListenerHack(
+      ViewGroup viewGroup) {
+    if (sOnHierarchyChangeListenerField == null) {
+      return null;
+    }
+
+    try {
+      return (ViewGroup.OnHierarchyChangeListener)sOnHierarchyChangeListenerField.get(viewGroup);
+    } catch (IllegalAccessException ex) {
+      // should not happen since we called setAccessible(true)
```
jasta: throw :)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDOMProvider.java
```diff
@@ -27,18 +30,20 @@ public AndroidDOMProvider(Application application) {
...
-        .register(Application.class, new ApplicationDescriptor())
+        .register(Application.class, new ApplicationDescriptor());
+    FragmentDescriptor.register(mDescriptorMap)
         .register(Object.class, new ObjectDescriptor())
         .register(TextView.class, new TextViewDescriptor())
         .register(View.class, new ViewDescriptor())
         .register(ViewGroup.class, new ViewGroupDescriptor())
         .register(Window.class, new WindowDescriptor())
-        .setListener(new DescriptorListener())
+        .setHost(this)
```
jasta: Curious why this wouldn't be part of the constructor?  It's the only thing not explicitly additive...

stetho/src/main/java/com/facebook/stetho/inspector/elements/Descriptor.java
```diff
@@ -2,24 +2,39 @@
...
-
-  private Listener mListener;
+  private Host mHost;
 
   protected Descriptor() {
   }
 
-  void setListener(Listener listener) {
-    mListener = listener;
+  // This is called by DescriptorMap to hook us up to our Host (probably the DOMProvider).
```
jasta: nit: use Javadocs and `{@link}`.  Because this is package private, it is safe to document it in this way.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDOMProvider.java
jasta: Actually I misspoke.  I didn't realize `setHost` existed twice.  The other usage is the one I'm not so sure about...

stetho/src/main/java/com/facebook/stetho/inspector/elements/Descriptor.java
```diff
@@ -2,24 +2,39 @@
...
+  // This method is idempotent in the sense that once you call it with a specific
+  // reference you must either 1) never call it again, or 2) call it again with that
+  // same reference.
+  void setHost(Host host) {
```
jasta: This has a strange smell to me.  It seems we're missing an explicit initialization step that would take this variable and that we should require it be non-null and require that the class be initialized properly.  This way it's clearer to the reader that the class, after construction, is not yet ready to be used.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/FragmentDescriptor.java
```diff
@@ -0,0 +1,86 @@
...
+
+  private static final FragmentDescriptor sInstance;
+  static {
+    sInstance = new FragmentDescriptor();
+  }
+
+  public static DescriptorMap register(DescriptorMap map) {
+    Class<?> supportFragmentClass = FragmentApi.tryGetSupportFragmentClass();
+    if (supportFragmentClass != null) {
+      LogUtil.i("Registering support Fragment descriptor");
```
jasta: This seems to me like it should be `.d`.  `.i` will log in the caller application.

stetho/src/main/java/com/facebook/stetho/common/android/ViewGroupUtil.java
rickbrew: Agreed, except for the unfortunate fact that the compiler yells at me and says it might already be assigned (in the catch block).

stetho/src/main/java/com/facebook/stetho/inspector/elements/Descriptor.java
rickbrew: So you're suggesting we just rename this method to, e.g. initialize() ?

stetho/src/main/java/com/facebook/stetho/common/android/ViewGroupUtil.java
jasta: Oops.  Perhaps:
```java
static {
  sOnHiera... = determineOnHierarchyChangeListenerField();
}

@Nullable
private static Field determineOn...() { ... }
```

stetho/src/main/java/com/facebook/stetho/inspector/elements/Descriptor.java
jasta: Yes, I think it's as simple as that, though I don't have the full context like you do.

rickbrew: I can simplify this by changing the protocol. If I don't re-use one instance of FragmentDescriptor to handle both Fragment APIs, it means I don't have to worry about the idempotency stuff. What I originally ran into was that my one instance of FragmentDescriptor was being initialized twice from DescriptorMap, once for each Fragment class (native and support).

stetho/src/main/java/com/facebook/stetho/common/android/ViewGroupUtil.java
```diff
@@ -0,0 +1,53 @@
...
+import javax.annotation.Nullable;
+
+public final class ViewGroupUtil {
+  private static final Field sOnHierarchyChangeListenerField =
+      tryGetOnHierarchyChangeListenerField();
+
+  @Nullable
+  private static Field tryGetOnHierarchyChangeListenerField() {
+    try {
+      Field field = ViewGroup.class.getDeclaredField("mOnHierarchyChangeListener");
```
jasta: This should probably be refacotred to use `ReflectionUtil`, no?

stetho/src/main/java/com/facebook/stetho/common/android/ViewGroupUtil.java
```diff
@@ -0,0 +1,53 @@
...
+
+  @Nullable
+  public static ViewGroup.OnHierarchyChangeListener tryGetOnHierarchyChangeListenerHack(
+      ViewGroup viewGroup) {
+    if (sOnHierarchyChangeListenerField == null) {
+      return null;
+    }
+
+    try {
+      return (ViewGroup.OnHierarchyChangeListener)sOnHierarchyChangeListenerField.get(viewGroup);
```
jasta: This too, possibly

jasta:
Looks good!  Squash/rebase into a single commit then feel free to merge.

PR #105, opened 2015-03-26
Add optional dependency support for maven plugin.

jasta:
DO NOT MERGE; Work in progress.

This works just fine (verified with `gradle installArchives` and examining the resulting .pom file), but we're not ready to actually make use of optional deps yet in stetho-urlconnection.  We need to do some refactoring before this solution is meaningful, but it took a while to figure out how to do it so archiving here for now :)


condemoreloer:
our partners

jasta:
Replaced by #107

PR #106, opened 2015-03-26
Application element is now the root node in the UI

### LINKED ISSUE #99: Elements tab should have Application as root
Application is currently emitted as the root DOM document node, but the root document node doesn't show up in the elements tab. It's useful to have the Application at the root so we can at least see the class name in order to aid code navigation.

rickbrew:
The node that we emit from DOM.getDocument() does not show up in the elements tab (it's just the way it works). Right now we are using the Application for that, which prevents us from seeing any information about it. By emitting a faux-node whose only purpose is to emit the Application as its sole child, we can now see details about Application itself. Even just having the class name helps answer the ever important question of "where is the code for this?"

screen shot 2015-03-26 at 3 20 00 pm

Closes #99


jasta:
Looks good, merge after Travis gives it a green light.

PR #107, opened 2015-03-28
Demonstrate an "optimal" release configuration with Stetho

jasta:


jasta:
@longinoa @dreiss @rightparen Looking to get this merged in tomorrow.  Can any of you take a look?

stetho-urlconnection/src/main/java/com/facebook/stetho/urlconnection/StethoURLConnectionManager.java
```diff
@@ -37,24 +31,37 @@
  */
 @NotThreadSafe
 public class StethoURLConnectionManager {
-  private static final AtomicInteger sSequenceNumberGenerator = new AtomicInteger(0);
+  private static boolean sIsStethoPresent = isStethoPresent();
+
+  private static boolean isStethoPresent() {
```
rightparen: Rather than a method, consider putting this in a 'static' block to avoid accidentally calling this more than once.

stetho-urlconnection/src/main/java/com/facebook/stetho/urlconnection/StethoURLConnectionManager.java
```diff
@@ -37,24 +31,37 @@
...
+      return true;
+    } catch (ClassNotFoundException e) {
+      return false;
+    }
+  }
 
-  private final NetworkEventReporter mStethoHook = NetworkEventReporterImpl.get();
-  private final int mRequestId;
-  @Nullable private final String mFriendlyName;
+  private final Holder mHolder;
```
rightparen: You should add a comment that explains the purpose of the indirection through Holder.  Otherwise this looks like unnecessary complexity.

stetho-urlconnection/src/main/java/com/facebook/stetho/urlconnection/StethoURLConnectionManager.java
```diff
@@ -37,24 +31,37 @@
...
+    if (sIsStethoPresent) {
+      mHolder = new Holder(friendlyName);
+    } else {
+      mHolder = null;
+    }
   }
 
   public boolean isStethoEnabled() {
-    return mStethoHook.isEnabled();
+    return isStethoPresent() && mHolder.impl.isStethoActive();
```
rightparen: Here and throughout, consider checking (mHolder != null) rather than calling isStethoPresent() or even using sIsStethoPresent.  Because the implementation already has to be robust to being called when not active, the main goal of this check is to avoid NPE.  This intention would be clearer with the null check.

stetho-urlconnection/src/main/java/com/facebook/stetho/urlconnection/StethoURLConnectionManager.java
jasta: I like this pattern because it plays well with `static final` assignment (which I forgot to use above, oops).  Otherwise you have to use a temporary variable.

That said, I think @rickbrew 's `ReflectionUtil.tryGetClassForName` can bail us out here.

jasta: lol, no it can't.  Because this code is testing for Stetho's existence :P

stetho-urlconnection/src/main/java/com/facebook/stetho/urlconnection/StethoURLConnectionManager.java
jasta: Ack'd.  I meant to use the member variable here but yes `mHolder` is the right call.

PR #108, opened 2015-03-29
Workaround serious throughput issue in Stetho's HTTP server

jasta:
All communication between host and Stetho is done by a LocalSocket (a
UNIX abstract domain socket) which has a very heavy-handed
implementation of LocalSocketImpl#flush which polls/sleeps waiting for
the kernel buffer to become empty.  This behaviour in practice reduces
observed throughput of dumpapp (using either the buffered stdout or
stderr) to between 300 and 600KB/s despite adb being capable of ~5MB/s
on the same device.

This workaround involves blocking the final flush call to the socket
output stream.  This still drains the local Java buffer as originally
intended, but does not ask the kernel to drain its buffer (and block
doing so).  This kernel draining behaviour is entirely unnecessary for
our purposes and presumably for everyone else as it appears to be added
for some special purpose hack in the Bluetooth implementation.

Offending AOSP commit appears to be: https://android.googlesource.com/platform/frameworks/base/+/71bfafc84af4b820748b12e1a1010b0dfa7bdea6%5E%21/

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/server/LocalSocketHttpServerConnection.java
+++ b/stetho/src/main/java/com/facebook/stetho/server/LocalSocketHttpServerConnection.java
@@ -147,5 +147,15 @@ public class LocalSocketHttpServerConnection extends AbstractHttpServerConnectio
         HttpParams params) throws IOException {
       init(socket.getOutputStream(), bufferSize, params);
     }
+
+    @Override
+    public void flush() throws IOException {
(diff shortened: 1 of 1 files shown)


dcolascione:
lgtm

dcolascione:
It wouldn't hurt to link to the clowny AOSP code.

jasta:
Linked.

PR #109, opened 2015-03-29
Add `dumpapp hprof` plugin

jasta:
This adds a convenient mechanism to dump an HPROF file from a running
Android application.  A companion script (`./scripts/hprof_dump.sh`) is
included to automatically run hprov-conf to convert it to the standard
Java HPROF file format for use in tools like Eclipse's MemoryAnalyzer.

### DIFF
--- a/scripts/hprof_dump.sh
+++ b/scripts/hprof_dump.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+DUMPAPP="$DIR/dumpapp"
+
+set -e
+
+# This will generate an hprof on the device, download it locally, convert the
+# hprof to the standard format, and store it in the current working directory.
+# The resulting file can be explored with a tool such as the standalone Eclipse
+# MemoryAnalyzer: https://eclipse.org/mat/
+
+if [[ -z "$1" ]]; then
+  OUTFILE="out.hprof"
+else
+  OUTFILE=$1
+fi
+TEMPFILE="${OUTFILE}-dalvik.tmp"
+
+echo "Generating hprof on device (this can take a while)..."
+$DUMPAPP hprof - > ${TEMPFILE}
+
+echo "Converting $TEMPFILE to standard format..."
+hprof-conv $TEMPFILE $OUTFILE
+rm $TEMPFILE
+
+echo "Stored ${OUTFILE}"
--- a/stetho/src/main/java/com/facebook/stetho/dumpapp/plugins/HprofDumperPlugin.java
+++ b/stetho/src/main/java/com/facebook/stetho/dumpapp/plugins/HprofDumperPlugin.java
@@ -1,0 +1,121 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.dumpapp.plugins;
+
+import android.content.Context;
+import android.os.Debug;
+
+import com.facebook.stetho.common.Util;
+import com.facebook.stetho.dumpapp.DumpException;
+import com.facebook.stetho.dumpapp.DumpUsageException;
+import com.facebook.stetho.dumpapp.DumperContext;
+import com.facebook.stetho.dumpapp.DumperPlugin;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.PrintStream;
+import java.util.Iterator;
+
+/**
+ * Generates an hprof on the sdcard and returns to the client the absolute path to the file.
+ * <p/>
+ * For convenient usage, see: {@code scripts/hprof_dump.sh}
+ */
+public class HprofDumperPlugin implements DumperPlugin {
+
+  private static final String NAME = "hprof";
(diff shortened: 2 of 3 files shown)


stetho/src/main/java/com/facebook/stetho/dumpapp/plugins/HprofDumperPlugin.java
```diff
@@ -0,0 +1,121 @@
...
+    } else {
+      if ("-".equals(outputPath)) {
+        handlePipeOutput(output);
+      } else {
+        File outputFile = new File(outputPath);
+        if (!outputFile.isAbsolute()) {
+          outputFile = mContext.getFileStreamPath(outputPath);
+        }
+        writeHprof(outputFile);
+        output.println("Wrote to " + outputFile);
```
rightparen: Consider just outputting the filename without the "Wrote to " to make it easier for tools to read the filename.  It should be pretty clear to users in the context what the filename means.

stetho/src/main/java/com/facebook/stetho/dumpapp/plugins/HprofDumperPlugin.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  @Override
+  public void dump(DumperContext dumpContext) throws DumpException {
+    final PrintStream output = dumpContext.getStdout();
+
+    Iterator<String> argsIter = dumpContext.getArgsAsList().iterator();
+    String outputPath = argsIter.hasNext() ? argsIter.next() : null;
+    if (outputPath == null) {
+      usage(output);
```
rightparen: Why not dumpContext.getStderr()?

rightparen:
One consideration and one question inline.  Otherwise lgtm.

stetho/src/main/java/com/facebook/stetho/dumpapp/plugins/HprofDumperPlugin.java
jasta: `stdout` should always be used for legitimate help output, as is the case when the path is not provided.  `stderr` would only be used on unexpected input.

stetho/src/main/java/com/facebook/stetho/dumpapp/plugins/HprofDumperPlugin.java
jasta: Ack'd, though this is intending to be human readable since a proper scripted approach would hardcode the path and therefore it would know the path that was written if the command simply exited successfully.  Also note that scripts would be much better served by the `"-"` output path, as is the case with `hprof_dump.sh`.  Prefer to keep the language here as-is for human beings...

scripts/hprof_dump.sh
```diff
@@ -0,0 +1,27 @@
...
+else
+  OUTFILE=$1
+fi
+TEMPFILE="${OUTFILE}-dalvik.tmp"
+
+echo "Generating hprof on device (this can take a while)..."
+$DUMPAPP hprof - > ${TEMPFILE}
+
+echo "Converting $TEMPFILE to standard format..."
+hprof-conv $TEMPFILE $OUTFILE
```
longinoa: You should check to make sure this is in the path, and if not provide a link to http://developer.android.com/tools/help/hprof-conv.html or at least a warning.

scripts/hprof_dump.sh
```diff
@@ -0,0 +1,27 @@
...
+
+if [[ -z "$1" ]]; then
+  OUTFILE="out.hprof"
+else
+  OUTFILE=$1
+fi
+TEMPFILE="${OUTFILE}-dalvik.tmp"
+
+echo "Generating hprof on device (this can take a while)..."
+$DUMPAPP hprof - > ${TEMPFILE}
```
longinoa: no process selector?

stetho/src/main/java/com/facebook/stetho/dumpapp/plugins/HprofDumperPlugin.java
longinoa: Per convention this should throw a `DumpUsageException`.

Consider adding a `--help`, yes its a bit over-engineering for this, but this is a dumper that is being checked in and should be an example.

stetho/src/main/java/com/facebook/stetho/dumpapp/plugins/HprofDumperPlugin.java
```diff
@@ -0,0 +1,121 @@
...
+        hprofFile.delete();
+      }
+    }
+  }
+
+  private void writeHprof(File outputPath) throws DumpException {
+    try {
+      // Test that we can write here.  dumpHprofData appears to hang if it cannot write
+      // to the target location on ART.
+      truncateAndDeleteFile(outputPath);
```
longinoa: this seems kind of scary, why dont we just return the file name used to standard out and require it be parsed when used as a script?

scripts/hprof_dump.sh
jasta: This is in the same directory as `adb` and all those other tools.  If someone doesn't have it in their path, odds are they're not using the command-line at all :)

scripts/hprof_dump.sh
jasta: Hmm, seems a bit heavy handed to just toss `"$@"` in there but I suppose that'll do it.

stetho/src/main/java/com/facebook/stetho/dumpapp/plugins/HprofDumperPlugin.java
jasta: This is concerning manual usage as per:
```shell
dumpapp hprof /sdcard/foo
```

If you can't write to `/sdcard/foo`, it silently fails due to a bug in ART.  This code tests for that possibility before it calls into the system.

jasta:
Will produce a follow-up diff to address these issues.

PR #110, opened 2015-03-29
Add `dumpapp crash` plugin

jasta:
Powerful mechanism to simulate a variety of crash/kill scenarios common
on Android to verify that crash recovery / reporters are functioning
properly.  Especially useful as a way to simulate the low memory killer
which normally requires a separate script to first determine your app's
pid then invoke adb shell run-as <yourpackage> kill <yourpid>.

Example usage:

  dumpapp crash throw
  dumpapp crash kill
  dumpapp crash kill USR1
  dumpapp crash exit 0

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/dumpapp/plugins/CrashDumperPlugin.java
+++ b/stetho/src/main/java/com/facebook/stetho/dumpapp/plugins/CrashDumperPlugin.java
@@ -113,61 +113,63 @@ public class CrashDumperPlugin implements DumperPlugin {
   }
 
   private void doUncaughtException(Iterator<String> argsIter) throws DumpException {
-    String throwableClass = ArgsHelper.nextOptionalArg(argsIter, OPTION_THROW_DEFAULT);
+    String throwableClassString = ArgsHelper.nextOptionalArg(argsIter, OPTION_THROW_DEFAULT);
+
+
     try {
-      Thread crashThread = new Thread(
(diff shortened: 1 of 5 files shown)


stetho/src/main/java/com/facebook/stetho/dumpapp/plugins/CrashDumperPlugin.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    @Override
+    public void run() {
+      try {
+        Throwable t = createThrowable();
+        ExceptionUtil.<Error>sneakyThrow(t);
+      } catch (InstantiationException | IllegalAccessException | InvocationTargetException e) {
+        // Hehe, ok whatever :)
+        ExceptionUtil.<Error>sneakyThrow(e);
```
rightparen: Consider creating the Exception on the dumpapp thread and just sneakyThrowing it here.  Then any strange errors are visible to the command.  Also, then the state of the class can just be the Exception and you can be more explicit about the parameterless getDeclaredConstructor() being a fallback that intentionally leaks NoSuchMethodException.

jasta: Ack'd, this design was admittedly sloppy.  I'll tidy it up so that it's much clearer that the throw is the only thing which must be on a separate thread.

PR #113, opened 2015-03-31
Implement Elements, Properties sub-tab

jasta:
With this, developers can explore the Java object structure live from
Stetho as if a traditional debugger was attached.  This UI is a bit
buried and hard to notice, but can be found by clicking on an element in
the Elements tab and then finding the Properties tab on the right-hand
panel.  It is sometimes hidden behind a "more options"-style button.

elements-properties


jasta:
The styles sub-tab is next :)

rickbrew:
Looks good!

jasta:
Resubmitting based on master now that it's merged.

PR #120, opened 2015-04-01
Properly encode multiple headers with the same name

### LINKED ISSUE #115: NetworkInterceptor just shows the last Cookie
When sending requests with more than one Cookie, Stetho is only showing the last on the the Inspector.

request_bin
stetho_inspect
stetho_inspect_02

Using:
```java
compile 'com.facebook.stetho:stetho:1.0.1'
compile 'com.facebook.stetho:stetho-okhttp:1.0.1'
compile 'com.squareup.okhttp:okhttp:2.3.0'
```
And to add Cookies to OkHttp I'm using another NetworkInterceptor.

Here a project I put together to help with debug: https://github.com/tsuharesu/StethoBug

jasta:
Unspecified in the protocol docs, we're expected to separate multiple
header values with the same name using a new line character.  This is
necessary because a normal JSON object cannot have the same name
repeated twice (and headers are encoded using a JSON object not an
array).

Closes #115

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkEventReporterImpl.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkEventReporterImpl.java
@@ -280,8 +280,15 @@ public class NetworkEventReporterImpl implements NetworkEventReporter {
   private static JSONObject formatHeadersAsJSON(InspectorHeaders headers) {
     JSONObject json = new JSONObject();
     for (int i = 0; i < headers.headerCount(); i++) {
+      String name = headers.headerName(i);
+      String value = headers.headerValue(i);
       try {
-        json.put(headers.headerName(i), headers.headerValue(i));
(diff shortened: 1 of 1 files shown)


rightparen:
lgtm

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkEventReporterImpl.java
```diff
@@ -18,6 +18,7 @@
 import com.facebook.stetho.inspector.protocol.module.Network;
 import com.facebook.stetho.inspector.protocol.module.Page;
 import org.apache.http.protocol.HTTP;
+import org.json.JSONArray;
```
longinoa: dont think you need this

jasta: Removed; force pushed.

PR #122, opened 2015-04-01
Add `-p <process>` option to scripts/hprof_dump.sh

jasta:
This allows a caller to respect the advice given by the error message
("Use -p to select a process...").  Also fix an issue where those error
messages would not be displayed correctly to the user by way of my
terrible misunderanding of Python3.


rightparen:
Looks good.

PR #125, opened 2015-04-02
Inspector button on Elements tab

### LINKED ISSUE #124: Implement inspector button on Elements tab
It would be fantastic if you could click on the hourglass button, then tap on the View you want to be highlighted in the elements tab.

rickbrew:
This allows the "inspector" magnifying glass to work on the elements tab. When you click on the button, it enters a modal state where you can pick which UI element should be highlighted. When you push your finger down on the Android app, it will hit-test and highlight the appropriate view on the device (and also as you move your finger around, if you'd like). When you move your finger up, a command will be sent back to Chrome and it will now highlight the view in the inspector UI.

Closes #124

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/common/android/ViewUtil.java
+++ b/stetho/src/main/java/com/facebook/stetho/common/android/ViewUtil.java
@@ -4,13 +4,95 @@ package com.facebook.stetho.common.android;
 
 import android.app.Activity;
 import android.content.Context;
+import android.graphics.PointF;
 import android.view.View;
+import android.view.ViewGroup;
 import android.view.ViewParent;
+
+import com.facebook.stetho.common.Predicate;
+import com.facebook.stetho.common.Util;
 
 import javax.annotation.Nullable;
 
 public final class ViewUtil {
   private ViewUtil() {
+  }
+
+  @Nullable
+  public static View hitTest(View view, float x, float y) {
+    return hitTest(view, x, y, null /* viewSelector */);
+  }
+
+  // x,y are in view's local coordinate space (relative to its own top/left)
+  @Nullable
+  public static View hitTest(View view, float x, float y, @Nullable Predicate<View> viewSelector) {
+    if (!ViewUtil.pointInView(view, x, y)) {
+      return null;
+    }
+
+    if (!(view instanceof ViewGroup)) {
+      return null;
+    }
+
+    final ViewGroup viewGroup = (ViewGroup)view;
+
+    // TODO: get list of Views that are sorted in Z- and draw-order, e.g. buildOrderedChildList()
+    for (int i = viewGroup.getChildCount() - 1; i >= 0; --i) {
+      final View child = viewGroup.getChildAt(i);
+
+      if (child.getVisibility() != View.VISIBLE) {
+        continue;
+      }
+
+      if (viewSelector != null && !viewSelector.apply(child)) {
+        continue;
+      }
+
+      PointF localPoint = new PointF();
+      if (ViewUtil.isTransformedPointInView(viewGroup, child, x, y, localPoint)) {
+        if (child instanceof ViewGroup) {
+          return hitTest(child, localPoint.x, localPoint.y, viewSelector);
+        } else {
+          return child;
+        }
+      }
+    }
+
+    return viewGroup;
+  }
+
+  public static boolean pointInView(View view, float localX, float localY) {
+    return localX >= 0 && localX < (view.getRight() - view.getLeft())
+        && localY >= 0 && localY < (view.getBottom() - view.getTop());
+  }
+
+  public static boolean isTransformedPointInView(
+      ViewGroup parent,
+      View child,
+      float x,
+      float y,
+      @Nullable PointF outLocalPoint) {
+    Util.throwIfNull(parent);
+    Util.throwIfNull(child);
+
+    float localX = x + parent.getScrollX() - child.getLeft();
+    float localY = y + parent.getScrollY() - child.getTop();
+
+    // TODO: handle transforms
+    /*Matrix childMatrix = child.getMatrix();
+    if (!childMatrix.isIdentity()) {
+      final float[] localXY = new float[2];
+      localXY[0] = localX;
+      localXY[1] = localY;
+      child.getInverseMatrix
+    }*/
+
+    final boolean isInView = ViewUtil.pointInView(child, localX, localY);
+    if (isInView && outLocalPoint != null) {
+      outLocalPoint.set(localX, localY);
+    }
+
+    return isInView;
   }
 
   @Nullable
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDOMProvider.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDOMProvider.java
@@ -4,24 +4,44 @@ package com.facebook.stetho.inspector.elements.android;
 
 import android.app.Activity;
 import android.app.Application;
+import android.content.Context;
+import android.graphics.Canvas;
+import android.os.Handler;
+import android.os.Looper;
+import android.view.MotionEvent;
 import android.view.View;
 import android.view.ViewGroup;
 import android.view.Window;
+import android.view.WindowManager;
 import android.widget.TextView;
 
+import com.facebook.stetho.common.Predicate;
+import com.facebook.stetho.common.Util;
+import com.facebook.stetho.common.android.ViewUtil;
 import com.facebook.stetho.inspector.elements.DOMProvider;
 import com.facebook.stetho.inspector.elements.Descriptor;
 import com.facebook.stetho.inspector.elements.DescriptorMap;
 import com.facebook.stetho.inspector.elements.NodeDescriptor;
 import com.facebook.stetho.inspector.elements.ObjectDescriptor;
 
+import java.util.ArrayList;
+import java.util.List;
+
 final class AndroidDOMProvider implements DOMProvider, AndroidDescriptorHost {
+  private static final int INSPECT_OVERLAY_COLOR = 0x40FFFFFF;
+  private static final int INSPECT_HOVER_COLOR = 0x404040ff;
+
+  private final Application mApplication;
+  private final Handler mHandler;
   private final DescriptorMap mDescriptorMap;
   private final AndroidDOMRoot mDOMRoot;
   private final ViewHighlighter mHighlighter;
+  private final InspectModeHandler mInspectModeOverlay;
   private Listener mListener;
 
   public AndroidDOMProvider(Application application) {
+    mApplication = Util.throwIfNull(application);
+    mHandler = new Handler(Looper.getMainLooper());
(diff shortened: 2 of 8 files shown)


stetho/src/main/java/com/facebook/stetho/common/android/ViewUtil.java
```diff
@@ -14,6 +19,85 @@ private ViewUtil() {
   }
 
   @Nullable
+  public static View hitTest(View view, float x, float y) {
+    return hitTest(view, x, y, null);
```
jasta: nit: `null /* viewSelector */`.

stetho/src/main/java/com/facebook/stetho/common/android/ViewUtil.java
```diff
@@ -4,9 +4,14 @@
 
 import android.app.Activity;
 import android.content.Context;
+import android.graphics.PointF;
 import android.view.View;
+import android.view.ViewGroup;
 import android.view.ViewParent;
 
+import com.android.internal.util.Predicate;
```
jasta: You cannot use classes in Android's `internal` package space.

stetho/src/main/java/com/facebook/stetho/common/android/ViewUtil.java
```diff
@@ -14,6 +19,85 @@ private ViewUtil() {
   }
 
   @Nullable
+  public static View hitTest(View view, float x, float y) {
+    return hitTest(view, x, y, null);
+  }
+
+  // x,y are in view's local coordinate space
```
jasta: Javadoc.  Also, what does "local coordinate space" mean?  Is it relative to the view's `getLeft()/getTop()`?  Be specific please.

stetho/src/main/java/com/facebook/stetho/common/android/ViewUtil.java
```diff
@@ -14,6 +19,85 @@ private ViewUtil() {
...
+      return null;
+    }
+
+    final ViewGroup viewGroup = (ViewGroup)view;
+
+    // TODO: get list of Views that are sorted in Z- and draw-order, e.g. buildOrderedChildList()
+    for (int i = viewGroup.getChildCount() - 1; i >= 0; --i) {
+      final View child = viewGroup.getChildAt(i);
+
+      if (child.getVisibility() == View.INVISIBLE) {
```
jasta: Test for `View.VISIBLE`.  You want to ignore both invisible and gone.

stetho/src/main/java/com/facebook/stetho/common/android/ViewUtil.java
```diff
@@ -14,6 +19,85 @@ private ViewUtil() {
...
+    return viewGroup;
+  }
+
+  // see: View.pointInView()
+  public static boolean pointInView(View view, float localX, float localY) {
+    return localX >= 0 && localX < (view.getRight() - view.getLeft())
+        && localY >= 0 && localY < (view.getBottom() - view.getTop());
+  }
+
+  // see: ViewGroup.isTransformedTouchPointInView()
```
jasta: nit: javadoc, with `{@link}`.  This makes it clickable.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDOMProvider.java
```diff
@@ -129,4 +157,140 @@ public View getHighlightingView(Object element) {
 
     return highlightingView;
   }
+
+  private List<Window> getWindows() {
```
jasta: nit: because this allocates a new list each time, prefer a non-getter verb like: `collectWindows()`.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDOMProvider.java
```diff
@@ -129,4 +157,140 @@ public View getHighlightingView(Object element) {
...
+      for (int wi = 0; wi < windowCount; ++wi) {
+        final Window window = (Window)activityDescriptor.getChildAt(activity, wi);
+        windows.add(window);
+      }
+    }
+
+    return windows;
+  }
+
+  public final class InspectModeHandler {
```
jasta: Should be private, no?

stetho/src/main/java/com/facebook/stetho/common/android/ViewUtil.java
rickbrew: Should I just create my own Predicate interface then? Or is there something I can re-use?

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDOMProvider.java
```diff
@@ -129,4 +157,140 @@ public View getHighlightingView(Object element) {
...
+
+    private final Predicate<View> mViewSelector = new Predicate<View>() {
+      @Override
+      public boolean apply(View view) {
+        return !(view instanceof DOMHiddenView);
+      }
+    };
+
+    public InspectModeHandler() {
+      mEnableRunnable = new Runnable() {
```
jasta: nit: I consider this lambda-style pattern to be most readable if you logically think of this expression as a function definition.  In that way, it would read:
```java
public void enable() { ... }
  
private void enableOnUiThread() { ... }
  
private final Runnable mEnableRunnable = () -> {
  // code here...
};
```

Yes it technically violates that member variables go up top, but the pattern really is about a defining a function and referring to it as a pointer.  In C/C++, for instance, do exactly that making it really unintuitive for you to declare the method inside a function body as a nested function.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDOMProvider.java
```diff
@@ -129,4 +157,140 @@ public View getHighlightingView(Object element) {
...
+    }
+
+    private final class OverlayView extends DOMHiddenView {
+      public OverlayView(Context context) {
+        super(context);
+      }
+
+      @Override
+      protected void onDraw(Canvas canvas) {
+        canvas.drawColor(0x40FFFFFF);
```
jasta: Use a named constant at the top of this file.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDOMProvider.java
```diff
@@ -129,4 +157,140 @@ public View getHighlightingView(Object element) {
...
+
+      @Override
+      public boolean onTouchEvent(MotionEvent event) {
+        if (getParent() instanceof View) {
+          final View parent = (View)getParent();
+          View view = ViewUtil.hitTest(parent, event.getX(), event.getY(), mViewSelector);
+
+          if (event.getAction() != MotionEvent.ACTION_CANCEL) {
+            if (view != null) {
+              mHighlighter.setHighlightedView(view, 0x404040ff);
```
jasta: named constnat again

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewGroupDescriptor.java
```diff
@@ -82,13 +100,60 @@ public void unhook() {
       }
     }
 
-    public Object getChildAt(ViewGroup element, int index) {
-      if (index < 0 || index >= element.getChildCount()) {
+    public void setIsDecorView(boolean value) {
```
jasta: nit: `markDecorView` which always sets it to true.  It makes it clear that you're just locating it, and that you won't be unsetting this value once marked.  It's less flexible, but on purpose.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -338,6 +361,22 @@ public void onChildRemoved(Object parentElement, Object childElement) {
     public RGBAColor contentColor;
   }
 
+  private static class InspectNodeRequestedEvent {
```
jasta: nit: I think we use `*Params` for events and `*Request/Response` for commands.  See `Network` for examples.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -251,6 +260,20 @@ public void onChildRemoved(Object parentElement, Object childElement) {
 
       removeElementTree(childElement);
     }
+
+    @Override
+    public void onInspectRequested(Object element) {
```
jasta: nit: seems like this should be `onInspectNodeClicked` or something.

jasta:
Mostly nits and minor fixes requested.  Feel free to ignore nits, but do address the rest.

stetho/src/main/java/com/facebook/stetho/common/android/ViewUtil.java
jasta: Define your own.  It was added in Java 8, but that's a long ways off on Android.

stetho/src/main/java/com/facebook/stetho/common/android/ViewUtil.java
rickbrew: Yes. I'm using the terminology used within View/ViewGroup. Easy enough to clarify locally too.

stetho/src/main/java/com/facebook/stetho/common/android/ViewUtil.java
rickbrew: hmmmmm unfortunately View.pointInView() is package-private (which is why I copied it here) and this doesn't seem to work ...

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDOMProvider.java
rickbrew: Yes! good catch

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: It's named this way so as to match the name of DOM.inspectNodeRequested, which is what this call will directly map to (and is intended to be thought of in that way). Except it isn't a node, so we omit that word. And the other methods omit 'element', so we don't replace 'node' with 'element.'

jasta: Tomato, tomato.  I was thinking the direction of the verb was backwards in this case since it implied that the inspect request came from Stetho, rather than asynchronously answering an inspect request from Chrome.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: This is consistent with the other "events" in this file (look above for a big handful of them). Should we just go ahead and rename them all right now?

jasta: No, not a big deal.  Just ignore me :)

PR #127, opened 2015-04-02
Avoid installing the DOM module when Activity detection is unavailable

jasta:
The DOM module effectively doesn't work due to our hack that lists all
activities not functioning properly on previous versions of Android.
Without this, we currently cannot provide the DOM feature.

We can fix these issues by providing a manual API that informs us of
activity lifecycle events similar to Romain Guy's ViewServer:
https://github.com/romainguy/ViewServer


rickbrew:
Looks good.

PR #128, opened 2015-04-02
DOM now calls into AndroidDOMProvider via the UI thread

### LINKED ISSUE #114: DOM.getDocument() must run on UI thread
We traverse the UI tree via the Descriptors, and this must happen on the UI thread. It's important that we run the entire traversal on the UI thread. So, DOM.getDocument() should have some way of delegating to, e.g. Activity.runOnUiThread(), for its call to createNodeForElement(rootElement).

rickbrew:
This patches DOM so that it calls into AndroidDOMProvider on the UI thread in the necessary locations. Otherwise we have a strong chance of stepping on UI objects from background threads. Proper enforcement and discovery of the threading model for AndroidDOMProvider needs a bit more chin-scratching, which we plan to do soon.

Closes #114

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/common/android/HandlerUtil.java
+++ b/stetho/src/main/java/com/facebook/stetho/common/android/HandlerUtil.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.common.android;
+
+import android.os.Handler;
+import android.os.Looper;
+
+public final class HandlerUtil {
+  private HandlerUtil() {
+  }
+
+  public static boolean postAndWait(Handler handler, final Runnable r) {
+    if (Looper.myLooper() == handler.getLooper()) {
+      r.run();
+      return true;
+    }
+
+    WaitableRunnable wrapper = new WaitableRunnable(r);
+    if (!handler.post(wrapper)) {
+      return false;
+    }
+
+    wrapper.join();
+    return true;
+  }
+
+  private static final class WaitableRunnable implements Runnable {
+    private final Runnable mRunnable;
+    private boolean mIsDone;
+
+    public WaitableRunnable(Runnable runnable) {
+      mRunnable = runnable;
+    }
+
+    @Override
+    public void run() {
+      try {
+        mRunnable.run();
+      } finally {
+        synchronized (this) {
+          mIsDone = true;
+          notifyAll();
+        }
+      }
+    }
+
+    public void join() {
+      synchronized (this) {
+        while (!mIsDone) {
+          try {
+            wait();
+          } catch (InterruptedException e) {
+          }
+        }
+      }
+    }
+  }
+}
--- a/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
@@ -46,7 +46,7 @@ public class DOM implements ChromeDevtoolsDomain {
   private final DOMObjectIdMapper mObjectIdMapper;
 
   @Nullable
-  private DOMProvider mDOMProvider;
+  private volatile DOMProvider mDOMProvider;
 
   public DOM(DOMProvider.Factory providerFactory) {
     mDOMProviderFactory = Util.throwIfNull(providerFactory);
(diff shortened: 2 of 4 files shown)


stetho/src/main/java/com/facebook/stetho/common/android/HandlerUtil.java
```diff
@@ -0,0 +1,64 @@
...
+package com.facebook.stetho.common.android;
+
+import android.os.Handler;
+
+public final class HandlerUtil {
+  private HandlerUtil() {
+  }
+
+  public static boolean postAndWait(Handler handler, final Runnable r) {
+    if (handler.getLooper().getThread() == Thread.currentThread()) {
```
jasta: Use: `Looper.myLooper() == handler.getLooper()`

stetho/src/main/java/com/facebook/stetho/common/android/HandlerUtil.java
```diff
@@ -0,0 +1,64 @@
...
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.common.android;
+
+import android.os.Handler;
+
+public final class HandlerUtil {
```
jasta: Package private, methinks.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -70,15 +70,18 @@ public void disable(JsonRpcPeer peer, JSONObject params) {
...
-
-    Node rootNode = createNodeForElement(rootElement);
+    final GetDocumentResponse result = new GetDocumentResponse();
+
+    mDOMProvider.postAndWait(new Runnable() {
+       @Override
+       public void run() {
+         Object rootElement = mDOMProvider.getRootElement();
+         if (rootElement != null) {
+           result.root = createNodeForElement(rootElement);
```
jasta: While this assignment is safe because `postAndWait` locks consistently on both threads, I'd prefer clarifying that by pattern as such:

```java
final AtomicReference<...> rootNode = new AtomicReference<>(...);
postAndWait(() -> {
  rootNode.set(theThing);
});
```

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -88,12 +91,18 @@ public void highlightNode(JsonRpcPeer peer, JSONObject params) {
...
-      RGBAColor contentColor = request.highlightConfig.contentColor;
+      final RGBAColor contentColor = request.highlightConfig.contentColor;
       if (contentColor == null) {
         LogUtil.w("DOM.highlightNode was not given a color to highlight with");
       } else {
-        Object element = mObjectIdMapper.getObjectForId(request.nodeId);
-        mDOMProvider.highlightElement(element, contentColor.getColor());
+        final Object element = mObjectIdMapper.getObjectForId(request.nodeId);
+
+        mDOMProvider.postAndWait(new Runnable() {
```
jasta: This strictly adds delay per our discussion.  Prefer `highlightElement` be made safe to run on any arbitrary thread to avoid this perf hit.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -195,7 +204,12 @@ protected synchronized void onFirstPeerRegistered() {
...
     protected synchronized void onLastPeerUnregistered() {
-      mObjectIdMapper.clear();
+      mDOMProvider.postAndWait(new Runnable() {
+        @Override
+        public void run() {
+          mObjectIdMapper.clear();
+        }
+      });
 
       mDOMProvider.dispose();
```
jasta: Is this thread safe?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -195,7 +204,12 @@ protected synchronized void onFirstPeerRegistered() {
...
       mDOMProvider = null;
```
jasta: This probably needs to be volatile.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
jasta: Oh, I think I see why this is done.  Presumably so we can keep access to the `*Descriptor` system consistent on the same thread as well.  Nevermind.

stetho/src/main/java/com/facebook/stetho/common/android/HandlerUtil.java
rickbrew: It's used outside of this package though

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: That doesn't seem necessary though? It's safe because postAndWait() doesn't return until the Runnable completes. Using an AtomicReference wouldn't gain us real safety, and then this code might look like it's being unnecessarily over-cautious due to the developer not being sure about the safety of not using AtomicReference in this case.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: Which part? Why wouldn't it be? ObjectIdMapper locks internally. clear() is implemented carefully to make sure that this is a plausible way to do things (it doesn't lock while broadcasting the onUnmapped calls).

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: Okie dokie

jasta:
Looks good, rebase and merge.

PR #129, opened 2015-04-03
Bring id pretty printing to <fragment> support as well.

jasta:
Previously only View instances in the DOM would format ids using human
readable names.  This refactors the logic into a common class and
modifies FragmentDescriptor to support this.

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
@@ -2,11 +2,10 @@
 
 package com.facebook.stetho.inspector.elements.android;
 
-import android.content.res.Resources;
 import android.view.View;
 
-import com.facebook.stetho.common.LogUtil;
 import com.facebook.stetho.common.StringUtil;
+import com.facebook.stetho.common.android.ResourcesUtil;
 import com.facebook.stetho.inspector.elements.AttributeAccumulator;
 import com.facebook.stetho.inspector.elements.ChainedDescriptor;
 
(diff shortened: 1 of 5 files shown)


stetho/src/main/java/com/facebook/stetho/common/android/ResourceIdUtil.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+package com.facebook.stetho.common.android;
+
+import android.content.res.Resources;
+import com.facebook.stetho.common.LogUtil;
+
+import javax.annotation.Nonnull;
+
+public class ResourceIdUtil {
```
rickbrew: since this is a class with utility methods for 'Resources', perhaps we should call this 'ResourcesUtil'

stetho/src/main/java/com/facebook/stetho/common/android/ResourceIdUtil.java
```diff
@@ -0,0 +1,65 @@
...
+  private ResourceIdUtil() {
+  }
+
+  @Nonnull
+  public static String getIdStringQuietly(Resources r, int resourceId) {
+    try {
+      return getIdString(r, resourceId);
+    } catch (Resources.NotFoundException e) {
+      String idString = "#" + Integer.toHexString(resourceId);
+      LogUtil.w("Unknown identifier encountered: " + idString);
```
rickbrew: if I saw this in logcat I'd have no idea where it came from ... "Unknown identifier encountered" is very generic! let's tag it with class/method name or something

jasta: Ack'd, will require context with this method.

stetho/src/main/java/com/facebook/stetho/common/android/ResourceIdUtil.java
jasta: Ack'd, will rename.

stetho/src/main/java/com/facebook/stetho/common/android/ResourceIdUtil.java
```diff
@@ -0,0 +1,65 @@
...
+      return idString;
+    }
+  }
+
+  public static String getIdString(Resources r, int resourceId) throws Resources.NotFoundException {
+    String prefix;
```
rickbrew: this code isn't handling the case where 'r' is null ... whereas the old code did.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -32,61 +31,13 @@ protected void onCopyAttributes(View element, AttributeAccumulator attributes) {
...
-  private static int getResourcePackageId(int id) {
-    return (id >>> 24) & 0xff;
-  }
-
   @Nullable
   private static String getIdAttribute(View element) {
-    // Adapted from View.toString()
-
     int id = element.getId();
     if (id == View.NO_ID) {
```
rickbrew: why don't we make it so that getIdStringQuietly() handles the 'id == NO_ID' case?

stetho/src/main/java/com/facebook/stetho/common/android/ResourceIdUtil.java
jasta: Resources cannot be null for either View or Fragment.  In the View case, it's impossible due to the final field; in the Fragment case, it will throw before it happens due to the resources coming from Activity which also guarantees non-null resources.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
jasta: In both of our use cases that would make the check redundant since you don't want to even report that the attribute exists unless the id is not -1.

stetho/src/main/java/com/facebook/stetho/common/android/ResourceUtil.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+package com.facebook.stetho.common.android;
+
+import android.content.res.Resources;
+import com.facebook.stetho.common.LogUtil;
+
+import javax.annotation.Nonnull;
+
+public class ResourceUtil {
```
rickbrew: nit ... Resources, not Resource

rickbrew:
Just the one last nit on name of Resource[s]Util, then good to squash 'n merge

PR #130, opened 2015-04-03
Fetch deep links to find large format image URLs for APOD sample

jasta:
This adds somewhat arbitrary network overhead to highlight the kinds of
bugs/inefficiencies that Stetho is good at rooting out.  Specifically in
this case, we improved the quality of the APOD thumbnail image but by
greatly overfetching data and slowing down the UI.  Stetho's Network tab
makes it very easy to see why (adding an extra round-trip through an
HTML page and downloading images that are 4x+ larger than the space they
occupy)

### DIFF
--- a/stetho-sample/src/main/java/com/facebook/stetho/sample/APODActivity.java
+++ b/stetho-sample/src/main/java/com/facebook/stetho/sample/APODActivity.java
@@ -109,7 +109,7 @@ public class APODActivity extends ListActivity {
       int bindPosition = cursor.getPosition();
       holder.position = bindPosition;
 
-      final String imageUrl = cursor.getString(APODPostsQuery.DESCRIPTION_IMAGE_URL_INDEX);
+      final String imageUrl = cursor.getString(APODPostsQuery.LARGE_IMAGE_URL_INDEX);
       holder.image.setImageDrawable(null);
       fetchImage(imageUrl, bindPosition, holder);
 
@@ -181,13 +181,13 @@ public class APODActivity extends ListActivity {
         APODContract.Columns._ID,
         APODContract.Columns.TITLE,
         APODContract.Columns.DESCRIPTION_TEXT,
-        APODContract.Columns.DESCRIPTION_IMAGE_URL,
+        APODContract.Columns.LARGE_IMAGE_URL,
     };
 
     public static final int ID_INDEX = 0;
     public static final int TITLE_INDEX = 1;
     public static final int DESCRIPTION_TEXT_INDEX = 2;
-    public static final int DESCRIPTION_IMAGE_URL_INDEX = 3;
+    public static final int LARGE_IMAGE_URL_INDEX = 3;
 
     public static CursorLoader createCursorLoader(Context context) {
       return new CursorLoader(
--- a/stetho-sample/src/main/java/com/facebook/stetho/sample/APODRssFetcher.java
+++ b/stetho-sample/src/main/java/com/facebook/stetho/sample/APODRssFetcher.java
@@ -9,25 +9,25 @@
 
 package com.facebook.stetho.sample;
 
-import java.io.ByteArrayInputStream;
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.List;
-
 import android.content.ContentProviderOperation;
 import android.content.ContentResolver;
 import android.content.ContentValues;
 import android.content.OperationApplicationException;
-import android.graphics.Color;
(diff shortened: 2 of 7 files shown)


stetho-sample/src/main/java/com/facebook/stetho/sample/APODRssFetcher.java
```diff
@@ -73,23 +75,46 @@ public void onFailure(IOException e) {
...
+    public List<ApodItem> decorateRssItemsWithLinkImages(List<RssItem> rssItems) {
+      ArrayList<ApodItem> apodItems = new ArrayList<>(rssItems.size());
+      final CountDownLatch fetchLinkLatch = new CountDownLatch(rssItems.size());
+      for (RssItem rssItem : rssItems) {
+        final ApodItem apodItem = new ApodItem();
+        apodItem.rssItem = rssItem;
+        fetchLinkPage(rssItem.link, new PageScrapedCallback() {
+          @Override
+          public void onPageScraped(@Nullable List<String> imageUrls) {
+            apodItem.largeImageUrl = imageUrls != null && !imageUrls.isEmpty()
```
longinoa: When would this ever be null?

stetho-sample/src/main/java/com/facebook/stetho/sample/HtmlScraper.java
```diff
@@ -0,0 +1,72 @@
...
+        .toString();
+
+    return strippedText.trim();
+  }
+
+  private static class ExtractImageGetter implements Html.ImageGetter {
+    @Nullable private final String mOriginUrl;
+    private final List<String> mSources;
+
+    public ExtractImageGetter(@Nullable String originUrl, List<String> outSources) {
```
longinoa: calling this `outSources` is confusing. these are Urls

stetho-sample/src/main/java/com/facebook/stetho/sample/APODRssFetcher.java
jasta: If it can't fetch the URL provided it uses `null` to indicate this instead of an empty list which communicates that it fetched the page but it had no `` tags in it.  The distinction is ignored by the APOD-specific code, but the lower level code reads more clearly IMO having a distinction.

stetho-sample/src/main/java/com/facebook/stetho/sample/HtmlScraper.java
jasta: It's using the semantics of `Html.ImageGetter`.  I can rename though.

stetho-sample/src/main/java/com/facebook/stetho/sample/APODActivity.java
```diff
@@ -109,7 +109,7 @@ public void bindView(View view, Context context, Cursor cursor) {
       int bindPosition = cursor.getPosition();
       holder.position = bindPosition;
 
-      final String imageUrl = cursor.getString(APODPostsQuery.DESCRIPTION_IMAGE_URL_INDEX);
+      final String imageUrl = cursor.getString(APODPostsQuery.LARGE_IMAGE_URL_INDEX);
```
longinoa: This is now `@Nullable` `DESCRIPTION_IMAGE_URL` when its null?

jasta: You mean `imageUrl` can now be null?  It could be null before too, technically, it just isn't in practice.  Also, I checked and I don't think the code will necessary do the wrong thing if it's null since the Networker will just fail the request immediately.  I haven't tested that, though.

stetho-sample/src/main/java/com/facebook/stetho/sample/APODRssFetcher.java
```diff
@@ -73,23 +75,46 @@ public void onFailure(IOException e) {
...
+    private List<RssItem> parseRss(byte[] body) throws IOException, XmlPullParserException {
       XmlPullParser parser = Xml.newPullParser();
       parser.setInput(new ByteArrayInputStream(body), "UTF-8");
       List<RssItem> items = new RssParser(parser).parse();
       Log.d(TAG, "Fetched " + items.size() + " items");
 
+      return items;
+    }
+
+    public List<ApodItem> decorateRssItemsWithLinkImages(List<RssItem> rssItems) {
```
longinoa: this isnt really decorating rss items, its more creating `ApodItem`

PR #134, opened 2015-04-03
Improve Inspector button's hit-testing

### LINKED ISSUE #133: Inspector button in Elements tab gets 'stuck' in certain ViewGroups
We haven't really implemented good "touch" hit-testing. In some apps, our hit-testing gets confused and gives us full-screen layout containers that aren't actually visible or even responding to input. From a pedantic rendering system point-of-view, it's the right answer, but it's not what we need for this feature.

rickbrew:
ViewUtil.hitTest() is now renamed to hitTestTouch() to better articulate what it should be used for.

We were having trouble in some apps where the inspector button wouldn't work very well: it would often latch onto layout containers that were empty overlay stubs. Since we can't actually send a touch event down and see how finally responds to it, we're adjusting our algorithm: ViewGroups are effectively a lower priority than regular Views. To implement this, we just do two passes of hit-testing: in the first pass, we don't allow a ViewGroup to be the result and instead return null. The second pass allows ViewGroups as a result.

Closes #133

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/common/android/ViewUtil.java
+++ b/stetho/src/main/java/com/facebook/stetho/common/android/ViewUtil.java
@@ -20,51 +20,44 @@ public final class ViewUtil {
   private ViewUtil() {
   }
 
-  private static boolean isTouchable(View view) {
+  private static boolean isHittable(View view) {
     if (view.getVisibility() != View.VISIBLE) {
       return false;
     }
 
-    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.HONEYCOMB) {
-      if (getAlphaHoneycomb(view) < 0.001f) {
-        return false;
-      }
+    if (ViewCompat.getInstance().getAlpha(view) < 0.001f) {
+      return false;
     }
 
     return true;
   }
 
-  @TargetApi(Build.VERSION_CODES.HONEYCOMB)
-  private static float getAlphaHoneycomb(View view) {
-    return view.getAlpha();
-  }
-
   @Nullable
-  public static View hitTestTouch(View view, float x, float y) {
-    return hitTestTouch(view, x, y, null /* viewSelector */);
+  public static View hitTest(View view, float x, float y) {
+    return hitTest(view, x, y, null /* viewSelector */);
   }
 
   // x,y are in view's local coordinate space (relative to its own top/left)
   @Nullable
-  public static View hitTestTouch(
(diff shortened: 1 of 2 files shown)


stetho/src/main/java/com/facebook/stetho/common/android/ViewUtil.java
```diff
@@ -18,47 +20,85 @@
   private ViewUtil() {
   }
 
+  private static boolean isTouchable(View view) {
+    if (view.getVisibility() != View.VISIBLE) {
+      return false;
+    }
+
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.HONEYCOMB) {
```
jasta: nit: this is the anti-pattern I discussed earlier that has the tendency to create perf problems when you aren't very careful about protecting the Dalvik verifier from seeing illegal code.  The pattern which makes this simple to avoid is the same as one you've used elsewhere:
```java
abstract class ViewGetAlphaCompat {
  @Nullable private static final ViewGetAlphaCompat sInstance;

  static {
    if (honeycomb+) {
      sInstance = new HoneycombAndBeyondImpl();
    } else {
      // Can't be done...
      sInstance = null;
    }
  }

  @Nullable
  public static ViewGetAlphaCompat getInstance() {
    return sInstance;
  }

  public float getAlpha(View v);

  @TargetApi(HONEYCOMB)
  private static class HoneycombAndBeyondImpl { ... }
}
```

stetho/src/main/java/com/facebook/stetho/common/android/ViewUtil.java
```diff
@@ -18,47 +20,85 @@
...
+    return result;
+  }
+
+  private static View hitTestTouchImpl(
+      View view,
+      float x,
+      float y,
+      @Nullable Predicate<View> viewSelector,
+      boolean allowViewGroupResult) {
+    if (!isTouchable(view)) {
```
jasta: Non-touchable views should still be selectable using the inspect button.  The purpose is to find views in the tree, not mimic touch behaviour...

jasta:
Discussed out of band: I think we want to drop the isTouchable test, but go with the rest of this heuristic.  There are gotchas, but we'll just need to let this soak in the wild to find a better strategy.

jasta:
The HONEYCOMB+ nit is totally optional.

jasta:
:+1:

PR #143, opened 2015-04-07
Fixed the implementation of ViewUtil.hitTestImpl()

### LINKED ISSUE #142: Typo in ViewUtil.hitTestImpl() defeats two-pass heuristic
ViewUtil.hitTest() is implemented as a 2-pass call to the recursive hitTestImpl(). The first pass uses allowViewGroupResult = false, the second pass uses false.

However, hitTestImpl() is recursively calling hitTest() instead of hitTestImpl(). This basically completely defeats the priority of View over ViewGroup for results.

rickbrew:
VewUtil.hitTestImpl needs to recursively call itself, not hitTest(). Otherwise it defeats our heuristic, preventing allowViewGroupResult from being passed down the call chain.

Closes #142


jasta:
:+1:

PR #152, opened 2015-04-10
Implements support for DOM.setAttributesAsText()

### LINKED ISSUE #145: Any plans to implement DOM.setAttributesAsText?
That would be awesome!

felipecsl:
Fixes #145 

This is my initial stab at implementing a new DOM request `setAttributesAsText()`.

Right now this correctly calls any setter that has a single argument of type `int`, `float`, `boolean`, `String` or `CharSequence` of `View` (and subclasses) using the `MethodInvoker`. 
Examples are `setTex()t`, `setId()`, `setAlpha()`, etc. More types can be supported by extending `TypedMethodInvoker`.

Question: 
* Do we need to send a response back after completed?

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/common/Util.java
+++ b/stetho/src/main/java/com/facebook/stetho/common/Util.java
@@ -23,6 +23,17 @@ public class Util {
     return item;
   }
 
+  public static <T1, T2> void throwIfNull(T1 item1, T2 item2) {
+    throwIfNull(item1);
+    throwIfNull(item2);
+  }
+
+  public static <T1, T2, T3> void throwIfNull(T1 item1, T2 item2, T3 item3) {
+    throwIfNull(item1);
+    throwIfNull(item2);
+    throwIfNull(item3);
+  }
+
   public static void throwIfNotNull(Object item) {
     if (item != null) {
       throw new IllegalStateException();
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/ChainedDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/ChainedDescriptor.java
@@ -185,4 +185,14 @@ public abstract class ChainedDescriptor<E> extends Descriptor {
 
   protected void onCopyAttributes(E element, AttributeAccumulator attributes) {
   }
+
+  @Override
+  @SuppressWarnings("unchecked")
+  public final void setAttributesAsText(Object element, String text) {
+    onSetAttributesAsText((E)element, text);
+  }
+
+  protected void onSetAttributesAsText(E element, String text) {
+    mSuper.setAttributesAsText(element, text);
+  }
 }
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/DOMProvider.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/DOMProvider.java
@@ -30,6 +30,8 @@ public interface DOMProvider extends ThreadBound {
 
   public void setInspectModeEnabled(boolean enabled);
 
+  public void setAttributesAsText(Object element, String text);
+
   public static interface Factory {
     DOMProvider create();
   }
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/Descriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/Descriptor.java
@@ -14,6 +14,8 @@ import com.facebook.stetho.common.UncheckedCallable;
 import com.facebook.stetho.common.Util;
 
 import javax.annotation.Nullable;
+import java.util.HashMap;
+import java.util.Map;
 
 public abstract class Descriptor implements NodeDescriptor {
   private Host mHost;
@@ -53,6 +55,41 @@ public abstract class Descriptor implements NodeDescriptor {
   @Override
   public final void postAndWait(Runnable r) {
     getHost().postAndWait(r);
+  }
+
+  /**
+   * Parses the text argument text from DOM.setAttributeAsText()
+   * Text will be in the format "attribute1=\"Value 1\" attribute2=\"Value2\""
+   * @param text the text argument to be parsed
+   * @return a map of attributes to their respective values to be set.
+   */
+  protected Map<String, String> parseSetAttributesAsTextArg(String text) {
+    String value = "";
+    String key = "";
+    StringBuilder buffer = new StringBuilder();
+    Map<String, String> keyValuePairs = new HashMap<>();
+    boolean isInsideQuotes = false;
+    for (int i = 0; i < text.length(); ++i) {
+      final char c = text.charAt(i);
+      if (c == '=') {
+        key = buffer.toString();
+        buffer.setLength(0);
+      } else if (c == '\"') {
+        if (isInsideQuotes) {
+          value = buffer.toString();
+          buffer.setLength(0);
+        }
+        isInsideQuotes = !isInsideQuotes;
+      } else if (c == ' ' && !isInsideQuotes) {
+        keyValuePairs.put(key, value);
+      } else {
+        buffer.append(c);
+      }
+    }
+    if (!key.isEmpty() && !value.isEmpty()) {
+      keyValuePairs.put(key, value);
+    }
+    return keyValuePairs;
   }
 
   public interface Host extends ThreadBound {
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDOMProvider.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDOMProvider.java
@@ -148,6 +148,16 @@ final class AndroidDOMProvider implements DOMProvider, AndroidDescriptorHost {
     }
   }
 
+  @Override
+  public void setAttributesAsText(Object element, String text) {
+    verifyThreadAccess();
+
+    Descriptor descriptor = mDescriptorMap.get(element.getClass());
+    if (descriptor != null) {
+      descriptor.setAttributesAsText(element, text);
+    }
+  }
+
   // Descriptor.Host implementation
   @Override
   public Descriptor getDescriptor(Object element) {
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/MethodInvoker.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/MethodInvoker.java
@@ -1,0 +1,131 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.elements.android;
+
+import com.facebook.stetho.common.LogUtil;
+import com.facebook.stetho.common.Util;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Tries to arbitrarily invoke single argument methods by name on an object instance by trying out
+ * different argument types.
+ */
+public class MethodInvoker {
+
+  private static final List<TypedMethodInvoker<?>> invokers = Arrays.asList(
+      new StringMethodInvoker(), new CharSequenceMethodInvoker(), new IntegerMethodInvoker(),
+      new FloatMethodInvoker(), new BooleanMethodInvoker());
+
+  /**
+   * Tries to invoke a method on receiver with a single argument by trying out different types
+   * for arg until it finds one that matches (or not). No exceptions are thrown on failure.
+   *
+   * @param methodName The method name to be invoked
+   * @param argument The single argument to be provided to the method
+   */
+  public void invoke(Object receiver, String methodName, String argument) {
+    Util.throwIfNull(receiver, methodName, argument);
+    int size = invokers.size();
+    for (int i = 0; i < size; ++i) {
+      final TypedMethodInvoker<?> invoker = invokers.get(i);
+      if (invoker.invoke(receiver, methodName, argument)) {
+        return;
+      }
+    }
+    LogUtil.w("Method with name " + methodName +
+              " not found for any of the MethodInvoker supported argument types.");
+  }
+
+  private static abstract class TypedMethodInvoker<T> {
+    private final Class<T> mArgType;
+
+    TypedMethodInvoker(Class<T> argType) {
+      mArgType = argType;
+    }
+
+    boolean invoke(Object receiver, String methodName, String argument) {
+      try {
+        Method method = receiver.getClass().getMethod(methodName, mArgType);
+        method.invoke(receiver, convertArgument(argument));
+        return true;
+      } catch (NoSuchMethodException ignored) {
+        // ignore
+      } catch (InvocationTargetException e) {
+        LogUtil.w("InvocationTargetException: " + e.getMessage());
+      } catch (IllegalAccessException e) {
+        LogUtil.w("IllegalAccessException: " + e.getMessage());
+      } catch (IllegalArgumentException e) {
+        LogUtil.w("IllegalArgumentException: " + e.getMessage());
+      }
+      return false;
+    }
+
+    abstract T convertArgument(String argument);
+  }
+
+  private static class StringMethodInvoker extends TypedMethodInvoker<String> {
+    StringMethodInvoker() {
+      super(String.class);
+    }
+
+    @Override
+    String convertArgument(String argument) {
+      return argument;
+    }
+  }
+
+  private static class CharSequenceMethodInvoker extends TypedMethodInvoker<CharSequence> {
+    CharSequenceMethodInvoker() {
+      super(CharSequence.class);
+    }
+
+    @Override
+    CharSequence convertArgument(String argument) {
+      return argument;
+    }
+  }
+
+  private static class IntegerMethodInvoker extends TypedMethodInvoker<Integer> {
+    IntegerMethodInvoker() {
+      super(int.class);
+    }
+
+    @Override
+    Integer convertArgument(String argument) {
+      return Integer.parseInt(argument);
+    }
+  }
+
+  private static class FloatMethodInvoker extends TypedMethodInvoker<Float> {
+    FloatMethodInvoker() {
+      super(float.class);
+    }
+
+    @Override
+    Float convertArgument(String argument) {
+      return Float.parseFloat(argument);
+    }
+  }
+
+  private static class BooleanMethodInvoker extends TypedMethodInvoker<Boolean> {
+    BooleanMethodInvoker() {
+      super(boolean.class);
+    }
+
+    @Override
+    Boolean convertArgument(String argument) {
+      return Boolean.parseBoolean(argument);
+    }
+  }
+}
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
@@ -10,16 +10,26 @@
 package com.facebook.stetho.inspector.elements.android;
 
 import android.view.View;
-
 import com.facebook.stetho.common.StringUtil;
 import com.facebook.stetho.common.android.ResourcesUtil;
 import com.facebook.stetho.inspector.elements.AttributeAccumulator;
 import com.facebook.stetho.inspector.elements.ChainedDescriptor;
 
 import javax.annotation.Nullable;
+import java.util.Map;
 
 final class ViewDescriptor extends ChainedDescriptor<View> implements HighlightableDescriptor {
   private static final String ID_ATTRIBUTE_NAME = "id";
+
+  private final MethodInvoker mMethodInvoker;
+
+  public ViewDescriptor() {
+    this(new MethodInvoker());
+  }
+
+  public ViewDescriptor(MethodInvoker methodInvoker) {
+    mMethodInvoker = methodInvoker;
+  }
 
   @Override
   protected String onGetNodeName(View element) {
@@ -38,6 +48,16 @@ final class ViewDescriptor extends ChainedDescriptor<View> implements Highlighta
     }
   }
 
+  @Override
+  protected void onSetAttributesAsText(View element, String text) {
+    Map<String, String> attributeToValueMap = parseSetAttributesAsTextArg(text);
+    for (Map.Entry<String, String> entry : attributeToValueMap.entrySet()) {
+      String methodName = "set" + capitalize(entry.getKey());
+      String propertyValue = entry.getValue();
+      mMethodInvoker.invoke(element, methodName, propertyValue);
+    }
+  }
+
   @Nullable
   private static String getIdAttribute(View element) {
     int id = element.getId();
@@ -50,5 +70,14 @@ final class ViewDescriptor extends ChainedDescriptor<View> implements Highlighta
   @Override
   public View getViewForHighlighting(Object element) {
     return (View)element;
+  }
+
+  private static String capitalize(String str) {
+    if (str == null || str.length() == 0 || Character.isTitleCase(str.charAt(0))) {
+      return str;
+    }
+    StringBuilder buffer = new StringBuilder(str);
+    buffer.setCharAt(0, Character.toTitleCase(buffer.charAt(0)));
+    return buffer.toString();
   }
 }
--- a/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
@@ -32,11 +32,10 @@ import com.facebook.stetho.json.annotation.JsonProperty;
 
 import org.json.JSONObject;
 
+import javax.annotation.Nullable;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
-
-import javax.annotation.Nullable;
 
 public class DOM implements ChromeDevtoolsDomain {
   private final ChromePeerManager mPeerManager;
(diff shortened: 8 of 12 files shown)


jasta:
Wow awesome, thanks.  Can't sink in now (on my mobile) but a few high level comments:

1. there seem to be unrelated changes upgrading gradle and such.
2. Why only support android:text?  The system we have defines something much more general that should be able to set any field we define.  Perhaps what's missing is a mapping function or a field/method name for reflection.  Then you could ask any descriptor to assign any attribute it emitted.

Also, I see a call to Runtime.mapObject but I'm guessing that mapping is not needed (you never reference it and you don't seem to unmap).

And finally, any thoughts on extending this and implementing Runtime's setProperties method (or something like that I don't have the docs handy)

felipecsl:
Great feedback, I'll take another look at it tonight to address your concerns. Thanks

felipecsl:
Ok so I had a quick look through the DevTools remote debugging protocol [documentation](https://developer.chrome.com/devtools/docs/protocol/1.1/index) (I assume that's the right one) but I couldn't find `Runtime.setProperties`. Are you referring to `callFunctionOn` or `evaluate` maybe?

Regarding making it more generic, allowing you to set any field, I'm already working on a reflection based solution which would try to basically find a setter method on the object, which would allow you to set attributes arbitrarily as long as there is a matching setter method for that in the object (eg.: `setId`, `setText`, etc.). Is this roughly what you had in mind? If so i will continue following this path and should have it ready and tested by tomorrow.

felipecsl:
Alright I moved on and implemented a `MethodInvoker` class that does basically what I described in the previous comment. Also added tests to it and `TextViewDescriptor`. 
Please let me know what your thoughts are about it

longinoa:
@felipecsl The documentation is here: https://code.google.com/p/chromium/codesearch#chromium/src/third_party/WebKit/Source/devtools/protocol.json&q=protocol.json&sq=package:chromium&type=cs

Cant speak tot he rest of @jasta's comments as I havent been following this PR.

stetho/src/main/java/com/facebook/stetho/common/Util.java
```diff
@@ -25,6 +25,12 @@
     return item;
   }
 
+  public static <T> void throwIfNull(T... items) {
```
rickbrew: Let's start by adding throwIfNull() overloads for the number of parameters that we need it for, instead of using varags. In this case, I see it being called with 3 parameters. This will avoid allocation of an array for the sake of parameter validation (which needs to be as close to zero impact as possible).

stetho/src/main/java/com/facebook/stetho/inspector/elements/DOMProvider.java
```diff
@@ -23,6 +23,8 @@
 
   public void setInspectModeEnabled(boolean enabled);
 
+  void setAttributesAsText(Object element, String text);
```
rickbrew: Please match the style of the surrounding code by putting 'public' in front of the method.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDOMProvider.java
```diff
@@ -114,6 +114,15 @@ public void setInspectModeEnabled(boolean enabled) {
     }
   }
 
+  @Override
+  public void setAttributesAsText(Object element, String text) {
+    Descriptor descriptor = mDescriptorMap.get(element.getClass());
+    if (descriptor != null && descriptor instanceof TextViewDescriptor) {
```
rickbrew: This code here makes me think that this really should be forwarded directly to a method on  NodeDescriptor. Otherwise we'll be making a trip back here every time we need to add setAttributeAsText() support for any new type. This also entails adding a pair of methods to ChainedDescriptor (it's a pretty straightforward pattern over there).

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -450,6 +460,17 @@ public int getColor() {
...
 
+  private static class SetAttributesAsTextRequest {
+    @JsonProperty(required = true)
+    public int nodeId;
+
+    @JsonProperty(required = true)
+    public String text;
+
+    @JsonProperty
+    public String name;
```
rickbrew: do we use this? if it's not required as per the JSON protocol documentation, then I bet we can just omit it

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -29,16 +28,13 @@
 import com.facebook.stetho.inspector.protocol.ChromeDevtoolsMethod;
 import com.facebook.stetho.json.ObjectMapper;
 import com.facebook.stetho.json.annotation.JsonProperty;
-import com.facebook.stetho.json.annotation.JsonValue;
-
 import org.json.JSONObject;
 
+import javax.annotation.Nullable;
```
rickbrew: hmm ... why'd this get moved up?

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/TextViewDescriptor.java
```diff
@@ -5,21 +5,30 @@
...
 
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.IdentityHashMap;
 import java.util.Map;
 
 final class TextViewDescriptor extends ChainedDescriptor<TextView> {
   private static final String TEXT_ATTRIBUTE_NAME = "text";
 
+  private final MethodInvoker mMethodInvoker;
```
rickbrew: this MethodInvoker class is pretty cool, but we don't (yet?) want to support arbitrary set___() calls. We only emit a very small subset of attributes to the element tree like 'id' and 'text', and so we should only support writing to that small subset. For now it should be sufficient to just check that name='text' and then call the setText() method. Supporting all get___/set___ method pairs, along with the marshaling of types to and from strings, would be a giant can of worms.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/TextViewDescriptor.java
```diff
@@ -32,6 +41,21 @@ protected void onUnhook(TextView element) {
     context.unhook();
   }
 
+  // TODO: Copied from ViewGroupDescriptor. Should we move it to the base class?
```
rickbrew: I thought about doing that, but it ended up working better as a pattern ... for now anyway. Every descriptor seems to have slight variations on what it's doing with the ElementContext, and we'd have to do a bunch of clumsy stuff with generics and casting (I think). Over time we'll have a better idea as to whether this should be consolidated into the base Descriptor or ChainedDescriptor class.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/TextViewDescriptor.java
```diff
@@ -71,5 +95,49 @@ public void afterTextChanged(Editable s) {
         getHost().onAttributeModified(mElement, TEXT_ATTRIBUTE_NAME, s.toString());
       }
     }
+
+    void setAttributeAsText(String text) {
+      // Text will be in the format "attribute1=\"Value 1\" attribute2=\"Value2\""
```
rickbrew: this parsing and splitting and mapping stuff should be in a utility method somewhere so that we can easily use it from other descriptors too. For now it's probably sufficient to put it as a static protected method in Descriptor.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/TextViewDescriptor.java
```diff
@@ -71,5 +95,49 @@ public void afterTextChanged(Editable s) {
...
+          keyValuePairs.put(key, value);
+        } else {
+          buffer += c;
+        }
+      }
+      if (!key.isEmpty() && !value.isEmpty()) {
+        keyValuePairs.put(key, value);
+      }
+
+      for (Map.Entry<String, String> entry : keyValuePairs.entrySet()) {
```
rickbrew: regarding my earlier comment about MethodInvoker, I think it's better for now if we just recognize that name == TEXT_ATTRIBUTE_NAME, and call setText() directly

rickbrew:
Okay so that's my first round of commentary :) This is heading in the right direction

felipecsl:
Thanks for the thorough review. I'm out this week, will probably be able to get back to this tuesday next week.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/TextViewDescriptor.java
felipecsl: I implemented this based on @jasta's comment above: `Why only support android:text? The system we have defines something much more general that should be able to set any field we define. Perhaps what's missing is a mapping function or a field/method name for reflection. Then you could ask any descriptor to assign any attribute it emitted.`.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/TextViewDescriptor.java
felipecsl: ok I'll remove the TODO for now then

felipecsl:
OK I think I have all the comments addressed.
Regarding your comment about limiting it to `setText` and `setId`, I actually disagree: With this solution as is, right now you can set other attributes via the Elements tree, for example alpha, gravity, etc. which is actually pretty handy. Even though Stetho only displays id and text inline, you're free to set anything you want, and that makes it really powerful!

Let me know what you think. Even though you might not want to be able to set other attributes (for now), I'm sure this will be useful in the future

stetho/src/main/java/com/facebook/stetho/inspector/elements/ChainedDescriptor.java
```diff
@@ -172,6 +172,11 @@ public final Object getChildAt(Object element, int index) {
     return onGetChildAt((E)element, thisIndex);
   }
 
+  @Override
+  public void setAttributesAsText(Object element, String text) {
```
rickbrew: hmm why doesn't this follow the pattern of all the other methods in this class? e.g. with an `onSetAttributesAsText(E, String)` method? In this case you can follow the pattern set by `getNodeType()` and `onGetNodeType()`: the default implementation of `onGetNodeType()` just calls mSuper's implementation.

stetho/src/main/java/com/facebook/stetho/inspector/elements/Descriptor.java
```diff
@@ -55,6 +58,44 @@ public final void postAndWait(Runnable r) {
     getHost().postAndWait(r);
   }
 
+  public void setAttributesAsText(Object element, String text) {
```
rickbrew: I think I see what you're doing here -- you want the default implementation to be a no-op, right? Let's not put this in `Descriptor`, and leave that up to `ChainedDescriptor` (none of the other methods on `NodeDescriptor` have any implementation in `Descriptor`). Then in `ObjectDescriptor`, which is the only one that doesn't derive from `ChainedDescriptor`, we'll leave it as an empty no-op method, which is what `ChainedDescriptor` will eventually call into anyway if it's not overriden.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/TextViewDescriptor.java
```diff
@@ -27,6 +26,7 @@
   private final Map<TextView, ElementContext> mElementToContextMap =
       Collections.synchronizedMap(new IdentityHashMap<TextView, ElementContext>());
 
+
```
rickbrew: random whitespace edits?

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/TextViewDescriptor.java
```diff
@@ -12,7 +12,6 @@
 import android.text.Editable;
 import android.text.TextWatcher;
 import android.widget.TextView;
-
```
rickbrew: random whitespace edits?

rickbrew:
Regarding the ability to set things like alpha, weight, etc.: first, I 100% agree that it's useful and it's definitely the direction we're wanting to go in, with high priority. I think we may actually want those (alpha, weight, etc.) to show up in the Styles sub-tab though, as a virtualized CSS. So in the main tree view we'd show a small, curated subset of attributes, and over in Styles we'd show everything. I'm still really not liking the idea of allowing arbitrary setters to be called via reflection and whatnot, but I'm wrestling with whether the convenience of this outweights my objections, especially in the shorter-term. Like maybe we'd hold onto this for now while making our way towards fleshing out the Styles stuff. Anyway give me like a day to think about this some more ... cool stuff so far

felipecsl:
Sure, I understand your concerns and I think we're on the same page! Thanks for the feedback

stetho/src/main/java/com/facebook/stetho/inspector/elements/Descriptor.java
felipecsl: fixing!

felipecsl: But how am I supposed to call `mSuper.setAttributesAsText(element, text);` on `ChainedDescriptor` if `Descriptor` doesn't implement that method?

felipecsl: Ohhhh ok. I think I get what you're saying. I'll move it to `NodeDescriptor`

felipecsl:
Fixed. @rickbrew is this what you had in mind?

rickbrew:
Yeah just add 'final' to that method :) (the one I commented on just now)

stetho/src/main/java/com/facebook/stetho/inspector/elements/ChainedDescriptor.java
```diff
@@ -172,6 +172,16 @@ public final Object getChildAt(Object element, int index) {
     return onGetChildAt((E)element, thisIndex);
   }
 
+  @Override
+  @SuppressWarnings("unchecked")
+  public final void setAttributesAsText(Object element, String text) {
```
rickbrew: oh ... `setAttributesAsText` and `onSetAttributesAsText` are sandwiched between `getChildAt` and `onGetChildAt`. We need to keep the paired methods next to each other.

stetho/src/main/java/com/facebook/stetho/inspector/elements/Descriptor.java
```diff
@@ -55,6 +57,41 @@ public final void postAndWait(Runnable r) {
...
   }
 
+  /**
+   * Parses the text argument text from DOM.setAttributeAsText()
+   * Text will be in the format "attribute1=\"Value 1\" attribute2=\"Value2\""
+   * @param text the text argument to be parsed
+   * @return a map of attributes to their respective values to be set.
+   */
+  protected Map<String, String> parseSetAttributesAsTextArg(String text) {
+    char[] chars = text.toCharArray();
```
rickbrew: making a copy of a String to a char-array for the purpose of looping over it is kind of a bad idea, performance-wise. Should use a regular for loop like so,
```
for (int i = 0; i < text.length(); ++i) {
  final char c = text.charAt(i);
  ...
}
```

stetho/src/main/java/com/facebook/stetho/inspector/elements/Descriptor.java
```diff
@@ -55,6 +57,41 @@ public final void postAndWait(Runnable r) {
...
+    String value = "";
+    String key = "";
+    String buffer = "";
```
rickbrew: we should use a StringBuilder for this. building a String by appending to it in a loop is a classic performance anti-pattern in languages like Java and C#

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDOMProvider.java
```diff
@@ -148,6 +148,14 @@ public void setInspectModeEnabled(boolean enabled) {
     }
   }
 
+  @Override
+  public void setAttributesAsText(Object element, String text) {
```
rickbrew: The first thing this method needs to do is call `verifyThreadAccess()`. We want to do that at the start of any method that is potentially dangerous if misused (e.g. `hook()` and `unhook()`), and also for any method that writes to an element (which this one definitely does). See the 4th paragraph of the long-winded javadocs for `ChainedDescriptor` for my explanation of this.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/MethodInvoker.java
```diff
@@ -0,0 +1,129 @@
...
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Tries to arbitrarily invoke single argument methods by name on an object instance by trying out
+ * different argument types.
+ */
+public class MethodInvoker {
```
rickbrew: final

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/MethodInvoker.java
```diff
@@ -0,0 +1,129 @@
...
+  private static abstract class TypedMethodInvoker<T> {
+    private Class<T> mArgType;
+
+    TypedMethodInvoker(Class<T> argType) {
+      mArgType = argType;
+    }
+
+    boolean invoke(Object receiver, String methodName, String argument) {
+      try {
+        Method method = TextView.class.getMethod(methodName, mArgType);
```
rickbrew: i'm really not sure why this is hard-coded for TextView?

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -51,4 +95,35 @@ private static String getIdAttribute(View element) {
   public View getViewForHighlighting(Object element) {
     return (View)element;
   }
+
+  private class ElementContext {
+    private View mElement;
+
+    public void hook(View element) {
```
rickbrew: If we're not actually doing anything in `hook()` or `unhook()`, and we're not keeping around any context or state for this `View`, then we don't need to have an `ElementContext` either. `onSetAttributesAsText()` then doesn't need to forward to `ElementContext.setAttributesAsText()` and can just inline that code. 

It's actually very important to try and avoid utilizing an `ElementContext` for the base `View` class because then we'll have an extra allocation for every single `View` in the system! :)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -51,4 +95,35 @@ private static String getIdAttribute(View element) {
...
+        String propertyValue = entry.getValue();
+        mMethodInvoker.invoke(mElement, methodName, propertyValue);
+      }
+    }
+
+    private String capitalize(String str) {
+      if (str == null || str.length() == 0) {
+        return str;
+      }
+      char[] buffer = str.toCharArray();
```
rickbrew: I believe it's better to round-trip this through a `StringBuilder` instead of `char[]`, e.g.
```
StringBuilder buffer = new StringBuilder(str);
buffer.setCharAt(0, Character.toTitleCase(buffer.charAt(0));
return buffer.toString();
```
It seems equivalent, but `StringBuilder` has an important optimization that saves an extra copying of the string. It creates a `String` that aliases the same internal `char[]`, and then also switches to copy-on-write mode. So it's saving an extra copy and a temporary allocation.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -51,4 +95,35 @@ private static String getIdAttribute(View element) {
...
+    private void setAttributeAsText(Map<String, String> attributeToValueMap) {
+      for (Map.Entry<String, String> entry : attributeToValueMap.entrySet()) {
+        String methodName = "set" + capitalize(entry.getKey());
+        String propertyValue = entry.getValue();
+        mMethodInvoker.invoke(mElement, methodName, propertyValue);
+      }
+    }
+
+    private String capitalize(String str) {
+      if (str == null || str.length() == 0) {
```
rickbrew: we could also check to see if the first character is already capitalized? probably won't gain us much in practice, but just seems like an obvious optimization.

stetho/src/test/java/com/facebook/stetho/inspector/elements/android/MethodInvokerTest.java
```diff
@@ -0,0 +1,64 @@
...
+import org.robolectric.RobolectricTestRunner;
+import org.robolectric.annotation.Config;
+
+import static org.junit.Assert.assertEquals;
+
+@Config(emulateSdk = Build.VERSION_CODES.JELLY_BEAN)
+@RunWith(RobolectricTestRunner.class)
+public class MethodInvokerTest {
+
+  private final TextView mTextView = new TextView(Robolectric.setupActivity(Activity.class));
```
rickbrew: let's also test something other than `TextView` ... like create a `CheckBox` and invoke `setChecked`

stetho/src/test/java/com/facebook/stetho/inspector/elements/android/ViewDescriptorTest.java
```diff
@@ -0,0 +1,54 @@
...
+import static org.mockito.Matchers.anyString;
+import static org.mockito.Mockito.*;
+
+@Config(emulateSdk = Build.VERSION_CODES.JELLY_BEAN)
+@RunWith(RobolectricTestRunner.class)
+public class ViewDescriptorTest {
+
+  private final MethodInvoker mMethodInvoker = mock(MethodInvoker.class);
+  private final ViewDescriptor mDescriptor = new ViewDescriptor(mMethodInvoker);
+  private final TextView mTextView = new TextView(Robolectric.setupActivity(Activity.class));
```
rickbrew: let's also test something other than `TextView` ... like create a `CheckBox` and then do a `setAttributesAsText(mCheckBox, "checked=\"true\"")`

rickbrew:
Ok there's another round of code review feedbacks

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/MethodInvoker.java
felipecsl: oops 😅

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/MethodInvoker.java
felipecsl: Mockito can't mock this class if it's final

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/MethodInvoker.java
```diff
@@ -0,0 +1,128 @@
...
+    Util.throwIfNull(receiver, methodName, argument);
+    for (TypedMethodInvoker<?> invoker : invokers) {
+      if (invoker.invoke(receiver, methodName, argument)) {
+        return;
+      }
+    }
+  }
+
+  private static abstract class TypedMethodInvoker<T> {
+    private Class<T> mArgType;
```
rickbrew: final

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/MethodInvoker.java
```diff
@@ -0,0 +1,128 @@
...
+  /**
+   * Tries to invoke a method on receiver with a single argument by trying out different types
+   * for arg until it finds one that matches (or not). No exceptions are thrown on failure.
+   *
+   * @param methodName The method name to be invoked
+   * @param argument The single argument to be provided to the method
+   */
+  public void invoke(Object receiver, String methodName, String argument) {
+    Util.throwIfNull(receiver, methodName, argument);
+    for (TypedMethodInvoker<?> invoker : invokers) {
```
rickbrew:
```
int size = invokers.size();
for (int i = 0; i < size; ++i) {
  final TypedMethodInvoker<?> invoker = invokers.get(i);
  ...
```

stetho/build.gradle
```diff
@@ -26,6 +26,7 @@ dependencies {
         exclude module: 'commons-logging'
         exclude module: 'httpclient'
     }
+    testCompile 'org.apache.maven:maven-ant-tasks:2.1.3'
```
jasta: Why do we have this?

felipecsl: I needed this in order to make the tests pass locally, can't remember the exact reason. I'll try to remove it and see if they still pass

stetho/src/main/java/com/facebook/stetho/common/Util.java
```diff
@@ -23,6 +23,17 @@
     return item;
   }
 
+  public static <T> void throwIfNull(T item1, T item2) {
+    throwIfNull(item1);
+    throwIfNull(item2);
+  }
+
+  public static <T> void throwIfNull(T item1, T item2, T item3) {
```
rickbrew: should we have multiple type arguments here, like so? (and for the 2 parameter version above of course)
```
public static <T1, T2, T3> void throwIfNull(T1 item1, T2 item2, T3 item3) { ...
```

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/MethodInvoker.java
rickbrew: aha, I see, that makes sense

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/MethodInvoker.java
```diff
@@ -0,0 +1,128 @@
...
+      mArgType = argType;
+    }
+
+    boolean invoke(Object receiver, String methodName, String argument) {
+      try {
+        Method method = receiver.getClass().getMethod(methodName, mArgType);
+        method.invoke(receiver, convertArgument(argument));
+        return true;
+      } catch (NoSuchMethodException e) {
+        LogUtil.w("Method with name " + methodName + " and argument of type " +
```
rickbrew: Won't this create a lot of debug spew? For instance, something like `setAlpha` is always going to fail with String, then fail with Char, then fail with Integer, before finally succeeding with Float. We'll get 3 log outputs with that every time, right? Probably better to only log in the calling method if it can't find any `invoke()` that returns true. The `InvocationTargetException`s should be fine though.

rickbrew:
Okay there's another few comments. Definitely look into why Travis is crying, too

rickbrew:
Looks good to me :) Merging ...

felipecsl:
Thanks :)

PR #156, opened 2015-04-11
Rework FragmentApi to not use reflection

### LINKED ISSUE #140: Update documentation
Seems stetho 1.1 now works for the element view but requires more reflexion meaning all proguarded app will crash : 

```
     Caused by: java.lang.RuntimeException: java.lang.NoSuchMethodException: getId []
            at com.facebook.stetho.common.ReflectionUtil.getMethod(ReflectionUtil.java:45)
            at com.facebook.stetho.common.android.FragmentApi$ReflectionFragmentAccessor.<init>(FragmentApi.java:157)
            at com.facebook.stetho.common.android.FragmentApi.<clinit>(FragmentApi.java:44)
            at com.facebook.stetho.inspector.elements.android.FragmentDescriptor.register(FragmentDescriptor.java:21)
```

It would be interesting to have an update of the documentation for all the needed (and correct) proguard configuration.

For my needs I use a simple :

```
-keep class com.facebook.stetho.** {*;}
-keepclassmembers public class * {
   *** get*();
}
```

But a smarter and proper configuration would be way better.

jasta:
This change has a few significant advantages:

1. Much more efficient
2. Lint-safe (requiring @TargetApi where appropriate)
3. Causes proguard to keep the support library methods we reference

Closes #140


jasta:
@rickbrew testing needs some work on this change, but i'd like you to take a look early in case there are any major gotchas functionally

stetho/src/main/java/com/facebook/stetho/common/android/FragmentActivityAccessor.java
```diff
@@ -6,10 +6,9 @@
...
-public interface FragmentActivityAccessor {
+public interface FragmentActivityAccessor<
+    FRAGMENT_ACTIVITY extends Activity,
+    FRAGMENT_MANAGER> {
   @Nullable
-  public Object getFragmentManager(Activity fragmentActivity);
-
-  @Nullable
-  public Object getSupportFragmentManager(Activity fragmentActivity);
+  public FRAGMENT_MANAGER getFragmentManager(FRAGMENT_ACTIVITY activity);
```
rickbrew: I'm pretty sure we can't remove the getSupportFragmentManager() method from this interface. The reason is that, at run-time, we could have an Activity class deriving from the v4 Support FragmentActivity. In this case we would have both getFragmentManager() and getSupportFragmentManager() and we do need to check both of them.

jasta: I believe I preserved the original behaviour by having two accessors,
rather than one with two methods.  See the caller code which previously
invoked getSupportFragmentManager.

rickbrew: Okay. I wanted to make sure this had been taken into consideration. This caused me quite a bit of chin-scratching on one particular afternoon.

stetho/src/main/java/com/facebook/stetho/common/android/FragmentApi.java
```diff
@@ -1,246 +0,0 @@
-// Copyright 2004-present Facebook. All Rights Reserved.
```
rickbrew: all code deleted = (y)

stetho/src/main/java/com/facebook/stetho/common/android/FragmentCompatFramework.java
```diff
@@ -0,0 +1,115 @@
...
+import android.app.Fragment;
+import android.app.FragmentManager;
+import android.content.res.Resources;
+import android.os.Build;
+import android.view.View;
+
+import javax.annotation.Nullable;
+
+@TargetApi(Build.VERSION_CODES.HONEYCOMB)
+class FragmentCompatFramework
```
rickbrew: can we make this final

stetho/src/main/java/com/facebook/stetho/common/android/FragmentCompatFramework.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+  }
+
+  @TargetApi(Build.VERSION_CODES.JELLY_BEAN_MR1)
+  private static class FragmentAccessorFrameworkJellyBean
+      extends FragmentAccessorFrameworkHoneycomb {
+    @Nullable
+    @Override
+    public FragmentManager peekChildFragmentManager(Fragment fragment) {
+      return fragment.getChildFragmentManager();
```
rickbrew: This peek___() method is calling get___() ? In the old code, the point of implementing a peek___() method via reflection was to avoid the lazy-instantiation consequence of calling get___(). That said, now that I think about it, I'm not completely sure if this is really worth the optimization since there aren't going to be THAT many Fragments, right? If so, we might rename our peek___() to get___(). Not sure what the overall impact to application performance would be, though, since these FragmentManagers will still be there if Stetho is disconnected.

stetho/src/main/java/com/facebook/stetho/common/android/FragmentCompatSupportLib.java
```diff
@@ -0,0 +1,95 @@
...
+
+import android.content.res.Resources;
+import android.support.v4.app.Fragment;
+import android.support.v4.app.FragmentActivity;
+import android.support.v4.app.FragmentManager;
+import android.view.View;
+
+import javax.annotation.Nullable;
+
+class FragmentCompatSupportLib
```
rickbrew: final?

stetho/src/main/java/com/facebook/stetho/common/android/FragmentCompatSupportLib.java
```diff
@@ -0,0 +1,95 @@
...
+    @Nullable
+    @Override
+    public View getView(Fragment fragment) {
+      return fragment.getView();
+    }
+
+    @Nullable
+    @Override
+    public FragmentManager peekChildFragmentManager(Fragment fragment) {
+      return fragment.getChildFragmentManager();
```
rickbrew: same comment as before w.r.t. get___() vs peek___()

rickbrew:
Okay other than some small nits and questions, looks good to me (y)

stetho/src/main/java/com/facebook/stetho/common/android/FragmentCompatFramework.java
jasta: Done.

stetho/src/main/java/com/facebook/stetho/common/android/FragmentCompatFramework.java
jasta: I see, I missed that detail when refactoring.  Personally, I don't think this is going to be a big deal in the end.  As you correctly noticed, there won't be a great many fragments in any given activity.  Facebook uses a lot and has only like 7 or 8 in the main chrome activity.

stetho/src/main/java/com/facebook/stetho/common/android/FragmentCompatSupportLib.java
jasta: Done.

stetho/src/main/java/com/facebook/stetho/common/android/FragmentCompatSupportLib.java
jasta: I'll rename.

jasta:
Hmm, really not sure why this build keeps failing on Travis CI.  It's working just fine locally.

rickbrew:
So what's the story here? Is just Travis having a bad week or something?

jasta:
I really don't know.  I'm gonna dig deeper today to figure this out.  It's
a persistent failure though it seems like it's using virtually the same
environment as me locally.

hzsweers:
Travis has been really iffy the past week or so. No real explanation, but this is what worked for me in a PR I had on another project.

```
android:
  components:
    - build-tools-22.0.1
    - android-22
    - extra-android-support
    - extra-android-m2repository
```

Key line might be `extra-android-support`. I've been told it's not necessary as long as you have `extra-android-m2repository`, but I had to have it to get the PR passing

jasta:
Success!

PR #157, opened 2015-04-11
Fix crash when rapidly adding/removing SharedPreferences keys

### LINKED ISSUE #151: Crash when switching activities: prefs change
Getting the following trace when switching activities while using Stetho (only while actively inspecting).

```
04-09 17:42:20.042    1697-1697/com.test.app E/AndroidRuntime﹕ FATAL EXCEPTION: main
    Process: com.adobe.dps.viewer, PID: 1697
    java.lang.RuntimeException: Prefs change detected for unknown key=ADMS_PauseDate
            at com.facebook.stetho.inspector.domstorage.DOMStoragePeerManager$DevToolsSharedPreferencesListener.onSharedPreferenceChanged(DOMStoragePeerManager.java:136)
            at android.app.SharedPreferencesImpl$EditorImpl.notifyListeners(SharedPreferencesImpl.java:476)
            at android.app.SharedPreferencesImpl$EditorImpl.access$600(SharedPreferencesImpl.java:302)
            at android.app.SharedPreferencesImpl$EditorImpl$3.run(SharedPreferencesImpl.java:484)
            at android.os.Handler.handleCallback(Handler.java:739)
            at android.os.Handler.dispatchMessage(Handler.java:95)
            at android.os.Looper.loop(Looper.java:135)
            at android.app.ActivityThread.main(ActivityThread.java:5221)
            at java.lang.reflect.Method.invoke(Native Method)
            at java.lang.reflect.Method.invoke(Method.java:372)
            at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:899)
            at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:694)
```

Relevant code: https://github.com/facebook/stetho/blob/master/stetho/src/main/java/com/facebook/stetho/inspector/domstorage/DOMStoragePeerManager.java#L145

The preference change is being triggered by a third-party analytics library we're using so I unfortunately don't know much about the code.  I have been able to deduce the following, though:

1. If I leave the app by tapping the home button I see the `ADMS_PauseDate` is created and stored.  No crash, Stetho continues to work, and I can see the preference in Local Storage in Chrome.
1. If I return to the app `ADMS_PauseDate` is removed from the prefs (reflected by Stetho).  Still no crash.
1. If I switch between activities in my app it crashes.

I'd guess that when switching activities the preference is created and then very quickly removed. Perhaps too quick for Stetho?

Whatever's happening I'd say a good solution would be to not throw the RuntimeException.  The reality is the preference never ended up in `mCopy` and isn't in `entries` so never had a chance to make it into the UI.  Maybe log a warning if you're concerned?

Thanks!

jasta:
This is a simple race condition that was overlooked during development
and manifests itself in practice via a third party analytics library
(though it is not known which one exactly).  The issue is clearly
demonstrable outside the context of this library though.

Closes #151

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/domstorage/DOMStoragePeerManager.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/domstorage/DOMStoragePeerManager.java
@@ -11,6 +11,7 @@ package com.facebook.stetho.inspector.domstorage;
 
 import android.content.Context;
 import android.content.SharedPreferences;
+import com.facebook.stetho.common.LogUtil;
 import com.facebook.stetho.inspector.console.CLog;
 import com.facebook.stetho.inspector.helper.ChromePeerManager;
 import com.facebook.stetho.inspector.helper.PeerRegistrationListener;
@@ -142,7 +143,9 @@ public class DOMStoragePeerManager extends ChromePeerManager {
             SharedPreferencesHelper.valueToString(newValue));
         mCopy.put(key, newValue);
       } else {
-        throw new RuntimeException("Prefs change detected for unknown key=" + key);
+        // This can happen due to the async nature of the onSharedPreferenceChanged callback.  A
+        // rapid put/remove as two separate commits on a background thread would cause this.
+        LogUtil.i("Detected rapid put/remove of %s", key);
       }
     }
   }


longinoa:
this LGTM but I wonder if we really need to log? I guess I can see the logic to try and combat bugs arising when developers dont see it show up.

jasta:
That's exactly why -- if a developer is attached to Stetho and this is happening I'd like Stetho to surface it somehow, even if just in a log statement.  It is unusual and likely the result of some programming error (it's inefficient at least).

PR #159, opened 2015-04-12
Improve support pre-JB for Activity detection

### LINKED ISSUE #123: getAllActivitiesHack fails on 4.1.1
See https://gist.github.com/longinoa/047d5195bc65c21dc514 for the log.

We probably want to find out what version this hack started working on and gate  the function to that call.

Failing that we should not allow this to NPE and crash the host app.

jasta:
This uses an older API (available to ICS+) to track activities instead
of relying on an ugly reflection trick known to work only on JB+ (and of
course not future proof).  There's a slight overhead increase in
Stetho.initialize as a result, but the timings added indicate it is
still very small (1-2ms on my Nexus 5).

The mechanism for tracking activities is partially exposed via
ActivityTracker which has a manual add/remove method which allows
developers to add support for pre-ICS even by hooking their own
application onCreate/onDestroy events just like HierarchyViewer's
ViewServer used in the past.

Closes #123


stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ApplicationDescriptor.java
```diff
@@ -58,106 +51,36 @@ protected Object onGetChildAt(Application element, int index) {
...
-        if (mCallbacks != null) {
-          mElement.unregisterActivityLifecycleCallbacks(mCallbacks);
-          mCallbacks = null;
-        }
-
-        mElement = null;
+    private final ActivityTracker.Listener mListener = new ActivityTracker.Listener() {
+      @Override
+      public void onActivityAdded(Activity activity) {
+        getHost().onChildInserted(mElement, null, activity);
```
rickbrew: We can't just pass 'null' for previousChild. We need to do a lookup into the list of activities to determine what the value is at the i-minus-1 position.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ActivityTracker.java
```diff
@@ -0,0 +1,209 @@
...
+    if (mAutomaticTracker == null) {
+      AutomaticTracker automaticTracker =
+          AutomaticTracker.newInstance(application, this /* tracker */);
+      if (automaticTracker != null) {
+        automaticTracker.register();
+        mAutomaticTracker = automaticTracker;
+        return true;
+      }
+    }
+    return false;
```
rickbrew: is it worth doing a call to LogUtil.i() here?

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ActivityTracker.java
```diff
@@ -0,0 +1,209 @@
...
+    mObservable.unregisterObserver(listener);
+  }
+
+  /**
+   * Start automatic tracking if we are running on ICS+.
+   *
+   * @return Automatic tracking has been started.  No need to manually invoke {@link #add} or
+   *     {@link #remove} methods.
+   */
+  public boolean beginTrackingIfPossible(Application application) {
```
rickbrew: opinion nit: I personally tend to prefer a "try prefix" naming convention, e.g. tryBeginTracking(), for something like this. Nice and succinct.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ActivityTracker.java
```diff
@@ -0,0 +1,209 @@
...
+    return mActivitiesUnmodifiable;
+  }
+
+  private static class ActivityTrackerObservable extends Observable<Listener> {
+    public void notifyActivityAdded(Activity activity) {
+      synchronized (mObservers) {
+        int N = mObservers.size();
+        for (int i = 0; i < N; i++) {
+          Listener listener = mObservers.get(i);
+          listener.onActivityAdded(activity);
```
rickbrew: calling foreign code while holding a lock ... hiss! I'm also concerned about re-entrancy here. like if we're notifying clients about onActivityAdded(), during which something else happens and a new Activity launches and we end up back inside ourself. From what I understand, Activities are launched fairly infrequently and there aren't very many of them, so it should be reasonable to copy the list in order to dispatch these notifications to listeners.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ActivityTracker.java
```diff
@@ -0,0 +1,209 @@
...
+import android.database.Observable;
+import android.os.Build;
+import android.os.Bundle;
+import android.os.Looper;
+import com.facebook.stetho.common.Util;
+
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.GuardedBy;
+import javax.annotation.concurrent.NotThreadSafe;
+import java.lang.ref.WeakReference;
```
rickbrew: not actually used?

stetho/src/main/java/com/facebook/stetho/Stetho.java
```diff
@@ -90,6 +91,11 @@ public static InitializerBuilder newInitializerBuilder(Context context) {
    * even low-end hardware without noticeably affecting performance.
    */
   public static void initialize(final Initializer initializer) {
+    // Hook activity tracking so that after Stetho is attached we can figure out what
+    // activities are present.
+    ActivityTracker.get().beginTrackingIfPossible(
```
rickbrew: should we LogUtil.i or .w if this returns false?

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ActivityTracker.java
```diff
@@ -0,0 +1,209 @@
...
+ * order for use by Stetho features.  Note that automatic tracking is not available for
+ * all versions of Android but it is possible to manually track activities using the {@link #add}
+ * and {@link #remove} methods exposed below.  Be aware that this is an easy opportunity to
+ * cause serious memory leaks in your application however.  Use with caution.
+ * <p/>
+ * Most callers can and should ignore this class, though it is necessary if you are implementing
+ * Activity tracking pre-ICS.
+ */
+@NotThreadSafe
+public class ActivityTracker {
```
rickbrew: can this class be made 'final' ?

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ActivityTracker.java
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+  public interface Listener {
+    public void onActivityAdded(Activity activity);
+    public void onActivityRemoved(Activity activity);
+  }
+
+  private static abstract class AutomaticTracker {
+    @Nullable
+    public static AutomaticTracker newInstance(Application application, ActivityTracker tracker) {
```
rickbrew: If this returns null, I'd prefer a name which makes that more apparent, e.g. tryNewInstance(). This makes is much more obvious to the caller that it might return null, and that its behavior doesn't just match the 'new' operator (which throws instead of returning null). It certainly makes it easier when reviewing code that calls a method like this.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ActivityTracker.java
```diff
@@ -0,0 +1,209 @@
...
+      if (automaticTracker != null) {
+        automaticTracker.register();
+        mAutomaticTracker = automaticTracker;
+        return true;
+      }
+    }
+    return false;
+  }
+
+  public boolean endTracking() {
```
rickbrew: do we really need to return true/false from this method? is the contract here that you should always call beginTracking() and then endTracking(), or do you only call endTracking() if beginTracking() returned true?

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AutomaticActivityTracker.java
```diff
@@ -0,0 +1,89 @@
...
+import android.app.Activity;
+import android.app.Application;
+import android.os.Build;
+import android.os.Bundle;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@TargetApi(Build.VERSION_CODES.ICE_CREAM_SANDWICH)
+public class AutomaticActivityTracker {
```
rickbrew: can we make this class 'final' ?

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ApplicationDescriptor.java
```diff
@@ -58,106 +51,36 @@ protected Object onGetChildAt(Application element, int index) {
...
-      LogUtil.w("Running on pre-ICS: must manually reload inspector when Activity changes");
-      return new ElementContextPreICS();
-    }
-  }
-
-  private abstract class ElementContext {
-    public abstract void hook(Application element);
+  private class ElementContext {
+    private final Application mElement;
+    private final ActivityTracker mActivityTracker = ActivityTracker.get();
```
rickbrew: nit: we can probably grab the ActivityTracker once in the outer class (ApplicationDescriptor). This is sort of an OCD nit because there will only ever be 1 Application, and thus only 1 ApplicationDescriptor.ElementDescriptor. Still, it's nice to have consistency for when someone else inevitably comes into this class and wonders where they should put more stuff.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AutomaticActivityTracker.java
```diff
@@ -0,0 +1,89 @@
...
+  private final Application mApplication;
+  private final ArrayList<Activity> mActivities = new ArrayList<>();
+
+  public static synchronized AutomaticActivityTracker register(Application application) {
```
rickbrew: why does this need to be synchronized?

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AutomaticActivityTracker.java
rickbrew: is it me or is this class not even used?

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ActivityTracker.java
```diff
@@ -0,0 +1,209 @@
...
+    if (mAutomaticTracker != null) {
+      mAutomaticTracker.unregister();
```
rickbrew: can/should we set mAutomaticTracker to null after calling unregister() ?

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ActivityTracker.java
```diff
@@ -0,0 +1,209 @@
...
+      return true;
+    }
+    return false;
+  }
+
+  public void add(Activity activity) {
+    Util.throwIfNull(activity);
+    Util.throwIfNot(Looper.myLooper() == Looper.getMainLooper());
```
rickbrew: We might want to wrap this pattern up into a utility method, e.g.

```
public static LooperUtil {
    public static void throwIfNotMainThread() { 
        Util.throwIfNot(Looper.myLooper() == Looper.getMainLooper());
    }
}
```

stetho/src/main/java/com/facebook/stetho/Stetho.java
jasta: Added

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ActivityTracker.java
jasta: Removed

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ActivityTracker.java
jasta: Done

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ActivityTracker.java
jasta: Prefer the caller doing this per your previous suggestion.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ActivityTracker.java
jasta: I thought it might be convenient to know if tracking was enabled when you called end.  It really doesn't matter I suppose, this method is never even called :)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ActivityTracker.java
jasta: Ah yes, we should :)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ActivityTracker.java
jasta: Hehe, yeah yeah, I'll implement the copy on modify pattern.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ActivityTracker.java
jasta: Renamed

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ApplicationDescriptor.java
jasta: Done.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ApplicationDescriptor.java
jasta: Hmm, this is what the code was doing before.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AutomaticActivityTracker.java
jasta: Oops, left behind the old implementation.  Good call.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ApplicationDescriptor.java
rickbrew: Hmm, you're right. It was. *scratches head* Okay, leave it alone but file an issue about it and assign it to me.

jasta: Come to think of it, I've seen this bug in practice and was very confused hehe

rickbrew:
Okay looks good to me, ship it (y)

PR #166, opened 2015-04-21
ignore meta-data for SQLite using WAL

frett:
when enabling WAL for SQLite the files used for journaling change.

See: https://www.sqlite.org/wal.html


jasta:
Nice, thanks :)

PR #168, opened 2015-04-21
Clarify and enforce thread safety contract of DOMProvider et. al.

### LINKED ISSUE #167: Clarify thread safety of DOMProvider et. al.
AndroidDOMProvider and Descriptors must be called into from the UI thread, although our abstraction is such that "UI thread" is an implementation detail (you could create a DOMProvider that didn't emit UI stuff, for instance). Let's make an interface to contractualize and enforce this.

Related to #114

rickbrew:
This introduces a new interface, ThreadBound, which provides methods for dealing with the UI thread safety contract that we need to enforce w.r.t. Android UI objects. We are also much more stringent about verifying that calls are being made on the correct thread.

Closes #167

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/common/android/HandlerUtil.java
+++ b/stetho/src/main/java/com/facebook/stetho/common/android/HandlerUtil.java
@@ -11,38 +11,106 @@ package com.facebook.stetho.common.android;
 
 import android.os.Handler;
 import android.os.Looper;
+import com.facebook.stetho.common.UncheckedCallable;
+import com.facebook.stetho.common.Util;
 
 public final class HandlerUtil {
   private HandlerUtil() {
   }
 
-  public static boolean postAndWait(Handler handler, final Runnable r) {
-    if (Looper.myLooper() == handler.getLooper()) {
-      r.run();
-      return true;
-    }
-
-    WaitableRunnable wrapper = new WaitableRunnable(r);
-    if (!handler.post(wrapper)) {
-      return false;
-    }
-
-    wrapper.join();
-    return true;
+  /**
+   * Checks whether the current thread is the same thread that the {@link Handler} is associated
+   * with.
+   * @return true if the current thread is the same thread that the {@link Handler} is associated
+   * with; otherwise false.
+   */
+  public static boolean checkThreadAccess(Handler handler) {
+    return Looper.myLooper() == handler.getLooper();
   }
 
-  private static final class WaitableRunnable implements Runnable {
-    private final Runnable mRunnable;
-    private boolean mIsDone;
+  /**
+   * Enforces that the current thread is the same thread that the {@link Handler} is associated
+   * with.
+   * @throws IllegalStateException if the current thread is not the same thread that the
+   * {@link Handler} is associated with.
+   */
+  public static void verifyThreadAccess(Handler handler) {
+    Util.throwIfNot(checkThreadAccess(handler));
+  }
 
-    public WaitableRunnable(Runnable runnable) {
-      mRunnable = runnable;
+  /**
+   * Synchronously executes an {@link UncheckedCallable} on the thread that this Handler is
+   * associated with, and returns its result.
+   * @param c the {@link UncheckedCallable} to execute
+   * @param <V> the return type of the {@link UncheckedCallable}
+   * @return the return value from {@link UncheckedCallable#call()}
+   * @throws RuntimeException if the {@link UncheckedCallable} could not be executed (the cause
+   * will be null), or if {@link UncheckedCallable#call()} threw an exception (the cause will be the
+   * exception that it threw).
+   */
+  public static <V> V postAndWait(Handler handler, final UncheckedCallable<V> c) {
+    if (checkThreadAccess(handler)) {
+      try {
+        return c.call();
+      } catch (Exception e) {
+        throw new RuntimeException(e);
+      }
+    }
+
+    WaitableRunnable<V> wrapper = new WaitableRunnable<V>() {
+      @Override
+      protected V onRun() {
+        return c.call();
+      }
+    };
+
+    return wrapper.invoke(handler);
+  }
+
+  /**
+   * Synchronously executes a {@link Runnable} on the thread that this Handler is associated with.
+   * @param r the {@link Runnable} to execute
+   * @throws RuntimeException if the {@link Runnable} could not be executed (the cause will be
+   * null), or if {@link Runnable#run()} threw an exception (the cause will be the exception that
+   * it threw).
+   */
+  public static void postAndWait(Handler handler, final Runnable r) {
+    if (checkThreadAccess(handler)) {
+      try {
+        r.run();
+        return;
+      } catch (RuntimeException e) {
+        throw new RuntimeException(e);
+      }
+    }
+
+    WaitableRunnable<Void> wrapper = new WaitableRunnable<Void>() {
+      @Override
+      protected Void onRun() {
+        r.run();
+        return null;
+      }
+    };
+
+    wrapper.invoke(handler);
+  }
+
+  private static abstract class WaitableRunnable<V> implements Runnable {
+    private boolean mIsDone;
+    private V mValue;
+    private Exception mException;
+
+    protected WaitableRunnable() {
     }
 
     @Override
-    public void run() {
+    public final void run() {
       try {
-        mRunnable.run();
+        mValue = onRun();
+        mException = null;
+      } catch (Exception e) {
+        mValue = null;
+        mException = e;
       } finally {
         synchronized (this) {
           mIsDone = true;
@@ -51,7 +119,23 @@ public final class HandlerUtil {
       }
     }
 
-    public void join() {
+    protected abstract V onRun();
+
+    public V invoke(Handler handler) {
+      if (!handler.post(this)) {
+        throw new RuntimeException("Handler.post() returned false");
+      }
+
+      join();
+
+      if (mException != null) {
+        throw new RuntimeException(mException);
+      }
+
+      return mValue;
+    }
+
+    private void join() {
       synchronized (this) {
         while (!mIsDone) {
           try {
--- a/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
@@ -14,6 +14,7 @@ package com.facebook.stetho.inspector.protocol.module;
 import android.graphics.Color;
 
 import com.facebook.stetho.common.LogUtil;
+import com.facebook.stetho.common.UncheckedCallable;
 import com.facebook.stetho.common.Util;
 import com.facebook.stetho.inspector.elements.AttributeAccumulator;
 import com.facebook.stetho.inspector.elements.DOMProvider;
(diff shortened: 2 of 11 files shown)


stetho/src/main/java/com/facebook/stetho/common/android/HandlerUtil.java
```diff
@@ -11,38 +11,104 @@
...
+  }
+
+  /**
+   * Enforces that the current thread is the same thread that the {@link Handler} is associated
+   * with.
+   * @throws IllegalStateException if the current thread is not the same thread that the
+   * {@link Handler} is associated with.
+   */
+  public static void verifyAccess(Handler handler) {
+    if (!checkAccess(handler)) {
```
jasta: I believe you can use `Util.throwIfNot`

stetho/src/main/java/com/facebook/stetho/common/android/HandlerUtil.java
```diff
@@ -11,38 +11,104 @@
...
-    if (Looper.myLooper() == handler.getLooper()) {
-      r.run();
-      return true;
+  /**
+   * Checks whether the current thread is the same thread that the {@link Handler} is associated
+   * with.
+   * @return true if the current thread is the same thread that the {@link Handler} is associated
+   * with; otherwise false.
+   */
+  public static boolean checkAccess(Handler handler) {
```
jasta: Strangely named method IMO.  Kind of a strange thing to expose publicly as well.  It's not significantly less code than writing the below expression and it's not more clear than the expression...

stetho/src/main/java/com/facebook/stetho/common/android/HandlerUtil.java
```diff
@@ -11,38 +11,104 @@
...
+   * @throws RuntimeException if the {@link Runnable} could not be executed (e.g.
+   * {@link Handler#post(Runnable)} returned false), or if {@link Runnable#run()} threw an exception
+   * (in which case the cause will be the actual exception that was thrown).
+   */
+  public static void postAndWait(Handler handler, final Runnable r) {
+    if (checkAccess(handler)) {
+      try {
+        r.run();
+        return;
+      } catch (Exception e) {
```
jasta: Runnable's don't have checked exceptions and don't need this statement.

stetho/src/main/java/com/facebook/stetho/common/android/HandlerUtil.java
```diff
@@ -11,38 +11,104 @@
...
+   * @throws RuntimeException if the {@link Callable} could not be executed (e.g.
+   * {@link Handler#post(Runnable)} returned false}, or if {@link Callable#call()} threw an
+   * exception (in which case the cause will be the actual exception that was thrown).
+   */
+  public static <V> V callAndWait(Handler handler, Callable<V> c) {
+    if (checkAccess(handler)) {
+      try {
+        return c.call();
+      } catch (Exception e) {
+        throw new RuntimeException(e);
```
jasta: See `ExceptionUtil.propagate`.  It will leave it as a RuntimeException if that's how it originally was defined.

stetho/src/main/java/com/facebook/stetho/common/android/HandlerUtil.java
```diff
@@ -11,38 +11,104 @@
...
+    }
+
+    WaitableCallable<V> wrapper = new WaitableCallable<>(c);
+    if (!handler.post(wrapper)) {
+      throw new RuntimeException("Handler.post() returned false");
+    }
+
+    wrapper.join();
+    if (wrapper.isException()) {
+      throw new RuntimeException(wrapper.getException());
```
jasta: Here too.

stetho/src/main/java/com/facebook/stetho/common/android/HandlerUtil.java
```diff
@@ -11,38 +11,104 @@
...
+    }
+
+    return wrapper.getResult();
```
jasta: Prefer a more concise utility that does all three of these things:
```java
getResultAndJoin();
```
It can throw as well.

stetho/src/main/java/com/facebook/stetho/common/android/HandlerUtil.java
```diff
@@ -62,4 +130,56 @@ public void join() {
       }
     }
   }
+
+  private static final class WaitableCallable<V> extends AbstractWaitableRunnable {
```
jasta: You can get away with writing less code by treating a Runnable as a Callable.  You're already allocating one extra object for the waitable version, might as well add another to make your code about half as large :)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/FragmentDescriptor.java
```diff
@@ -64,6 +68,8 @@ protected int onGetChildCount(Object element) {
 
   @Override
   protected Object onGetChildAt(Object element, int index) {
+    verifyAccess();
```
jasta: I think this is way overzealous.  I believe you should enforce this only at the root node of the tree and then just document that you must not leak or manipulate children views manually without also taking care of thread safety and ensuring you don't leak.  You're making one of the ways this is dangerous safe, but leaving all the other ways around.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -72,44 +72,50 @@ public void disable(JsonRpcPeer peer, JSONObject params) {
...
+      LogUtil.w("DOM.highlightNode was not given a color to highlight with");
+      return;
     }
+
+    mDOMProvider.postAndWait(new Runnable() {
+      @Override
+      public void run() {
+        Object element = mObjectIdMapper.getObjectForId(request.nodeId);
+        if (element != null) {
+          mDOMProvider.highlightElement(element, contentColor.getColor());
```
jasta: Doesn't this also punt to the main thread internally?  I would prefer to avoid adding extra round trips through the main thread since this can be highly contended.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -123,7 +129,7 @@ public ResolveNodeResponse resolveNode(JsonRpcPeer peer, JSONObject params)
     Runtime.RemoteObject remoteObject = new Runtime.RemoteObject();
     remoteObject.type = Runtime.ObjectType.OBJECT;
     remoteObject.subtype = Runtime.ObjectSubType.NODE;
-    remoteObject.className = remoteObject.getClass().getName();
+    remoteObject.className = object.getClass().getName();
```
jasta: Please fix in a separate diff for clarity.

stetho/src/main/java/com/facebook/stetho/common/android/HandlerUtil.java
rickbrew: WPF uses this extensively and I found it to work quite well: https://msdn.microsoft.com/en-us/library/system.windows.threading.dispatcherobject.checkaccess%28v=vs.110%29.aspx

Perhaps it would work better with a name like checkThreadAccess(). It is a lot less code when you have to repeat the call in many places; it's intended to be used non-sparingly in order to enforce that code is running on the right thread. It's not really different in principle than something like Util.throwIfNull().

stetho/src/main/java/com/facebook/stetho/common/android/HandlerUtil.java
rickbrew: The point here is to have a consistent exception handling contract for callers. You only ever have to catch a RuntimeException and look at getCause(). You don't need a filter for exceptions A, B, C in the case it ran on this thread, and then a separate clause for RuntimeException where you call getCause() and then check to see if it's of type A, B, or C. You only need the latter.

stetho/src/main/java/com/facebook/stetho/common/android/HandlerUtil.java
rickbrew: See comment above. If ExceptionUtil.propagate() works cross-thread, then I'm happy to use that. My goal is consistency for the caller and their catch clause, if they choose to worry about that.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/FragmentDescriptor.java
rickbrew: Over-zealous? No, I certainly disagree -- if this code doesn't execute on the UI thread, then super bad things can occur, and in subtle ways that aren't really deterministic. This over-zealousness has saved me on many, many occasions over the years in other UI code.

However, I think requiring every class to do this is a bit burdensome. I'm thinking that having something like AndroidDescriptor deriving from ChainedDescriptor and then it'll do the verifyAccess() before passing off to the derived class. At the very least it gives us a central place to implement this policy, rather than farming it out all over the place.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: 1) That's not the contract established by DOMProvider. It requires all calls into it be made on the thread that is stipulated by check/verifyAccess().

2) Yes but no (re: "doesn't this also punt to the main thread internally"). If you look at the code for AndroidDOMProvider.highlightElement(), it makes two calls after verifyAccess(). First it calls getHighlightingView(), which must run on the UI thread. Then it calls mHighlighter.clearHighlight() or setHighlightedView(). It doesn't marshal to the UI thread by itself. Those two methods on the highlighter then do their own post-ing but that's for debouncing purposes.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: Gotcha. Filed #170 for that

stetho/src/main/java/com/facebook/stetho/common/android/HandlerUtil.java
rickbrew: I was actually about to lean in the other direction, by making a new base class that the caller could use in order to *save* an allocation. Let's call it WaitableCallable<V> and it'll implement Callable<V>, Runnable, and the code for join() / getResult().

stetho/src/main/java/com/facebook/stetho/common/android/HandlerUtil.java
rickbrew: (y)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/FragmentDescriptor.java
rickbrew: Also, what specifically are you worried about w.r.t. performance? If we look at the chain, `verifyAccess()` just calls `checkAccess()` (trivial to inline), which then calls `HandlerUtil.checkAccess()` (another trivial inline), which then calls two methods, `Looper.myLooper()` and `handler.getLooper()` (another trivial method). The only thing that seems worrisome here is that `Looper.myLooper()` makes use of thread-local storage via `ThreadLocal<T>.get()`. Is that what you're concerned about? I'm not yet familiar with the cost of TLS on Android.

rickbrew: Summary of outside-of-github conversation: We're going to call verifyAccess() in DOMProvider.getDescriptor(), but not in the NodeDescriptor method calls. This will remove the TLS access cost and the verifyAccess() repetition. NodeDescriptors should then call verifyAccess() in 'dangerous' functions, such as anything that *writes* to them (e.g. setAttributesAsText), particularly if they're low-volume. NodeDescriptor will still implement ThreadBound.

jasta:
Looks good, squash and merge.

PR #172, opened 2015-04-25
Integration with Timber

fourlastor:
This is an implementation of `Timber.Tree` that sends the messages to the chrome developer console when logging with `Timber.*`

So after doing `Timber.plant(new StethoTree())` we will have something like this:

schermata del 2015-04-25 22 42 43

This could be useful to avoid flakiness of Android Studio's logcat viewer (sometimes it just ignores the logs for a while), while still keeping a live filtering functionality (which for example is difficult in solutions like [pidcat](https://github.com/JakeWharton/pidcat)).


stetho-timber/src/main/java/com/facebook/stetho/timber/StethoTree.java
```diff
@@ -0,0 +1,54 @@
...
+        logLevel = Console.MessageLevel.WARNING;
+        break;
+      case Log.ERROR:
+        logLevel = Console.MessageLevel.ERROR;
+        break;
+      default:
+        logLevel = Console.MessageLevel.LOG;
+    }
+
+    CLog.writeToConsole(
```
jasta: You've stumbled upon an unfortunate weakness in our API.  We don't really want to be using the NetworkPeerManager here (it's a subtle point admittedly but this may not be the set of peers who are observing console messages).

Instead, we want to make a new `ConsolePeerManager` which has a similar hacky API (`getInstanceOrNull`) and that is used by `Console` (peers are added/removed in `Console.enable`/`disable`.  Strictly speaking this means your log messages will be forwarded only to those peers which have invoked `Console.enable`, not necessarily those that called `Network.enable`.

jasta:
This is super useful thanks!  My only comment is a fairly nit picky one, though it should be easy to address in your patch.  Looks good otherwise.

fourlastor:
Good to go, I annotated `CLog#writeToConsole` to have the peerManager as not null, since it's expected to be so in the method implementation.

I'm wondering if it would be a good idea to move `ConsolePeerManager#getInstanceOrNull` inside the `writeToConsole` method tho, instead of checking for null everywhere and passing it afterwards.

stetho-timber/src/main/java/com/facebook/stetho/timber/StethoTree.java
```diff
@@ -0,0 +1,54 @@
...
+
+package com.facebook.stetho.timber;
+
+import android.util.Log;
+import com.facebook.stetho.inspector.console.CLog;
+import com.facebook.stetho.inspector.console.ConsolePeerManager;
+import com.facebook.stetho.inspector.protocol.module.Console;
+import timber.log.Timber;
+
+public class StethoTree extends Timber.Tree {
```
jasta: Maybe some javadocs explaining usage?  Maybe a one-liner <pre> statement showing how to plant the stetho tree in timber?

stetho/src/main/java/com/facebook/stetho/inspector/console/CLog.java
```diff
@@ -20,7 +22,7 @@
   private static final String TAG = "CLog";
 
   public static void writeToConsole(
-      ChromePeerManager chromePeerManager,
+      @Nonnull ChromePeerManager chromePeerManager,
```
jasta: Technically `@Nonnull` is the default policy; prefer to drop this as a result.

jasta:
Looks good to me.  I do agree that we probably should just offer a convenience form of `CLog#writeToConsole` which assumes you mean `ConsolePeerManager`.   I'd be fine with that change in this diff or punting on it, doesn't really matter too much.

Take a look through the final nits and update as you see fit then squash into a single commit and force push.  I'll merge when you're ready.

stetho/src/main/java/com/facebook/stetho/inspector/console/CLog.java
fourlastor: Ok, I added the annotation because I noticed without it Intellij won't give warnings to developers while passing null for example, maybe it's worth investigating why they took this decision.

fourlastor:
Squashed and added the convenience method to `CLog#writeToConsole`, not sure why the build can't find the android-test-plugin honestly.

fourlastor:
Apparently Travis CI has been having problems.

https://twitter.com/traviscistatus/status/592902357144498176

jasta:
Looks good.  I'll kick Travis in a little bit and merge once it passes.

PR #175, opened 2015-05-01
Introduce plugin framework for Console tab evaluation support

jasta:
This allows third parties to provide their own REPL implementation to
interpret commands from the Console.  Eventually we'd like to explore
possibly using bsh or some other Java evaluation framework that is
enabled if detected in the classpath.  This is the first small step in
that direction.

Tested that exception processing and arbitrary object return types do
indeed work.

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/console/RuntimeRepl.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/console/RuntimeRepl.java
@@ -1,0 +1,20 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.console;
+
+/**
+ * Allows callers to specify their own Console tab REPL for the DevTools UI.  This is part of
+ * early support for a possible optionally included default implementation for Android.
+ * <p />
+ * Experimental API.  Depend on it at your own risk...
+ */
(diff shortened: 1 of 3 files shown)


jasta:
Cc @potyl.  We're looking at messing with this soon so it made sense to sneak this support into the next release.  Hopefully we can collaborate :)

potyl:
I will try to rebase my work on top of this next week. I didn't had too much time to work on the project this week. I did some cleanup where I moved the JavaScript REPL implementation into its own project. This should facilitate rebases.

I think that the RuntimeRepl interface might need to be improved. Variable and method bindings might have to be added in there. I didn't get to that point yet, that will be my next step.

jasta:
Agreed that the API will likely need to become more sophisticated.  Especially since there are all kinds of crazy features available like code complete and "transmitting" variables from other tabs in the UI.  There is also this jQuery-like "command line" syntax that Chrome supports which lets you navigate the view hierarchy interactively with the Console which would be nice to support but will probably require some level of cooperation with the `RuntimeRepl`.

This is mostly intended to get things started so we can begin maturing one particular implementation without major churn in `Runtime.java`.

potyl:
Perfect, I was also thinking that it should be possible to send multiple "responses" through the wire back to the chrome inspector.

That would allow us to have eval code that could call "console.log()" from the eval code and to send back the proper console payload while allowing us to return the results of the evaluated expression too.

I haven't looked a lot at the chrome protocol spec but i would guess that this should be possible. Chrome does it already with google chrome android apps.

jasta:
Someone recently wired up a `CLog.writeToConsole` method signature which makes it possible to have the code you call into invoke this predictably.  You don't really have multiple responses, it's more that the protocol is re-entrant by design (while processing the evaluate JSON-RPC call, we can invoke other JSON-RPC calls to the peer such as to send console log data).

PR #177, opened 2015-05-01
Fix crash when the support library is present but not used for a particular Activity

### LINKED ISSUE #176: Crash due to FragmentCompat not checking instance types
FragmentCompat doesn't validate that the Activity instance is a FragmentActivity from the support library before casting it as such:

https://gist.github.com/anonymous/755add0f0749df986e54

This was recently introduced by #156

jasta:
This is done by simply checking whether we have an instance of
FragmentActivity before taking the support library path.  The framework
path doesn't need this check since it's impossible that if we have an
Activity and we are on a sufficiently high API level that the Activity
doesn't have fragment support.

Closes #176

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/common/android/FragmentCompatUtil.java
+++ b/stetho/src/main/java/com/facebook/stetho/common/android/FragmentCompatUtil.java
@@ -26,13 +26,20 @@ public final class FragmentCompatUtil {
   @Nullable
   private static Object findFragmentForViewInActivity(Activity activity, View view) {
     FragmentCompat supportLib = FragmentCompat.getSupportLibInstance();
-    if (supportLib != null) {
+
+    // Try the support library version if it is present and the activity is FragmentActivity.
+    if (supportLib != null &&
+        supportLib.getFragmentActivityClass() == activity.getClass()) {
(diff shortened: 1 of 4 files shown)


jasta:
@rickbrew didn't properly test this in `stetho-sample` and it bit me :)

Should be a simple fix then we should be ready to roll the 1.1.1 release.

stetho/src/main/java/com/facebook/stetho/common/android/FragmentCompatUtil.java
```diff
@@ -26,13 +26,20 @@ public static Object findFragmentForView(View view) {
   @Nullable
   private static Object findFragmentForViewInActivity(Activity activity, View view) {
     FragmentCompat supportLib = FragmentCompat.getSupportLibInstance();
-    if (supportLib != null) {
+
+    // Try the support library version if it is present and the activity is FragmentActivity.
+    if (supportLib != null &&
+        supportLib.getFragmentActivityClass() == activity.getClass()) {
```
rickbrew: This should be an isAssignableFrom() call, not a == comparison between Class<?>'s, right?

jasta: Oops

PR #183, opened 2015-05-04
Fix Fragment support in Elements tab

### LINKED ISSUE #182: Fragments not showing up in Elements tab
We access the Fragments via FragmentManager and we have to use reflection because there's no public API for it. The field is called mAdded and actually sits on a derived class called FragmentManagerImpl. Unfortunately we are trying to grab the Field off the base class, which of course fails. This code is in `FragmentCompat.FragmentManagerAccessorViaReflection`.

rickbrew:
This one is deceptively fragile because FragmentManager is an abstract class and we actually need to pull mAdded from the derived class FragmentManagerImpl.

Closes #182


jasta:
Oops, lgtm :)

PR #184, opened 2015-05-04
capture the result of trimming whitespace

frett:


rickbrew:
This looks like an easy "oops" to fix. Thanks!

stetho/src/main/java/com/facebook/stetho/inspector/database/DatabasePeerManager.java
```diff
@@ -159,7 +159,7 @@ private static String removeSuffix(String str, String[] suffixesToRemove) {
   }
 
   private static String getFirstWord(String s) {
-    s.trim();
+    s = s.trim();
```
jasta: yolo

jasta:
lgtm

PR #187, opened 2015-05-09
Use first-party test support in gradle plugin

felipecsl:
First party unit testing support is available in the android gradle build tools since version 1.1.0 ([more info](http://tools.android.com/tech-docs/new-build-system)).

This means we don't need to use a custom test plugin and can use the latest gradle build tools, which was incompatiple with the `android-unit-test` plugin.

With this, we get syntax highlighting and autocompletion support for the tests in Android Studio, which was also not available before version 1.1.0

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
-#Wed Jan 28 14:53:04 PST 2015
+#Sat May 09 15:41:17 PDT 2015
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.2.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.2.1-all.zip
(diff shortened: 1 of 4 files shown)


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Wed Jan 28 14:53:04 PST 2015
+#Sat May 09 15:41:17 PDT 2015
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.2.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.2.1-all.zip
```
jasta: Was this needed?

jasta:
Thanks for doing this :)

gradle/wrapper/gradle-wrapper.properties
felipecsl: So this is a suggestion that Android Studio makes about using "distribution with sources". I think the `-all` variant includes the gradle sources as well, so you can step into the code and look at it. Probably not needed, I can revert that change if you want

felipecsl:
:+1: :smile:

gradle/wrapper/gradle-wrapper.properties
jasta: Sounds good to me :)

Issue #81, opened 2015-03-05
Implement Console tab


jasta:
Console tab can be implemented using [bsh](http://www.beanshell.org/) (or maybe somethign more modern?).  Because bsh is large and optional, prefer a solution that treats it as a provided dep with a helpful console message if it is not included.

jasta:
@potyl I noticed your fork working on this.  Any chance you're interested in collaborating?  We started a crude version of this based on `bsh` (http://www.beanshell.org/) but are not married to it.

So far we're considering this an optional dependency so folks may be able to select their preferred runtime (or none at all if they want a small APK).  Stetho would detect this dynamically and use it if available or provide a help message in the console if not.

potyl:
I didn't noticed that there was a ticket about this. I started trying to add a REPL to stetho, my idea is to have this generic enough so that any interpreter language could be added to stetho.

As a proof of concept I started with adding a [groovy REPL](https://github.com/potyl/stetho/tree/stetho/repl-groovy) but this turned out to be quite difficult because groovy generates highly optimized JVM bytecode and Dalvik (or ART) require DEX bytecode. I got it to work by using [grooid](https://github.com/melix/grooidshell-example) but this is too slow. I will need to contact the groovy team to see if it is possible to run in full interpreted mode.

I then switched to Mozilla's Rhino, a [JavaScript REPL](https://github.com/potyl/stetho/tree/stetho/repl). That one does work better as we can run in full interpreted mode.

What's left provide a better way to hook into the Runtime module so that we can plug in any runtime. My way is kind of hackish and extending Runtime seems to be quite difficult. Whatever I did to try to plug in my own runtime always ended up breaking stetho. At the end I opted for adding a 'REPL' instead to the runtime module.

I also need to find out how to return proper data structures (objects / arrays) to chrome's console. I can return simple types just fine, complex types are another thing.  I haven't found the full spec for the return types.

I will be glad to join, specially that I will be needing help as how to integrate a plugable REPL to steho.

jasta:
I took a look at your code and I like your idea of generalizing only the REPL not Runtime (Runtime has some non-REPL stuff in it that we should keep as-is).  We've tested internally with `bsh` and know that it works and is relatively small (by comparison to, for instance, Groovy).  It does work though the syntax is not exactly greatly.  We don't have a PR for this up yet though as it's still highly experimental on our side as well.

[...]

PR #188, opened 2015-05-10
Implement Console tab support with Rhino

potyl:
This is an implementation of issue facebook/stetho#81 using Mozilla's Rhino JavaScript library.


stetho-sample/src/debug/java/com/facebook/stetho/sample/SampleDebugApplication.java
```diff
@@ -28,10 +36,33 @@ public void onCreate() {
 
     long startTime = SystemClock.elapsedRealtime();
     final Context context = this;
+
+    // Setup a javascript runtime
```
jasta: Err, accidentally commented on the diff previously.  Let's try to simplify this boilerplate (possibly with an Android-specific helper?)

jasta:
Can you post a screenshot or a sample session that shows what you can do with this?  I'm also interested in how large the dependencies are (how big is the sample APK with this change)?

stetho-sample/src/debug/java/com/facebook/stetho/sample/SampleDebugApplication.java
potyl: Only 2 lines are mandatory:

    JavaScriptRuntimeBuilder jsRuntimeBuilder = new JavaScriptRuntimeBuilder(context);
    Stetho.newInitializerBuilder(context).enableWebKitInspector(jsRuntimeBuilder.javaScriptInspectorModulesProvider());

The other lines where adding to show everything that can be done:
 - import a class
 - import a package
 - add a variable
 - add a custom function

What I should do instead is to add these code snippets to the documentation. Let me rework the patch with documentation & screenshots. I'll also comment on the dex increase, apk size, proguard settings.

potyl:
Ok, I have updated my repository with a new branch.  See the branch https://github.com/potyl/stetho/commits/stetho/repl-js.

The integration in the sample app is kept to a bare minimum. I have added a dedicated [stetho-repl-js readme](https://github.com/potyl/stetho/tree/stetho/repl-js/stetho-repl-js) for the plugin. In thee i cover examples on how to enhance the runtime, documented the APK file sizes with tips on how to proguard the app.

Mandatory screenshot:
Javascript Inspector

I also have a branch that adds this screenshot to the gh-pages, see https://github.com/potyl/stetho/tree/gh-pages.

jasta:
Wow, amazing.  A whopping 1.2MB jar file though, but as it is optional I think we can definitely work with that.

This PR is based off your repl branch, not repl-js.  Was that on purpose?  I'd like to put comments on an actual PR instead of the diffs since it's easier to manage through the review.

potyl:
I've changed the branch so that the old one would still be present just in case. For now i wanted to have your opinion. I see that i forgot to add the APK sizes, I'l add that to my documentation.

I will do a reset --hard to the original branch of the PR.

As you mentioned the APK does take a load with the javascript REPL, i would suggest that we keep it optional. For instance in our main app at work I can't enable the module since it busts the dex count even with proguard :/

potyl:
I have updated the read me to include apk sizes and tweaked a bit the documentation. The branch stetho/repl is now final.

I'm waiting for your comments.

jasta:
Can you take a look at the Travis CI failures?  Looks like simple lint errors which you can see locally by checking out the XML report generated by `gradle lint`

potyl:
I pushed a fix to lint. I don't understand why my branch still fails. The error i get is for:

```
OldTargetApi: Target SDK attribute is not targeting latest version
```

Which is triggered by:

```
defaultConfig {
    minSdkVersion 9
    targetSdkVersion 21
    versionCode 1
    versionName "1.0"
```

But all the other projects have `targetSdkVersion 21` so i don't understand why stetho-repl-js fails.
Any ideas?

potyl:
Any news on this pull request?

stetho-repl-js/README.md
```diff
@@ -0,0 +1,223 @@
...
+### Putting it together
+The javascript integration is similar to standard Stetho integration.
+The main difference is with the WebKitInspector used.
+There is a simple initialization step which occurs in your `Application` class:
+
+```java
+public class MyApplication extends Application {
+  public void onCreate() {
+    super.onCreate();
+    JavaScriptRuntimeBuilder jsRuntimeBuilder = new JavaScriptRuntimeBuilder(this);
```
jasta: nit: would be nice if you bolded the lines that are unique to the repl-js integration.  Not 100% sure what that syntax is tho when inside a preformatted code block...

stetho-repl-js/README.md
```diff
@@ -0,0 +1,223 @@
+# Stetho's Javascript Console
```
jasta: Just fantastic work on this documentation.  I'm impressed :)

stetho-repl-js/build.gradle
```diff
@@ -0,0 +1,26 @@
...
+    lintOptions {
+        // Rhino has references to awt and swing
+        disable 'InvalidPackage'
+    }
+}
+
+dependencies {
+    compile project(':stetho')
+    compile 'com.google.code.findbugs:jsr305:2.0.1'
+    compile 'org.mozilla:rhino:1.7.6'
```
jasta: It's a shame this is just one uberjar that can't be broken up a little :(

stetho-repl-js/gradle.properties
```diff
@@ -0,0 +1,4 @@
+POM_NAME=Stetho REPL JavaScript module
+POM_ARTIFACT_ID=stetho-repl-js
```
jasta: Do we maybe want to call this stetho-rhino or stetho-repl-rhino?  Using `js` here might limit us in the future (for instance, Nashorn and then even more broadly JSR-223).

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeBuilder.java
```diff
@@ -0,0 +1,248 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.console.js;
```
jasta: To avoid confusion, can we migrate these packages to `com.facebook.stetho.rhino`?  They ideally shouldn't depend on package private access into the main stetho module

stetho-repl-js/README.md
potyl: From what I've read there's not that much that we can do in code blocks besides selecting the language. I'll try to see if I can use a table with the original block on the left and the modified block on the right.

If it would work i could mimic the output of sdiff and it will help to understand what was added.

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeBuilder.java
```diff
@@ -0,0 +1,248 @@
...
+
+  /**
+   * Variables to bind to the javascript environment.
+   */
+  private final Map<String, Object> variables = new HashMap<>();
+
+  /**
+   * Global functions to add to the javascript environment.
+   */
+  private final Map<String, Function> functions = new HashMap<>();
```
jasta: style: `mFunctions` and `mVariables`

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeBuilder.java
```diff
@@ -0,0 +1,248 @@
...
+    try {
+      ScriptableObject scope = initJsScope(jsContext);
+      return scope;
+    } finally {
+      Context.exit();
+    }
+  }
+
+  private ScriptableObject initJsScope(Context jsContext) {
+    LogUtil.i("js) init JS scope");
```
jasta: This seems a little verbose to me.  Maybe drop it to .d or just remove it?

stetho-repl-js/gradle.properties
potyl: I'm ok with renaming the project/package. I might be tempted at renaming it stetho-js-rhino  (we can ditch repl). This way it is clear that it is a Javascript implementation and leaves the path to other implementations.

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeBuilder.java
```diff
@@ -0,0 +1,248 @@
...
+      }
+    }
+
+    // Import the packages that the caller requested
+    for (String packageName : mPackages) {
+      try {
+        String expression = String.format("importPackage(%s)", packageName);
+        LogUtil.i("js) init import package: %s", expression);
+        jsContext.evaluateString(scope, expression, "chrome", 1, null);
+      } catch (Exception e) {
```
jasta: Prefer to tidy up the exception handling with a simple `jsCLog` method that takes an exception and some error string.  Also note that you can extend `writeToConsole` to actually write the call stack to the JavaScript console (I _think_ we have that code in `Page` or `Runtime` today but I can't remember.  In any case, check out https://code.google.com/p/chromium/codesearch#chromium/src/third_party/WebKit/Source/devtools/protocol.json&q=protocol.json&sq=package:chromium&type=cs (see `Runtime.evaluate`)

stetho-repl-js/README.md
```diff
@@ -0,0 +1,223 @@
...
+
+```java
+public class MyApplication extends Application {
+  public void onCreate() {
+    super.onCreate();
+    JavaScriptRuntimeBuilder jsRuntimeBuilder = new JavaScriptRuntimeBuilder(this);
+    Stetho.initialize(
+        Stetho.newInitializerBuilder(this)
+            .enableDumpapp(Stetho.defaultDumperPluginsProvider(this))
+            .enableWebKitInspector(jsRuntimeBuilder.javaScriptInspectorModulesProvider())
```
jasta: Sigh, I really need to clean up this crappy extensibility API :)

I'm curious for your thoughts on something like this:

```java
.enableWebKitInspector(new CustomInspectorModulesProvider() {
  @Override
  public List<DevtoolsDomain> overrideModules() {
    return new Collections.singletonList(new Runtime(new JavaScriptRepl(context));
  }
})
```

I know it's more verbose, but it's more flexible and I think a bit more intuitive than the hacky stuff we've been recommending up until now.

stetho-repl-js/README.md
jasta: Also, we should do this construction lazily inside the inspector modules provider.  The intention of this awkward interface is to make it so that Stetho.initialize does no real work at all until a client actually connects.

stetho-repl-js/README.md
potyl: CustomInspectorModulesProvider would be provided by stetho? It does make the initialization simpler.

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeBuilder.java
```diff
@@ -0,0 +1,248 @@
...
+      }
+    }
+
+    // Define the functions
+    for (Map.Entry<String, Function> entrySet : functions.entrySet()) {
+      String functionName = entrySet.getKey();
+      Function function = entrySet.getValue();
+      try {
+        ScriptableObject.putProperty(scope, functionName, function);
+      } catch (Exception e) {
```
jasta: One thing you can do to centralize some of the exception handling as well is break this up into separate functions so the call stack includes part of the operation being carried out.  You can also do a simple custom exception class to centralize the flow control as with this simple example:

```java
void centralFunction() {
  try {
    doFoo();
    doBar();
    doBaz();
  } catch (StethoJSException e) {
    Throwable realException = e.getCause();
    String message = e.getMessage();
    jsCLog(message, realException);
  }
}

void doFoo() throws StethoJSException {
  try {
    thisCrazyRhinoThing();
  } catch (Exception e) {
    throw new StethoJSException("Error doing this crazy Rhino thing", e);
  }
}
```

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeConsole.java
```diff
@@ -0,0 +1,86 @@
...
+import org.mozilla.javascript.Scriptable;
+import org.mozilla.javascript.ScriptableObject;
+import org.mozilla.javascript.annotations.JSFunction;
+
+public class JavaScriptRuntimeConsole extends ScriptableObject {
+
+  /**
+   * Serial version UID.
+   */
+  private static final long serialVersionUID = 1L;
```
jasta: I highly doubt this is necessary :)

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeConsole.java
```diff
@@ -0,0 +1,86 @@
...
+
+  private JavaScriptRuntimeConsole thePrototypeInstance = null;
```
jasta: nit: use `@Nullable` syntax to help IntelliJ et al :)

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeConsole.java
```diff
@@ -0,0 +1,86 @@
...
+
+  /**
+   * The zero-parameter constructor.
+   *
```
jasta: nit: `<p />`

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeConsole.java
```diff
@@ -0,0 +1,86 @@
...
+   * When Context.defineClass is called with this class, it will construct
+   * File.prototype using this constructor.
+   */
+  public JavaScriptRuntimeConsole() {
+    if (thePrototypeInstance == null) {
```
jasta: Err, I don't get this at all.  An object should so very rarely need a reference to itself.  Strongly prefer separating it out so that you must pass in `ScriptableObject` but that you provide some kind of reasonable default in a separate class.

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeConsole.java
```diff
@@ -0,0 +1,86 @@
...
+      thePrototypeInstance = this;
+    }
+  }
+
+  public JavaScriptRuntimeConsole(ScriptableObject scope) {
+    setParentScope(scope);
+    Object ctor = ScriptRuntime.getTopLevelProp(scope, "Console");
```
jasta: Why is this sometimes referred to as "console" and other times as "Console"?

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeConsole.java
```diff
@@ -0,0 +1,86 @@
...
+import com.facebook.stetho.inspector.protocol.module.Console.MessageLevel;
+import com.facebook.stetho.inspector.protocol.module.Console.MessageSource;
+import org.mozilla.javascript.Context;
+import org.mozilla.javascript.Function;
+import org.mozilla.javascript.ScriptRuntime;
+import org.mozilla.javascript.Scriptable;
+import org.mozilla.javascript.ScriptableObject;
+import org.mozilla.javascript.annotations.JSFunction;
+
+public class JavaScriptRuntimeConsole extends ScriptableObject {
```
jasta: This function might need some work.  Any chance we could punt on this and rework it in a follow-up?

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeConsole.java
```diff
@@ -0,0 +1,86 @@
...
+      for (int i = 0; i < args.length; ++i) {
+        Object arg = rawArgs[i + 1];
+        arg = Context.jsToJava(arg, Object.class);
+        args[i] = arg;
+      }
+      message = String.format(format, args);
+    }
+
+    System.out.println(message);
+    CLog.writeToConsole(MessageLevel.LOG, MessageSource.JAVASCRIPT, message);
```
jasta: Hmm, technically this writes to all consoles that are attached rather than reply on the currently attached one which I think makes more sense.  We'd need to thread in the `JsonRpcPeer` to this layer to make it work properly and then initialize a new instance per peer.

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeRepl.java
```diff
@@ -0,0 +1,65 @@
...
+
+  private final ScriptableObject mJsScope;
+
+  JavaScriptRuntimeRepl(ScriptableObject scope) {
+    mJsScope = scope;
+  }
+
+  @Override
+  public Object evaluate(String expression) throws Throwable {
+      LogUtil.i("js) eval expression: %s", expression);
```
jasta: This doesn't seem useful since it will definitely give you feedback via the Chrome console

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeRepl.java
```diff
@@ -0,0 +1,65 @@
...
+
+      Object result;
+      final Context jsContext = enterJsContext();
+      try {
+        result = jsContext.evaluateString(mJsScope, expression, "chrome", 1, null);
+      } finally {
+        Context.exit();
+      }
+
+      LogUtil.i("js) Result: %s; type: %s", result, result == null ? null : result.getClass().getName());
```
jasta: Here too :)

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeRepl.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  @Override
+  public Object evaluate(String expression) throws Throwable {
+      LogUtil.i("js) eval expression: %s", expression);
+
+      Object result;
+      final Context jsContext = enterJsContext();
+      try {
+        result = jsContext.evaluateString(mJsScope, expression, "chrome", 1, null);
```
jasta: What does "chrome" mean here?

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeRepl.java
```diff
@@ -0,0 +1,65 @@
...
+   * @return a proper javascript context
+   */
+  static Context enterJsContext() {
+    LogUtil.i("js) init JS context");
+    final Context jsContext = Context.enter();
+
+    // If we cause the context to throw a runtime exception from this point
+    // we need to make sure that exit the context.
+    try {
+      jsContext.setLanguageVersion(Context.VERSION_1_8);
```
jasta: 1.8?  Err?

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeRepl.java
```diff
@@ -0,0 +1,65 @@
...
+
+      // We can't let Rhino to optimize the JS and to use a JIT because it would generate JVM bytecode
+      // and android runs on DEX bytecode. Instead we need to go in interpreted mode.
+      jsContext.setOptimizationLevel(-1);
+    }
+    catch (RuntimeException e) {
```
jasta: style nit: `} catch` on same line

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeRepl.java
potyl: We're requesting to use javascript 1.8, we have to tell rhino which version of JS to interpret since it supports quite a few.

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeRepl.java
```diff
@@ -0,0 +1,65 @@
...
+      Context.exit();
```
jasta: This doesn't look right.  You swallow the exception and proceed as if the context is usable though you exited the scope.  Also, are you sure that you need to even enter a new context for each evaluated expression anyway?  Couldn't you enter one globally and re-use it?

stetho-sample/build.gradle
```diff
@@ -26,11 +26,17 @@ android {
...
+    lintOptions {
+        // Rhino has references to awt and swing
+        disable 'InvalidPackage'
+    }
 }
 
 dependencies {
     debugCompile project(':stetho')
     debugCompile project(':stetho-urlconnection')
+    debugCompile project(':stetho-repl-js')
```
jasta: Hmm, might need to think carefully about including this.  Specifically, folks look at stetho-sample as a clue as to how big Stetho will make their APK (~100KB bigger) but with Rhino it will explode.  That could scare off some folks that don't dig into it enough to realize that Rhino is optional.

Furthermore, folks might copy the sample into their project and again not realize why they now hit the dex limit.

Perhaps a separate sample for the repl-js stuff makes sense?

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeRepl.java
potyl: "chrome" is the "file name" and 1 is the line number. Remember this is a JS interpreter, in case where it would find an error parsing the JS string the interpreter wants to show a nice error message such as:  "compilation error in `chrome` (this would be afile name) at line `1 + $lineno`"

So we need to tell the interpreter the name of the "file" and at which line number to start counting lines.

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeConsole.java
potyl: Do you have an example on how I can do that?

jasta:
Sorry for the review delay, I've been focusing on a big thorn in my side: https://github.com/facebook/stetho/pull/197 :)

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeConsole.java
potyl: We're getting pretty low with the JS integration. I copied this from one of rhino's examples. My understanding is that in JS everything is an object (a JS object) and when calling "static" functions you still get a handle to the prototype.

In order to make sure that i didn't break anything i followed their example and retrofitted it for a console.

stetho-repl-js/README.md
jasta: Don't bother re the formatting.  It's not that important :)

stetho-repl-js/README.md
jasta: Yes, I'm imagining I'd do this work in a separate PR I was just curious what you thought about Stetho providing this kind of API.  We've got other issues just like the `stetho-repl-js` which just end up looking ugly and hacky when they all get merged into one initialize step.

For instance, folks can configure which directory hosts their databases as well using the same trick that `stetho-repl-js` has to use.

stetho-repl-js/gradle.properties
jasta: I think `stetho-js-rhino` sounds awesome.

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeConsole.java
jasta: Sort of.  `CLog.writeToConsole` currently operates on the assumption that you'd want to broadcast the message to all peers associated with a particular `ChromePeerManager`.  For instance, if 2 inspector instances attach and are viewing network traffic when an unexpected stetho network issue occurs, they both receive the error in the console.

For your case, however, you really just want to send the message to the peer that wrote `console.log`.  For that, you'd need a separate `CLog.writeToConsole` method that instead accepted a `JsonRpcPeer` and instead of calling `sendNotificationToPeers` it would just call `peer.invokeMethod`.  See `ChromePeerManager#sendNotificationToPeers` for more info on how that works today.

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeRepl.java
jasta: Ack'd, sounds good.  It does appear in multiple places though so I'd prefer if it were centralized in a named constant somewhere.

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeRepl.java
jasta: I see, so this is actually JS 1.8 not Java 1.8?

jasta:
Do you mind changing the PR title to something more descriptive like "Implement Console tab support with Rhino"?  Makes things nicer when it gets merged :)

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeRepl.java
potyl: Ah, yes, *javascript*, i didn't realized that the versions of js and java matched. Now is see why you where so worried :)

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeRepl.java
potyl: We definitely need to enter the context each time since this is running on a different thread. The context needs to be enter each time that we run the JS code in a new thread. Without this each js eval call will throw an exception.

Think of it as in android when we operate on a UI thread object (ex: label.setText("foo")). When we do that android checks that the running thread is the main UI thread if not it throws a runtime exception.

Rhino does the same, it wants operations done through the interpreter form a different thread to be done by grabbing a context.

Github doesn't give me more context in the snippet but from what i remember this is the global js call that i have where I tell that i want to bytecode optimizations and to set the language to JS 1.8, the thing is that if one of these methods fails I need to ensure that I do exit the context.

The caller's pattern is:

```java
Context = enterJsContext(); // if this fails with a runtime error I need someone to close the context, but we're not in a try / finally yet
try {
  // use the context
}
finally {
  context.exit()
}
```

If I fail to build the context in enterJsContext() i still need to exit it as long as it is created. This is why inside enterJsContext() there's a catch block that exits the context.

I'll add more documentation for this.

stetho-sample/build.gradle
potyl: I'll remove it. I have documentation on my sub project. I think that we can just leave the patch that affects the main README.md and to show users how to enable javascript by themselves.

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeConsole.java
potyl: This was added by the IDE in order to avoid a warning. I've just left it there and updated the value.

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeBuilder.java
potyl: My bad, this is now fixed.

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeBuilder.java
potyl: com.facebook.stetho.rhino it is!

stetho-repl-js/gradle.properties
potyl: Renamed to:

```
POM_NAME=Stetho JavaScript (Rhino) module
POM_ARTIFACT_ID=stetho-js-rhino
```

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeBuilder.java
potyl: I'm removing all debug, I had it there to see if things worked.

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeConsole.java
potyl: I didn't understood this comment. What should I do?

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeConsole.java
potyl: I've looked into it. I believe that this is supposed to be the 'JavaScript' object class.

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeRepl.java
potyl: yeah, the logging through LogUtil is one of the first thing to land. Before I had access to the console all that I had was the logcat logs :)

I'm removing them.

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeRepl.java
potyl: fixed.

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeConsole.java
potyl: Ok, I found where I took this from [Rhino's Environment object](https://github.com/mozilla/rhino/blob/master/toolsrc/org/mozilla/javascript/tools/shell/Environment.java#L53). They setup the `environment` object in the same way (see [Rhino's Global's environment](https://github.com/mozilla/rhino/blob/master/toolsrc/org/mozilla/javascript/tools/shell/Global.java#L114).

The suspect that the upper case name is the prototype.

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeConsole.java
potyl: Can you elaborate more? I want to add more methods to the class so that it implements most of (or all of):

 - [Chrome's console](https://developer.chrome.com/devtools/docs/console-api)
 - [Mozilla's console](https://developer.mozilla.org/en/docs/Web/API/console)

jasta:
Any more updates?  I think with the latest round of feedback implemented this should be merge-able.

potyl:
I didn't finish going through all comments/remarks. Let me review the branch this evening and merge the commits. I now have a bit more free time and I can work on this.

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeBuilder.java
potyl: I couldn't find the code to extend writeToConsole.

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeConsole.java
potyl: Done

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeRepl.java
potyl: I reviewed the code and i do believe that i was missing a ```throw e``` statement in there. It's now added.

stetho-repl-js/src/main/java/com/facebook/stetho/inspector/console/js/JavaScriptRuntimeConsole.java
potyl: I'm a bit lost with the JsonRpcPeer. How can I get an instance to my corresponding peer? I can see this in Runtime.java:

```java
  @ChromeDevtoolsMethod
  public JsonRpcResult evaluate(JsonRpcPeer peer, JSONObject params) {
    return getSession(peer).evaluate(mReplFactory, params);
  }
```

So Runtime.java has access to a peer but evaluate() doesn't pass the peer around. Furthermore in the code linked with this review (JavaScriptRuntimeConsole.java) I don't have access to peer yet.

potyl:
Ok, I've rebased all my  changes and removed the commit that modifies stetho-sample so that it doesn't use the javascript repl (I believe that this was requested at some point).

Code can still be improved at some point:
 - use a custom writeToConsole
 - find a way to get a handle to JsonRpcPeer for the console output

I still believe that as a preview the current basis is solid enough to be used, so we could merge this.

stetho-repl-js/gradle.properties
jasta: Looks like this rename never occurred?  I'm fine with it as-is, but I will probably rename before we cut the release if that's OK.

jasta:
Oh I see, you actually changed the branch.  It looks like you're really expecting to merge https://github.com/potyl/stetho/tree/stetho/js-rhino, is that right?  If so, can you just push to the original `repl` branch from `js-rhino` so I can merge with github's UI?

potyl:
My bad! Once I get home i'll update the proper branch

On Friday, July 17, 2015, Josh Guilfoyle <notifications@github.com> wrote:

> Oh I see, you actually changed the branch. It looks like you're really
> expecting to merge https://github.com/potyl/stetho/tree/stetho/js-rhino,
> is that right? If so, can you just push to the original repl branch from
> js-rhino so I can merge with github's UI?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/facebook/stetho/pull/188#issuecomment-122222896>.
>


-- 
Emmanuel Rodriguez

potyl:
I've pushed to the proper branch. Turns out that I also confused the commits and I previously left the commit to add the javascript to the stetho-sample app and removed the documentation on how to integrate it. This is also now corrected in the stetho/repl branch.

stetho-js-rhino/README.md
```diff
@@ -0,0 +1,224 @@
...
+### Download
+Download [the latest JARs](https://github.com/facebook/stetho/releases/latest) or grab via Gradle:
+```groovy
+compile 'com.facebook.stetho:stetho-repl-js:1.1.1'
+```
+or Maven:
+```xml
+<dependency>
+  <groupId>com.facebook.stetho</groupId>
+  <artifactId>stetho-repl-js</artifactId>
```
jasta: should be `stetho-js-rhino`

stetho-js-rhino/README.md
```diff
@@ -0,0 +1,224 @@
+# Stetho's JavaScript Module
+
+This [Stetho](https://facebook.github.io/stetho) plugin adds a JavaScript console by embedding Mozilla's [Rhino](https://github.com/mozilla/rhino).
+
+## Set-up
+
+### Download
+Download [the latest JARs](https://github.com/facebook/stetho/releases/latest) or grab via Gradle:
+```groovy
+compile 'com.facebook.stetho:stetho-repl-js:1.1.1'
```
jasta: here too

jasta:
Very minor documentation comments; this is otherwise ready to merge.  Great work, I'm so excited for this feature!

potyl:
Thanks for spotting the 2 documentation errors. I've squashed the fix and ran a grep. No more occurrences of the old repo name.

jasta:
Excellent, I'm AFK this weekend but will merge when I get back.  Awesome
stuff :)

potyl:
Thanks a lot!
Cheers.

ubaierbhat:
[EvaluatorException for sql query](https://github.com/facebook/stetho/issues/546) is this caused by Rhino?

Issue #189, opened 2015-05-14
Implement Screencasting


auchenberg:
Chrome DevTools supports [screencasting](http://blog.chromium.org/2013/12/chrome-devtools-for-mobile-emulate-and.html) of the screen of the attached device. Since Stetho is using the embedded Chrome DevTools via ``chrome://inspect `` it's straightforward to implement, as DevTools is already launched with the correct parameters.

[![ScreenShot](http://img.youtube.com/vi/Q7rEFEMpwe4/maxresdefault.jpg)](http://youtu.be/Q7rEFEMpwe4)

#### Steps to implement:

- Reply ``{ result: true }`` to the ``Page.canScreencast`` method request made from DevTools.

- Implement logic to capture screenshot of running Android app as JPEG, base64 encode the image and emit a ``Page.screencastFrame`` event, with the following data-structure:

```javascript
{
    method: 'Page.screencastFrame',
    params: {
        data: <base64 image>,
        metadata: {
            pageScaleFactor: 1,
            offsetTop: 0,
            deviceWidth: <width>,
            deviceHeight: <height>,
            scrollOffsetX: 0,
            scrollOffsetY: 0
        }
    }
}
```

jasta:
You seem to have implemented this already based on the screenshot?  Should
I expect a PR? :)

auchenberg:
Nah, its a screenshot from the introduction video showing Paul Irish demoing it in Chrome on Android.

I'm not a Java guy so I can't help implement the server part, but I can assist on the DevTools part :)

felipecsl:
I up for implementing this. Will give it a try and ping you if I have any DevTools questions, @auchenberg  :)

facebook-github-bot-7:
Thank you for reporting this issue and appreciate your patience. We've  notified the core team for an update on this issue. We're looking for a response within the next 30 days or the issue may be closed.

auchenberg:
This can be closed now.

PR #190, opened 2015-05-15
Screencasting support

felipecsl:
Tries to implement screencasting support as described in #189 
It is still a bit janky but works. I'll try to get it improved.

screen shot 2015-05-14 at 11 44 36 pm

Still needs to be fixed:

 * When the app is closed while the screencast is running, we get a crash since the bitmap is disposed and we never got a chance to set `mIsRunning` to false. Need to find a way to detect that the activity is closing, maybe `isFinishing()` would do.
 * Jankiness

Nice to have:

 * Touch support

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ActivityTracker.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ActivityTracker.java
@@ -111,6 +111,13 @@ public final class ActivityTracker {
     return mActivitiesUnmodifiable;
   }
 
+  public Activity tryGetTopActivity() {
+    if (mActivitiesUnmodifiable.isEmpty()) {
+      return null;
+    }
+    return mActivitiesUnmodifiable.get(mActivitiesUnmodifiable.size() - 1);
+  }
+
   public interface Listener {
     public void onActivityAdded(Activity activity);
     public void onActivityRemoved(Activity activity);
--- a/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Page.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Page.java
@@ -9,25 +9,30 @@
 
 package com.facebook.stetho.inspector.protocol.module;
 
+import android.content.Context;
+import android.support.annotation.Nullable;
+import com.facebook.stetho.common.ProcessUtil;
+import com.facebook.stetho.inspector.domstorage.SharedPreferencesHelper;
+import com.facebook.stetho.inspector.jsonrpc.JsonRpcPeer;
+import com.facebook.stetho.inspector.jsonrpc.JsonRpcResult;
+import com.facebook.stetho.inspector.protocol.ChromeDevtoolsDomain;
+import com.facebook.stetho.inspector.protocol.ChromeDevtoolsMethod;
+import com.facebook.stetho.inspector.screencast.ScreencastDispatcher;
+import com.facebook.stetho.json.ObjectMapper;
+import com.facebook.stetho.json.annotation.JsonProperty;
+import com.facebook.stetho.json.annotation.JsonValue;
+import org.json.JSONObject;
+
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 
-import android.content.Context;
-import com.facebook.stetho.common.ProcessUtil;
-import com.facebook.stetho.inspector.jsonrpc.JsonRpcPeer;
-import com.facebook.stetho.inspector.jsonrpc.JsonRpcResult;
-import com.facebook.stetho.inspector.protocol.ChromeDevtoolsDomain;
-import com.facebook.stetho.inspector.protocol.ChromeDevtoolsMethod;
-import com.facebook.stetho.inspector.domstorage.SharedPreferencesHelper;
-import com.facebook.stetho.json.annotation.JsonProperty;
-import com.facebook.stetho.json.annotation.JsonValue;
-
-import org.json.JSONObject;
-
 public class Page implements ChromeDevtoolsDomain {
   private final Context mContext;
+  private final ObjectMapper mObjectMapper = new ObjectMapper();
+  @Nullable
+  private ScreencastDispatcher mScreencastDispatcher;
 
   public Page(Context context) {
     mContext = context;
@@ -130,7 +135,7 @@ public class Page implements ChromeDevtoolsDomain {
 
   @ChromeDevtoolsMethod
   public JsonRpcResult canScreencast(JsonRpcPeer peer, JSONObject params) {
-    return new SimpleBooleanResult(false);
+    return new SimpleBooleanResult(true);
   }
 
   @ChromeDevtoolsMethod
@@ -144,6 +149,30 @@ public class Page implements ChromeDevtoolsDomain {
 
   @ChromeDevtoolsMethod
   public void clearDeviceOrientationOverride(JsonRpcPeer peer, JSONObject params) {
+  }
+
+  @ChromeDevtoolsMethod
+  public void startScreencast(final JsonRpcPeer peer, JSONObject params) {
+    final StartScreencastRequest request = mObjectMapper.convertValue(
+        params, StartScreencastRequest.class);
+    if (mScreencastDispatcher == null) {
+      mScreencastDispatcher = new ScreencastDispatcher();
+      mScreencastDispatcher.startScreencast(peer, request);
+    }
+  }
+
+  @ChromeDevtoolsMethod
+  public void stopScreencast(JsonRpcPeer peer, JSONObject params) {
+    if (mScreencastDispatcher != null) {
+      mScreencastDispatcher.stopScreencast();
+      mScreencastDispatcher = null;
+    }
+  }
+
+  @ChromeDevtoolsMethod
+  public void screencastFrameAck(JsonRpcPeer peer, JSONObject params) {
+    // Nothing to do here, just need to make sure Chrome doesn't get an error that this method
+    // isn't implemented
   }
 
   @ChromeDevtoolsMethod
@@ -238,5 +267,39 @@ public class Page implements ChromeDevtoolsDomain {
 
     @JsonProperty(required = true)
     public int id;
+  }
+
+  public static class ScreencastFrameEvent {
+    @JsonProperty(required = true)
+    public String data;
+
+    @JsonProperty(required = true)
+    public ScreencastFrameEventMetadata metadata;
+  }
+
+  public static class ScreencastFrameEventMetadata {
+    @JsonProperty(required = true)
+    public int pageScaleFactor;
+    @JsonProperty(required = true)
+    public int offsetTop;
+    @JsonProperty(required = true)
+    public int deviceWidth;
+    @JsonProperty(required = true)
+    public int deviceHeight;
+    @JsonProperty(required = true)
+    public int scrollOffsetX;
+    @JsonProperty(required = true)
+    public int scrollOffsetY;
+  }
+
+  public static class StartScreencastRequest {
+    @JsonProperty
+    public String format;
+    @JsonProperty
+    public int quality;
+    @JsonProperty
+    public int maxWidth;
+    @JsonProperty
+    public int maxHeight;
   }
 }
--- a/stetho/src/main/java/com/facebook/stetho/inspector/screencast/ScreencastDispatcher.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/screencast/ScreencastDispatcher.java
@@ -1,0 +1,154 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.screencast;
+
+import android.app.Activity;
+import android.graphics.Bitmap;
+import android.graphics.Canvas;
+import android.graphics.Matrix;
+import android.graphics.RectF;
+import android.os.Handler;
+import android.os.HandlerThread;
+import android.os.Looper;
+import android.util.Base64;
+import android.util.Base64OutputStream;
+import android.view.View;
+import com.facebook.stetho.common.LogUtil;
+import com.facebook.stetho.inspector.elements.android.ActivityTracker;
+import com.facebook.stetho.inspector.jsonrpc.JsonRpcPeer;
+import com.facebook.stetho.inspector.protocol.module.Page;
+
+import java.io.ByteArrayOutputStream;
+
+public final class ScreencastDispatcher {
+  private static final long FRAME_DELAY = 200l;
+
+  private final Handler mMainHandler = new Handler(Looper.getMainLooper());
+  private final BitmapFetchRunnable mBitmapFetchRunnable = new BitmapFetchRunnable();
+  private final ActivityTracker mActivityTracker = ActivityTracker.get();
+  private final EventDispatchRunnable mEventDispatchRunnable = new EventDispatchRunnable();
+  private final RectF mTempSrc = new RectF();
+  private final RectF mTempDst = new RectF();
+
+  private boolean mIsRunning;
+  private Handler mBackgroundHandler;
+  private JsonRpcPeer mPeer;
+  private HandlerThread mHandlerThread;
+  private Bitmap mBitmap;
+  private Canvas mCanvas;
+  private Page.StartScreencastRequest mRequest;
+  private ByteArrayOutputStream mStream;
+  private Page.ScreencastFrameEvent mEvent = new Page.ScreencastFrameEvent();
+  private Page.ScreencastFrameEventMetadata mMetadata = new Page.ScreencastFrameEventMetadata();
+
+  public ScreencastDispatcher() {
+  }
+
+  public void startScreencast(JsonRpcPeer peer, Page.StartScreencastRequest request) {
+    LogUtil.d("Starting screencast");
+    mRequest = request;
+    mHandlerThread = new HandlerThread("Screencast Thread");
+    mHandlerThread.start();
+    mPeer = peer;
+    mIsRunning = true;
+    mStream = new ByteArrayOutputStream();
+    mBackgroundHandler = new Handler(mHandlerThread.getLooper());
+    mMainHandler.postDelayed(mBitmapFetchRunnable, FRAME_DELAY);
+  }
+
+  public void stopScreencast() {
+    LogUtil.d("Stopping screencast");
+    mBackgroundHandler.post(new CancellationRunnable());
+  }
+
+  private class BitmapFetchRunnable implements Runnable {
+    @Override
+    public void run() {
+      updateScreenBitmap();
+      mBackgroundHandler.post(mEventDispatchRunnable.withEndAction(this));
+    }
+
+    private void updateScreenBitmap() {
+      if (!mIsRunning) {
+        return;
+      }
+      Activity activity = mActivityTracker.tryGetTopActivity();
+      if (activity == null) {
+        return;
+      }
+      // This stuff needs to happen in the UI thread
+      View rootView = activity.getWindow().getDecorView();
+      try {
+        if (mBitmap == null) {
+          int viewWidth = rootView.getWidth();
+          int viewHeight = rootView.getHeight();
+          float scale = Math.min((float) mRequest.maxWidth / (float) viewWidth,
+              (float) mRequest.maxHeight / (float) viewHeight);
+          int destWidth = (int) (viewWidth * scale);
+          int destHeight = (int) (viewHeight * scale);
+          mBitmap = Bitmap.createBitmap(destWidth, destHeight, Bitmap.Config.RGB_565);
+          mCanvas = new Canvas(mBitmap);
+          Matrix matrix = new Matrix();
+          mTempSrc.set(0, 0, viewWidth, viewHeight);
+          mTempDst.set(0, 0, destWidth, destHeight);
+          matrix.setRectToRect(mTempSrc, mTempDst, Matrix.ScaleToFit.CENTER);
+          mCanvas.setMatrix(matrix);
+        }
+        rootView.draw(mCanvas);
+      } catch (OutOfMemoryError e) {
+        LogUtil.w("Out of memory trying to allocate screencast Bitmap.");
+      }
+    }
+  }
+
+  private class EventDispatchRunnable implements Runnable {
+    private Runnable mEndAction;
+
+    private EventDispatchRunnable withEndAction(Runnable endAction) {
+      mEndAction = endAction;
+      return this;
+    }
+
+    @Override
+    public void run() {
+      if (!mIsRunning || mBitmap == null) {
+        return;
+      }
+      int width = mBitmap.getWidth();
+      int height = mBitmap.getHeight();
+      mStream.reset();
+      Base64OutputStream base64Stream = new Base64OutputStream(mStream, Base64.DEFAULT);
+      // request format is either "jpeg" or "png"
+      Bitmap.CompressFormat format = Bitmap.CompressFormat.valueOf(mRequest.format.toUpperCase());
+      mBitmap.compress(format, mRequest.quality, base64Stream);
+      mEvent.data = mStream.toString();
+      mMetadata.pageScaleFactor = 1;
+      mMetadata.deviceWidth = width;
+      mMetadata.deviceHeight = height;
+      mEvent.metadata = mMetadata;
+      mPeer.invokeMethod("Page.screencastFrame", mEvent, null);
+      mMainHandler.postDelayed(mEndAction, FRAME_DELAY);
+    }
+  }
+
+  private class CancellationRunnable implements Runnable {
+    @Override
+    public void run() {
+      mHandlerThread.interrupt();
+      mMainHandler.removeCallbacks(mBitmapFetchRunnable);
+      mBackgroundHandler.removeCallbacks(mEventDispatchRunnable);
+      mIsRunning = false;
+      mHandlerThread = null;
+      mBitmap = null;
+      mCanvas = null;
+      mStream = null;
+    }
+  }
+}


auchenberg:
Damn, that was quick! :smiley:

stetho/src/main/java/com/facebook/stetho/inspector/protocol/ScreencastDispatcher.java
```diff
@@ -0,0 +1,131 @@
...
+    private EventDispatchRunnable withEndAction(Runnable endAction) {
+      mEndAction = endAction;
+      return this;
+    }
+
+    @Override
+    public void run() {
+      ScreencastFrameEvent screencastFrameEvent = new ScreencastFrameEvent();
+      ByteArrayOutputStream stream = new ByteArrayOutputStream();
+      mBitmap.compress(Bitmap.CompressFormat.PNG, 100, stream);
```
auchenberg: Chrome DevTools prefixes the base64 encoded url with [``data:image/jpg;base64``](https://github.com/auchenberg/chrome-devtools-app/blob/master/app/devtools/front_end/screencast/ScreencastView.js#L152), so you should probably compress the image as JPEG.

felipecsl:
Thanks, @auchenberg, fixed. Also, do you know is the ideal framerate we should use?

auchenberg:
By looking at the implementation there doesnt seem to be a "magic" framerate predefined, but you can listen to the rate of ``screencastFrameAck``'s and adjust the frame-rate thereafter.

jasta:
At a high level I'm a bit concerned about the possible performance impact this may have on apps being inspected.  Also, how is this feature enabled in the DevTools UI?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/ScreencastDispatcher.java
```diff
@@ -0,0 +1,142 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.protocol;
```
jasta: Let's move this to `com.facebook.stetho.screencast.android`.  We're beginning to separate Stetho internals from Android-specific pieces so that Stetho can one day be used without Android (maybe *grin*)

stetho/src/main/java/com/facebook/stetho/inspector/protocol/ScreencastDispatcher.java
```diff
@@ -0,0 +1,142 @@
...
+import com.facebook.stetho.common.LogUtil;
+import com.facebook.stetho.inspector.elements.android.ActivityTracker;
+import com.facebook.stetho.inspector.jsonrpc.JsonRpcPeer;
+import com.facebook.stetho.json.annotation.JsonProperty;
+
+import java.io.ByteArrayOutputStream;
+
+public class ScreencastDispatcher {
+  private final Handler mMainHandler = new Handler(Looper.getMainLooper());
+  private final HandlerThread mHandlerThread = new HandlerThread("Screencast Thread");
```
jasta: Prefer an explicit init step that creates/starts the thread.  Constructors with side effects are often very hard to debug/predict.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/ScreencastDispatcher.java
```diff
@@ -0,0 +1,142 @@
...
+  private final BitmapFetchRunnable mBitmapFetchRunnable = new BitmapFetchRunnable();
+  private final ActivityTracker mActivityTracker = ActivityTracker.get();
+  private final EventDispatchRunnable mEventDispatchRunnable = new EventDispatchRunnable();
+  private final ScreencastFrameEvent mEvent = new ScreencastFrameEvent();
+  private final ScreencastFrameEvent.Metadata mMetadata = new ScreencastFrameEvent.Metadata();
+  private boolean mIsRunning;
+  private static final float FRAMERATE = 60f;
```
jasta: Group your variables a bit and put whitespace in between them to help readability.  Also, statics and static finals go up top :)

stetho/src/main/java/com/facebook/stetho/inspector/protocol/ScreencastDispatcher.java
```diff
@@ -0,0 +1,142 @@
...
+  private Handler mBackgroundHandler;
+  private JsonRpcPeer mPeer;
+
+  public ScreencastDispatcher() {
+    mHandlerThread.start();
+  }
+
+  public void startScreencast(JsonRpcPeer peer) {
+    LogUtil.d("Starting screencast");
+    mPeer = peer;
```
jasta: There can be more than one peer.  Look into using `ChromePeerManager`.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/ScreencastDispatcher.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    @Override
+    public void run() {
+      if (!mIsRunning) {
+        return;
+      }
+      ByteArrayOutputStream stream = new ByteArrayOutputStream();
+      mBitmap.compress(Bitmap.CompressFormat.JPEG, 80, stream);
+      mEvent.data = Base64.encodeToString(stream.toByteArray(), Base64.DEFAULT);
```
jasta: You can, and likely should, optimize this using `Base64OutputStream` instead of `ByteArrayOutputStream`.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/ScreencastDispatcher.java
```diff
@@ -0,0 +1,142 @@
...
+      mMetadata.offsetTop = 0;
```
jasta: Using a member variable here introduces a strange and hard to predict side effect.  Why are you not simply making a new instance each time?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/ScreencastDispatcher.java
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  private static class ScreencastFrameEvent {
+    @JsonProperty(required = true)
+    public String data;
+
+    @JsonProperty(required = true)
+    public Metadata metadata;
+
+    private static class Metadata {
```
jasta: Prefer not to use inner classes for this type of thing.  Also, this should be defined in the `Page` module and made public.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Page.java
```diff
@@ -9,25 +9,27 @@
...
 import org.json.JSONObject;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+
 public class Page implements ChromeDevtoolsDomain {
   private final Context mContext;
+  private ScreencastDispatcher mScreencastDispatcher = new ScreencastDispatcher();
```
jasta: Use final where applicable please :)

jasta: You probably want this to not be constructed here but rather managed lazily when a caller actually asks for screen casting.  A subclassed `ChromePeerManager` may well be what you're missing...

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Page.java
```diff
@@ -147,6 +149,21 @@ public void clearDeviceOrientationOverride(JsonRpcPeer peer, JSONObject params)
...
+  }
+
+  @ChromeDevtoolsMethod
+  public void stopScreencast(JsonRpcPeer peer, JSONObject params) {
+    mScreencastDispatcher.stopScreencast();
+  }
+
+  @ChromeDevtoolsMethod
+  public void screencastFrameAck(JsonRpcPeer peer, JSONObject params) {
+    LogUtil.i("screencastFrameAck()");
```
jasta: Prefer to drop this :)

jasta:
This is really exciting stuff, great work!

stetho/src/main/java/com/facebook/stetho/inspector/protocol/ScreencastDispatcher.java
felipecsl: Since this is a very hot code path, I decided to reuse the event instance, otherwise we'd be creating 2 new instances per frame. I had it create it every time in the first version, but once I changed it to reuse the instances, I observed a noticeable improvement in the screencasting performance.

felipecsl:
This feature can be accessed in the top right corner of the devtools window, there's a little phone icon where you can toggle screencasting on/off (see the screenshot above) - I didn't know it existed either :)

I indeed noticed a small impact on the app performance while the screencast is running, but I suppose that is somewhat expected. I'm sure there's more optimizations we can do to reduce the impact. This was just my first solution.

auchenberg:
This is a power feature that was to be enabled in the DevTools, and the screencapturing isn't started before requested, so I don't see a performance-risk in adding the feature. 

It's highly useful, as you don't have to look at the screen of the device.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/ScreencastDispatcher.java
jasta: Hmm.  That sounds a bit fishy on account of the fact that you're compressing a bitmap into a byte array output stream which means you're probably likely to allocate and copy many large arrays as the byte[] grows.  That slowdown you're seeing was probably due to the extreme CPU pressure that this operation applies independent of these re-used objects :)

In code like this, honestly, I think it's better to just admit defeat on the small stuff if you can't optimize away the big things.

jasta:
@felipecsl / @auchenberg I agree that because the optional nature of this we can regress perf a bit.  I think it's totally fine to just get this working with perhaps a warning to logcat when it turns on that it may regress app performance.  I just wanted to make sure we're paying attention to little details like that :)

stetho/src/main/java/com/facebook/stetho/inspector/protocol/ScreencastDispatcher.java
felipecsl: :) makes sense!

felipecsl:
I'm still seeing OOMs after screencasting after a couple minutes and apparently the memory doesn't get reclamed even after stopping the screencast, so we might have a leak somewhere. I'll get a heapdump later

jasta:
@felipecsl have you been able to work on a solution to the memory leaks yet?  I've very interested in getting this merged in but obviously we can't just leak memory until we OOM -- that's a pretty frustrating experience for a debug tool...

felipecsl:
@jasta definitely! I didn't have time to touch it recently but I'm still plan to fix the OOMs. I'll update it soon!

felipecsl:
Alright just had another stab at this, memory consumption looks much better now that we manage (and reuse) the bitmap ourselves (instead of using `View.getDrawingCache()`) however it's still pretty memory hungry. It doens't seem to be leaking memory or causing OOMs anymore but it obviously needs more thorough testing. Also fixed issues with the size of the bitmap that we sent back to match the requested maxWidth and maxHeight.

A couple screenshots of the current state below:

screen shot 2015-06-23 at 1 49 42 am

screen shot 2015-06-23 at 1 49 49 am

stetho/src/main/java/com/facebook/stetho/screencast/android/ScreencastDispatcher.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    @Override
+    public void run() {
+      if (!mIsRunning || mBitmap == null) {
+        return;
+      }
+      int width = mBitmap.getWidth();
+      int height = mBitmap.getHeight();
+      ByteArrayOutputStream stream = new ByteArrayOutputStream();
```
rickbrew: This code creates a ByteArrayOutputStream which houses a growable byte[] buffer. This is then copied to a new array, and then copied into a String. Lots of copying :)

I believe we can reduce allocation and churn by pooling the ByteArrayOutputStream as well and avoiding its toByteArray() method. In order to do that we'll have to make a derived class in order to access the protected 'buf' field (like with a method called getBuffer()). 

If we're willing to go all the way, we can make our serializer understand something other than 'String' for when we do the "event.data" assignment. Perhaps we can make our serializer understand CharSequence? And then we can plug in our own CharSequence implementation that references the byte[] buffer and the length we care about. (I haven't looked at the serializer yet to see how or if this could work without copying ... it would need to avoid accessing the char[] array)

felipecsl: Agreed, this is probably the reason why we're still seeing memory usage spikes during the screencasting. I'll look into making it more efficient

felipecsl:
I don't think we can easily avoid the step of converting the byte array into a char array, since both `JSONObject ` and `SimpleSession` require a String to work. If we wanted to send the byte array directly into the wire, that would require a lot of changes.

stetho/src/main/java/com/facebook/stetho/screencast/android/ScreencastDispatcher.java
rickbrew: Yeah I agree. It sure would be nice though. Let's definitely not worry about that right now!

I think we can easily reuse the ByteArrayOutputStream (and its big byte[] array), too. What about promoting 'stream' to be an instance field, mStream? You can call reset() before you write into it, which will seek to the beginning and *not* throw away the byte[] that it's already allocated (which means new ones won't be allocated too!).

rickbrew:
*bump* Are we waiting on anything else here? :)

felipecsl:
i'm not sure either :)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ActivityTracker.java
```diff
@@ -107,10 +107,17 @@ public void remove(Activity activity) {
     }
   }
 
-  public List<Activity> getActivitiesView() {
+  public List<Activity> getActivitiesList() {
```
rickbrew: not sure why this method's name was changed from `...View()` to `...List()` ?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Page.java
```diff
@@ -147,6 +150,22 @@ public void clearDeviceOrientationOverride(JsonRpcPeer peer, JSONObject params)
...
+  }
+
+  @ChromeDevtoolsMethod
+  public void stopScreencast(JsonRpcPeer peer, JSONObject params) {
+    mScreencastDispatcher.stopScreencast();
+  }
+
+  @ChromeDevtoolsMethod
+  public void screencastFrameAck(JsonRpcPeer peer, JSONObject params) {
+  }
```
rickbrew: since this is empty, it can be help to add a comment like this so that future readers don't think this is accidental, e.g.
```
// Nothing to do here, just need to make sure Chrome doesn't get an error that this method isn't implemented
```

stetho/src/main/java/com/facebook/stetho/screencast/android/ScreencastDispatcher.java
```diff
@@ -0,0 +1,149 @@
...
+    @Override
+    public void run() {
+      if (!mIsRunning || mBitmap == null) {
+        return;
+      }
+      int width = mBitmap.getWidth();
+      int height = mBitmap.getHeight();
+      mStream.reset();
+      Base64OutputStream base64Stream = new Base64OutputStream(mStream, Base64.DEFAULT);
+      mBitmap.compress(Bitmap.CompressFormat.JPEG, mRequest.quality, base64Stream);
```
rickbrew: shouldn't we honor the compression type requested in the `StartScreencastRequest`? The protocol says that both JPEG and PNG are supported

rickbrew:
Okay here's another small round of feedback, mostly nits. Otherwise it looks mostly ready to go from what I can tell

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ActivityTracker.java
felipecsl: Well.. It returns a list of activities, so I thought it made sense to say that it returns an ActivitiesList (instead of ActivitiesView). In case it was intentionally named that way I can rever that (i'm assuming it was a mistake)

felipecsl:
alright comments addressed and commits squashed

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ActivityTracker.java
rickbrew: @jasta , what do you think ?

rickbrew:
Looks pretty good except there are still 2 commits to be squashed :)

felipecsl:
DONE!

rickbrew:
Awesome :) I will take this for a spin tomorrow and then merge it if it all looks good.

stetho/src/main/java/com/facebook/stetho/screencast/android/ScreencastDispatcher.java
```diff
@@ -0,0 +1,151 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.screencast.android;
```
rickbrew: We need to move this class into a package over in `com.facebook.stetho.inspector` .... something like `com.facebook.stetho.inspector.screencast.ScreencastDispatcher` sounds pretty good

stetho/src/main/java/com/facebook/stetho/screencast/android/ScreencastDispatcher.java
```diff
@@ -0,0 +1,151 @@
...
+
+  private final Handler mMainHandler = new Handler(Looper.getMainLooper());
+  private final BitmapFetchRunnable mBitmapFetchRunnable = new BitmapFetchRunnable();
+  private final ActivityTracker mActivityTracker = ActivityTracker.get();
+  private final EventDispatchRunnable mEventDispatchRunnable = new EventDispatchRunnable();
+
+  private boolean mIsRunning;
+  private Handler mBackgroundHandler;
+  private JsonRpcPeer mPeer;
+  private HandlerThread mHandlerThread;
```
rickbrew: We really should tear down this thread when we stop screencasting

stetho/src/main/java/com/facebook/stetho/screencast/android/ScreencastDispatcher.java
```diff
@@ -0,0 +1,151 @@
...
+        return;
+      }
+      int width = mBitmap.getWidth();
+      int height = mBitmap.getHeight();
+      mStream.reset();
+      Base64OutputStream base64Stream = new Base64OutputStream(mStream, Base64.DEFAULT);
+      // request format is either "jpeg" or "png"
+      Bitmap.CompressFormat format = Bitmap.CompressFormat.valueOf(mRequest.format.toUpperCase());
+      mBitmap.compress(format, mRequest.quality, base64Stream);
+      mEvent.data = mStream.toString();
```
rickbrew: I'm seeing a NullPointerException happening here, and it's easy to reproduce. There's a race between `stopScreencasting()` and this method: `mStream` is non-null when we call its `reset()` method, and then it's null by the time we call its `toString()` method. There are various ways to fix this. One possibility is that this method can load all the values that it needs from the outer class into local variables and then only use those. Another possibility is to implement a better cancellation system, e.g. `stopScreencasting()` would enqueue a `Runnable` to `mHandlerThread` whose job would be to tear everything down (e.g. set fields to null and stuff). That would guarantee that these don't have races.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Page.java
```diff
@@ -147,6 +150,24 @@ public void clearDeviceOrientationOverride(JsonRpcPeer peer, JSONObject params)
   }
 
   @ChromeDevtoolsMethod
+  public void startScreencast(final JsonRpcPeer peer, JSONObject params) {
+    final StartScreencastRequest request = mObjectMapper.convertValue(
+        params, StartScreencastRequest.class);
+    mScreencastDispatcher.startScreencast(peer, request);
```
rickbrew: What if multiple peers want to initiate screencasting?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Page.java
```diff
@@ -9,25 +9,28 @@
...
 import org.json.JSONObject;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+
 public class Page implements ChromeDevtoolsDomain {
   private final Context mContext;
+  private final ScreencastDispatcher mScreencastDispatcher = new ScreencastDispatcher();
```
rickbrew: Related to my comment, 'what if multiple peers want to initiate screencasting?' ...

I'd like to recommend a design change here. First, I don't think we should construct a ScreencastDispatcher until we receive a screencasting request. When screencasting stops, we need to completely dispose of all the resources associated with screencasting (like the `HandlerThread`!), including the setting of mScreencastDispatcher to null. 

If a 2nd peer initiates screencasting, `ScreencastDispatcher` should be able to add that to its broadcast list (e.g. instead of `mPeer.invokeMethod()` we would loop over a thread-safe `List` of peers). This last part in particular might be kind of frustrating at this point, especially since I communicated that everything was looking pretty good and almost ready to merge, and also because I'm not sure that this is needed "in practice." So, I'd support a short-term strategy of simply refusing a 2nd peer's screencasting request, and then filing a new issue to support multi-peer screencasting.

stetho/src/main/java/com/facebook/stetho/screencast/android/ScreencastDispatcher.java
```diff
@@ -0,0 +1,151 @@
...
+import android.util.Base64OutputStream;
+import android.view.View;
+import com.facebook.stetho.common.LogUtil;
+import com.facebook.stetho.inspector.elements.android.ActivityTracker;
+import com.facebook.stetho.inspector.jsonrpc.JsonRpcPeer;
+import com.facebook.stetho.inspector.protocol.module.Page;
+
+import java.io.ByteArrayOutputStream;
+
+public class ScreencastDispatcher {
```
rickbrew: final please :)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ActivityTracker.java
rickbrew: Let's go ahead and change this back to `getActivitiesView`

rickbrew:
Okay unfortunately I found some serious issues in the code after reviewing it more thoroughly, along with a reliable crash when I tried it out.

felipecsl:
no worries, I'm sure it's not 100% solid yet. Lets keep iterating

felipecsl:
ok all issues taken care of (except for the multiple peers). I'll get back to it once I have more time this week.

felipecsl:
Thanks for the thorough review

jasta:
@felipecsl thanks for sticking with this PR BTW!  I'm really excited to have this in and I think you and @rickbrew are on the right track trying to massage this into a state where it will be stable and high quality for users.

felipecsl:
thank you :)

felipecsl:
@rickbrew regarding the multiple peers support, this may sound weird, but I got 2 devtools windows open, both with stetho connected to the same device and it seemed to worked fine. I was able to see the device screen on both devtools windows and they seemed identical in functionality.
Here's a screenshot. Am I missing something?
screen shot 2015-07-30 at 10 23 05 pm

felipecsl:
Wait, are those considered to be the same peer because they are running on the same chrome instance?

rickbrew:
I'm not entirely sure :)

I should be able to look at this on Wednesday

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Page.java
```diff
@@ -147,6 +150,26 @@ public void clearDeviceOrientationOverride(JsonRpcPeer peer, JSONObject params)
   }
 
   @ChromeDevtoolsMethod
+  public void startScreencast(final JsonRpcPeer peer, JSONObject params) {
+    final StartScreencastRequest request = mObjectMapper.convertValue(
+        params, StartScreencastRequest.class);
+    mScreencastDispatcher = new ScreencastDispatcher();
```
rickbrew: We need to check if `mScreencastDispatcher` is non-null and not overwrite it with a new reference.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Page.java
```diff
@@ -147,6 +150,26 @@ public void clearDeviceOrientationOverride(JsonRpcPeer peer, JSONObject params)
...
+    mScreencastDispatcher.startScreencast(peer, request);
+  }
+
+  @ChromeDevtoolsMethod
+  public void stopScreencast(JsonRpcPeer peer, JSONObject params) {
+    mScreencastDispatcher.stopScreencast();
```
rickbrew: don't call stopScreencast() if mScreencastDistpacher is null

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Page.java
```diff
@@ -9,25 +9,28 @@
...
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+
 public class Page implements ChromeDevtoolsDomain {
   private final Context mContext;
+  private final ObjectMapper mObjectMapper = new ObjectMapper();
+  private ScreencastDispatcher mScreencastDispatcher;
```
rickbrew: this should have a @Nullable annotation on it

stetho/src/main/java/com/facebook/stetho/inspector/screencast/ScreencastDispatcher.java
```diff
@@ -0,0 +1,160 @@
...
+import android.view.View;
+import com.facebook.stetho.common.LogUtil;
+import com.facebook.stetho.inspector.elements.android.ActivityTracker;
+import com.facebook.stetho.inspector.jsonrpc.JsonRpcPeer;
+import com.facebook.stetho.inspector.protocol.module.Page;
+
+import java.io.ByteArrayOutputStream;
+
+public final class ScreencastDispatcher {
+  private static final float FRAMERATE = 60f;
```
rickbrew: Calling this a "framerate" is quite a misnomer since we won't actually be able to compress a PNG/JPEG at 60fps. We're not making an attempt to actually stream at 60 frames per second. Rather, this value is being used to calculate an inter-frame delay. And, having an inter-frame delay of floor(1000/60)= 16 milliseconds is very, very aggressive.

I'd prefer just having the FRAME_DELAY constant and setting it to something that hopefully won't cause low-end devices to spend almost all of their available CPU time on screencast. Maybe try 200ms?

stetho/src/main/java/com/facebook/stetho/inspector/screencast/ScreencastDispatcher.java
```diff
@@ -0,0 +1,160 @@
...
+  private static final long FRAME_DELAY = (long) (1000 / FRAMERATE);
+
+  private final Handler mMainHandler = new Handler(Looper.getMainLooper());
+  private final BitmapFetchRunnable mBitmapFetchRunnable = new BitmapFetchRunnable();
+  private final ActivityTracker mActivityTracker = ActivityTracker.get();
+  private final EventDispatchRunnable mEventDispatchRunnable = new EventDispatchRunnable();
+
+  private boolean mIsRunning;
+  private Handler mBackgroundHandler;
```
rickbrew: This is never reassigned to another reference or to null, and can thus be final.

And, for any of the other fields, they should be final if they aren't ever reassigned to something else.

stetho/src/main/java/com/facebook/stetho/inspector/screencast/ScreencastDispatcher.java
```diff
@@ -0,0 +1,160 @@
...
+  private Page.ScreencastFrameEventMetadata mMetadata = new Page.ScreencastFrameEventMetadata();
+
+  public ScreencastDispatcher() {
+  }
+
+  public void startScreencast(JsonRpcPeer peer, Page.StartScreencastRequest request) {
+    mRequest = request;
+
+    LogUtil.d("Starting screencast");
+    if (mHandlerThread == null) {
```
rickbrew: It looks like ScreencastDispatcher supports starting, then stopping, then starting again. This isn't how the class is used in Page -- the object is thrown away after stopScreencast() is called. We therefore don't need to support restarting dispatcher over here, and it'll make the code a lot simpler if we assume we're "trash" once stopScreencasting() is called.

stetho/src/main/java/com/facebook/stetho/inspector/screencast/ScreencastDispatcher.java
felipecsl: sounds good!

stetho/src/main/java/com/facebook/stetho/inspector/screencast/ScreencastDispatcher.java
felipecsl: actually nothing is initialized in the constructor (it's currently empty). Stuff only gets allocated inside `startScreencast` thus this can't be final

felipecsl: Since `startScreencast()` is only ever called right after the constructor, we could potentially move the object allocations to the constructor and just start the runnables in `startScreencast()`. What do you think?

felipecsl:
Since this PR is getting quite big (and old), I'd propose moving the multiple peers support into a separate follow-up PR. I'd be happy to do that after this one is merged (and also other stuff like support for `emulateTouchFromMouseEvent` to allow interacting with the app from DevTools ;))

stetho/src/main/java/com/facebook/stetho/inspector/screencast/ScreencastDispatcher.java
```diff
@@ -0,0 +1,156 @@
...
+
+  public void stopScreencast() {
+    LogUtil.d("Stopping screencast");
+    mBackgroundHandler.post(new CancellationRunnable());
+  }
+
+  private class BitmapFetchRunnable implements Runnable {
+    @Override
+    public void run() {
+      getScreenBitmap();
```
rickbrew: We're calling a 'get' method but then we're not using the return value ... and there are no other calls to this method, so there isn't really a good reason for this choice of verb. This method should probably be called `updateScreenBitmap`, and have its return type changed to void.

rickbrew:
Yeah I'm perfectly happy to defer multi-peer support, so long as we have protection against it in the current PR (which it looks like you added!).

jasta:
@rickbrew I postponed the 1.2.0 release to after I get back from vacation (I set the new milestone to be Aug 31).  I'd love to get this feature in that release, so I'll defer to you to just merge at will.

rickbrew:
This looks good other than my 1 comment above about the name of the `getBitmap` method. I just tested and it looks like things work pretty well! There's 1 issue I found but I think it's Chrome's fault: element highlighting stops working when screencasting is enabled. `DOM.highlightNode` just isn't getting called. We can file an issue to track that separately though. Might be "by design."

auchenberg:
5 months and still counting :)

felipecsl:
sorry I was travelling the last few weeks and got distracted. Will get it finished tonight in time for the Aug 31th milestone :) thanks for the bump, @auchenberg

felipecsl:
Alright, method renamed and commits squashed. #shipit!

jasta:
lgtm, thanks so much!

auchenberg:
:muscle: :clap: :tada: :balloon: 

Bam! Happy to see this landed, after 5 months of work. Congrats!

jasta:
@auchenberg glad to see this get in as well, I think it's a huge new step forward for Stetho's feature set.  In particular I really want to add keyboard and mouse support and maybe even add features similar to @koush's latest Chrome screencasting plugin offering.  Though I understand his solution is paid so it's unlikely we'll be able to integrate directly.

felipecsl:
finally :beers: 
I'd love to get mouse support as well, let's see if I can hack something together again :laughing:

auchenberg:
@jasta Agreed. Mouse/Keyboard/Touch would be the ultimate experience, no need to pick-up the device again :)

You mean this, https://seo-michael.co.uk/mirror-screencast-app-for-android-by-koush/?

For the controlling logic, there's probably some inspiration to take from https://github.com/openstf/stf

Turbo87:
nice work! does anyone know what the timeframe is for releasing this feature?

jasta:
I meant this: http://www.vysor.io/

From the user point of view it works similar to @felipecsl's screencasting feature but it is better optimized by utilizing the same machinery as screenrecord and directly pumps input events to the input manager service (I presume).  It won't work for us for security reasons however, unless we wanted to require an additional daemon be running on your device via a Chrome extension or something.  I don't know that we want to go that far, however.

paulirish:
re:vysor…

The screencasting that's part of http://openstf.io/ might be the best I've seen. I was earlier urging Koush (of vysor) to adopt it.  https://github.com/openstf/minicap

We're planning on using the same core technique for screencasting v2 for chrome, as well.

<hr>

That said, awesome to see screencasting here in stetho! very nice.

koush:
@paulirish Yeah, I dug into minicap. Minicap sends a high frequency jpegs across in a stream. Vysor is using MediaCodec to send an h264 stream. Minicap requires OS and architecture specific native binaries as well.

Latency is higher with Vysor's h264 pipeline (though still sub 100ms), and the host PC requirements are steeper, but that lets me keep the required bitrate under 500kbps and still get acceptable quality on a Nexus 6. This bitrate is tunable. Minicap requires a couple MBps. Varies by device resolution obviously.

Chose this route for Vysor Share specifically for remote testing and remote debugging, ie, non-local usage.

PR #204, opened 2015-06-18
Implementation of DOM tree diffing (fixes ListView, et. al.)

### LINKED ISSUE #198: ListView doesn't work correctly in Elements tab
ListView is causing some problems for us. Using a ViewGroup.OnHierarchyChangedListener is not sufficient. We're not getting events for when child Views are reordered due to scrolling, so they appear to be in the wrong order even though they have the right data. 

At other times we're getting notified of children being added in the wrong order (or something), so when we raise the childInserted event, the previousChild ends up being an element that isn't mapped yet and then we crash:
```
  java.lang.NullPointerException
      at com.facebook.stetho.inspector.protocol.module.DOM$ProviderListener.onChildInserted(DOM.java:324)
      at com.facebook.stetho.inspector.elements.android.AndroidDOMProvider.onChildInserted(AndroidDOMProvider.java:183)
      at com.facebook.stetho.inspector.elements.android.ViewGroupDescriptor$ElementContext.onChildViewAdded(ViewGroupDescriptor.java:245)
      at android.view.ViewGroup.onViewAdded(ViewGroup.java:3685)
      at android.view.ViewGroup.addViewInner(ViewGroup.java:3821)
      at android.view.ViewGroup.addViewInLayout(ViewGroup.java:3747)
      at android.widget.ListView.setupChild(ListView.java:1853)
      at android.widget.ListView.makeAndAddView(ListView.java:1806)
      at android.widget.ListView.fillDown(ListView.java:697)
      at android.widget.ListView.fillGap(ListView.java:661)
```

rickbrew:
The original problem is that ListView updates just didn't work: the ordering was all wrong after scrolling around because we didn't get any updates from ViewGroup.OnHierarchyChangedListener. 

We've known for awhile that we need to have some kind of support for DOM tree diffing because every location in the Android UI tree seems to require a different protocol for subscribing to change notifications (if it's even possible), some of which have require brittle, reflection-based hacks (yuck). 

DOM tree diffing necessitates the use of a "shadow DOM" which contains a copy of the DOM so can keep track of what we've sent to Chrome and compare it to what the DOMProvider says the tree should look like now.

DOM tree diffing is split into 3 phases:
* Collecting changes to the shadow DOM. We currently brute-force this in two ways:
  1. AndroidDOMProvider does not currently have the ability to detect when (or where!) changes are made to the Android UI tree. Instead, it has a timer that fires every second and it just announces to its DOMProvider.Listener that it "changed."
  2. Since DOMProvider.Listener.onChanged() takes no parameters, this means that DOM's only option is to read the entire tree from mDOMProvider. Later on we could make it so onChange() takes more information about what changed.
* Sending the changes over to Chrome. This is accomplished in DOM.applyShadowDOMUpdate() with the participation of a few other friends.
* Merging the changes back into the shadow DOM. This is implemented in ShadowDOM.Update.commit().

While the first phase is brute-forced and takes linear time with respect to the number of elements in the tree, the other two phases are not brute-forced. Their time and space requirements should be roughly proportional to the number of elements that have changed. Disconnected elements are included in that equation: if 1 element is removed but it has a sub-tree of 50 elements, then all 51 of those elements must be considered.

Closes #198

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/common/ListUtil.java
+++ b/stetho/src/main/java/com/facebook/stetho/common/ListUtil.java
@@ -1,0 +1,249 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.common;
+
+import java.util.AbstractList;
+import java.util.Collections;
+import java.util.List;
+import java.util.RandomAccess;
+
+public final class ListUtil {
+  private ListUtil() {
+  }
+
+  /**
+   * Compares the contents of two {@link List}s by using object identity.
+   */
+  public static <T> boolean identityEquals(List<? extends T> list1, List<? extends T> list2) {
+    if (list1 == list2) {
+      return true;
+    }
+    int size = list1.size();
+    if (size != list2.size()) {
+      return false;
+    }
+    for (int i = 0; i < size; ++i) {
+      if (list1.get(i) != list2.get(i)) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  /**
+   * Copies the given {@link List} and returns the copy as an immutable {@link List}.
+   */
+  public static <T> List<T> copyToImmutableList(List<T> list) {
+    if (list instanceof ImmutableList) {
+      return list;
+    }
+    int size = list.size();
+    switch (size) {
+      case 0:
+        return Collections.emptyList();
+      case 1:
+        return new OneItemImmutableList<>(list.get(0));
+      case 2:
+        return new TwoItemImmutableList<>(list.get(0), list.get(1));
+      case 3:
+        return new ThreeItemImmutableList<>(list.get(0), list.get(1), list.get(2));
+      case 4:
+        return new FourItemImmutableList<>(list.get(0), list.get(1), list.get(2), list.get(3));
+      case 5:
+        return new FiveItemImmutableList<>(
+            list.get(0), list.get(1), list.get(2), list.get(3), list.get(4));
+      default:
+        Object[] array = list.toArray();
+        return new ImmutableArrayList<>(array);
+    }
+  }
+
+  private static interface ImmutableList<E> extends List<E>, RandomAccess {
+  }
+
+  private static final class ImmutableArrayList<E>
+      extends AbstractList<E> implements ImmutableList<E> {
+    private final Object[] mArray;
+
+    public ImmutableArrayList(Object[] array) {
+      mArray = array;
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public E get(int location) {
+      return (E) mArray[location];
+    }
+
+    @Override
+    public int size() {
+      return mArray.length;
+    }
+  }
+
+  private static final class OneItemImmutableList<E>
+      extends AbstractList<E> implements ImmutableList<E> {
+    private final E mItem;
+
+    public OneItemImmutableList(E item) {
+      mItem = item;
+    }
+
+    @Override
+    public E get(int location) {
+      if (location == 0) {
+        return mItem;
+      } else {
+        throw new IndexOutOfBoundsException();
+      }
+    }
+
+    @Override
+    public int size() {
+      return 1;
+    }
+  }
+
+  private static final class TwoItemImmutableList<E>
+      extends AbstractList<E> implements ImmutableList<E> {
+    private final E mItem0;
+    private final E mItem1;
+
+    public TwoItemImmutableList(E item0, E item1) {
+      mItem0 = item0;
+      mItem1 = item1;
+    }
+
+    @Override
+    public E get(int location) {
+      switch (location) {
+        case 0:
+          return mItem0;
+        case 1:
+          return mItem1;
+        default:
+          throw new IndexOutOfBoundsException();
+      }
+    }
+
+    @Override
+    public int size() {
+      return 2;
+    }
+  }
+
+  private static final class ThreeItemImmutableList<E>
+      extends AbstractList<E> implements ImmutableList<E> {
+    private final E mItem0;
+    private final E mItem1;
+    private final E mItem2;
+
+    public ThreeItemImmutableList(E item0, E item1, E item2) {
+      mItem0 = item0;
+      mItem1 = item1;
+      mItem2 = item2;
+    }
+
+    @Override
+    public E get(int location) {
+      switch (location) {
+        case 0:
+          return mItem0;
+        case 1:
+          return mItem1;
+        case 2:
+          return mItem2;
+        default:
+          throw new IndexOutOfBoundsException();
+      }
+    }
+
+    @Override
+    public int size() {
+      return 3;
+    }
+  }
+
+  private static final class FourItemImmutableList<E>
+      extends AbstractList<E> implements ImmutableList<E> {
+    private final E mItem0;
+    private final E mItem1;
+    private final E mItem2;
+    private final E mItem3;
+
+    public FourItemImmutableList(E item0, E item1, E item2, E item3) {
+      mItem0 = item0;
+      mItem1 = item1;
+      mItem2 = item2;
+      mItem3 = item3;
+    }
+
+    @Override
+    public E get(int location) {
+      switch (location) {
+        case 0:
+          return mItem0;
+        case 1:
+          return mItem1;
+        case 2:
+          return mItem2;
+        case 3:
+          return mItem3;
+        default:
+          throw new IndexOutOfBoundsException();
+      }
+    }
+
+    @Override
+    public int size() {
+      return 4;
+    }
+  }
+
+  private static final class FiveItemImmutableList<E>
+      extends AbstractList<E> implements ImmutableList<E> {
+    private final E mItem0;
+    private final E mItem1;
+    private final E mItem2;
+    private final E mItem3;
+    private final E mItem4;
+
+    public FiveItemImmutableList(E item0, E item1, E item2, E item3, E item4) {
+      mItem0 = item0;
+      mItem1 = item1;
+      mItem2 = item2;
+      mItem3 = item3;
+      mItem4 = item4;
+    }
+
+    @Override
+    public E get(int location) {
+      switch (location) {
+        case 0:
+          return mItem0;
+        case 1:
+          return mItem1;
+        case 2:
+          return mItem2;
+        case 3:
+          return mItem3;
+        case 4:
+          return mItem4;
+        default:
+          throw new IndexOutOfBoundsException();
+      }
+    }
+
+    @Override
+    public int size() {
+      return 5;
+    }
+  }
+}
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/ElementInfo.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/ElementInfo.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.elements;
+
+import com.facebook.stetho.common.ListUtil;
+import com.facebook.stetho.common.Util;
+
+import javax.annotation.concurrent.Immutable;
+
+import java.util.List;
+
+@Immutable
+public final class ElementInfo {
+  public final Object element;
+  public final Object parentElement;
+  public final List<Object> children;
+
+  public ElementInfo(
+      Object element,
+      Object parentElement,
+      List<Object> children) {
+    this.element = Util.throwIfNull(element);
+    this.parentElement = parentElement;
+    this.children = ListUtil.copyToImmutableList(children);
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (o == this) {
+      return true;
+    }
+
+    if (o instanceof ElementInfo) {
+      ElementInfo other = (ElementInfo) o;
+      return this.element == other.element
+          && this.parentElement == other.parentElement
+          && ListUtil.identityEquals(this.children, other.children);
+    }
+
+    return false;
+  }
+}
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/ShadowDOM.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/ShadowDOM.java
@@ -1,0 +1,372 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.elements;
+
+import android.app.Activity;
+import com.facebook.stetho.common.Accumulator;
+import com.facebook.stetho.common.ListUtil;
+import com.facebook.stetho.common.Util;
+
+import java.util.ArrayDeque;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.IdentityHashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Queue;
+import java.util.Set;
+
+public final class ShadowDOM implements DOMView {
+  private final Object mRootElement;
+  private final IdentityHashMap<Object, ElementInfo> mElementToInfoMap = new IdentityHashMap<>();
+  private boolean mIsUpdating;
+
+  public ShadowDOM(Object rootElement) {
+    mRootElement = Util.throwIfNull(rootElement);
+  }
+
+  @Override
+  public Object getRootElement() {
+    return mRootElement;
+  }
+
+  @Override
+  public ElementInfo getElementInfo(Object element) {
+    return mElementToInfoMap.get(element);
+  }
+
+  public UpdateBuilder beginUpdate() {
+    if (mIsUpdating) {
+      throw new IllegalStateException();
+    }
+
+    mIsUpdating = true;
+
+    return new UpdateBuilder();
+  }
+
+  public final class UpdateBuilder {
+    /**
+     * We use a {@link LinkedHashMap} to preserve ordering between
+     * {@link UpdateBuilder#setElementChildren(Object, List)} and
+     * {@link Update#getChangedElements(Accumulator)}. This isn't needed for correctness but it
+     * significantly improves performance.<p/>
+     *
+     * Transmitting DOM updates to Chrome works best if we can do it in top-down order because it
+     * allows us to skip processing (and, more importantly, transmission) of an element that was
+     * already transmitted in a previous DOM.childNodeInserted event (i.o.w. we can skip
+     * transmission of E2 if it was already bundled up in E1's event, where E2 is any element in
+     * E1's sub-tree). DOM.childNodeInserted transmits the node being inserted by-value, so it takes
+     * time and space proportional to the size of that node's sub-tree. This means the difference
+     * between O(n^2) and O(n) time for transmitting updates to Chrome.<p/>
+     *
+     * We currently only have one implementation of {@link DOMProvider},
+     * {@link com.facebook.stetho.inspector.elements.android.AndroidDOMProvider}, and it already
+     * supplies element changes in top-down order. Because of this, we can just use
+     * {@link LinkedHashMap} instead of adding some kind of post-process sorting of the elements to
+     * put them in that order. If we reach a point where we can't or shouldn't rely on elements
+     * being forwarded to us in top-down order, then we should change this field to an
+     * {@link IdentityHashMap} and sort them before relaying them via
+     * {@link Update#getChangedElements(Accumulator)}.<p/>
+     *
+     * When a large sub-tree is added (e.g. starting a new {@link Activity}), the use of
+     * {@link LinkedHashMap} instead of {@link IdentityHashMap} can mean the difference between an
+     * update taking 500ms versus taking more than 30 seconds.<p/>
+     *
+     * Technically we actually want something like a LinkedIdentityHashMap because we do want
+     * to key off of object identity instead of allowing for the possibility of value identity.
+     * Given the difference in performance, however, the risk of potential protocol abuse seems
+     * reasonable.<p/>
+     */
+    private final Map<Object, ElementInfo> mElementToInfoChangesMap = new LinkedHashMap<>();
+
+    /**
+     * This contains every element in {@link #mElementToInfoChangesMap} whose
+     * {@link ElementInfo#parentElement} is null. {@link ShadowDOM} provides access to a tree, which
+     * means it has a single root (only one element with a null parent). During an update, however,
+     * the DOM can be conceptually thought of as being a forest. The true root is identified by
+     * {@link #mRootElement}, and all other roots identify disconnected trees full of elements that
+     * must be garbage collected.
+     */
+    private final HashSet<Object> mRootElementChanges = new HashSet<>();
+
+    /**
+     * This is used during {@link #setElementChildren}. We allocate 1 on-demand and reuse it.
+     */
+    private HashSet<Object> mCachedNotNewChildrenSet;
+
+    public void setElementChildren(Object element, List<Object> children) {
+      // If we receive redundant information, then nothing needs to be done.
+      ElementInfo changesElementInfo = mElementToInfoChangesMap.get(element);
+      if (changesElementInfo != null &&
+          ListUtil.identityEquals(children, changesElementInfo.children)) {
+        return;
+      }
+
+      ElementInfo oldElementInfo = mElementToInfoMap.get(element);
+      if (changesElementInfo == null &&
+          oldElementInfo != null &&
+          ListUtil.identityEquals(children, oldElementInfo.children)) {
+        return;
+      }
+
+      ElementInfo newElementInfo;
+      if (changesElementInfo != null &&
+          oldElementInfo != null &&
+          oldElementInfo.parentElement == changesElementInfo.parentElement &&
+          ListUtil.identityEquals(children, oldElementInfo.children)) {
+        // setElementChildren() was already called for element with changes during this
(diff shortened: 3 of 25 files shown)


stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -172,64 +212,245 @@ public void run() {
...
-    NodeDescriptor descriptor = mDOMProvider.getNodeDescriptor(element);
-
-    Node node = new Node();
-    node.nodeId = mObjectIdMapper.putObject(element);
-    node.nodeType = descriptor.getNodeType(element);
-    node.nodeName = descriptor.getNodeName(element);
-    node.localName = descriptor.getLocalName(element);
-    node.nodeValue = descriptor.getNodeValue(element);
+  private void updateTree() {
+    //
```
jasta: what's the deal with the empty comments?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -172,64 +212,245 @@ public void run() {
...
-    return childNodes;
+    releaseChildrenAccumulator(childrenAccumulator);
+
+    return updateBuilder.build();
   }
 
-  private void removeElementTree(Object element) {
-    mDOMProvider.verifyThreadAccess();
+  private void applyDOMUpdate(final ShadowDOM.Update domUpdate) {
+    // TODO: it'd be nice if we could delegate our calls into mPeerManager.sendNotificationToPeers()
```
jasta: Why can't we?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -172,64 +212,245 @@ public void run() {
...
+      }
+    });
+
+    domUpdate.commit();
+
+    VerifyShadowDOMMatchesDOMProvider();
+  }
+
+  // TODO: organization of test code
+  private void VerifyShadowDOMMatchesDOMProvider() {
```
jasta: nit: verify (lowercase v)

Also, what does the TODO mean?

stetho/src/main/java/com/facebook/stetho/common/ListUtil.java
```diff
@@ -0,0 +1,248 @@
...
+        return false;
+      }
+    }
+    return true;
+  }
+
+  /**
+   * Copies the given {@link List} and returns the copy as an immutable {@link List}.
+   */
+  public static <T> List<T> copyToImmutableList(List<T> list) {
```
jasta: Strongly prefer not adding this code.  Immutable lists is a broken, half-assed concept in Java and should be rethought outside the List interface.  And it should be rethought somewhere else that isn't Stetho :)

jasta: ...another thing to note: Stetho is supposed to be _very small_.  Immutable lists is a luxury we don't need...

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewGroupDescriptor.java
```diff
@@ -111,11 +87,13 @@ public void markDecorView() {
...
 
-    public int getChildCount() {
-      if (mIsDecorView) {
-        return getDecorViewChildCount();
-      } else {
-        return mElement.getChildCount();
+    public void getChildren(Accumulator<Object> children) {
+      for (int i = 0, N = mElement.getChildCount(); i < N; ++i) {
+        final View child = mElement.getChildAt(i);
+        if (isChildVisible(child)) {
```
jasta: Why does it matter if it's visible?

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDOMProvider.java
```diff
@@ -37,6 +38,7 @@
 import java.util.List;
 
 final class AndroidDOMProvider implements DOMProvider, AndroidDescriptorHost {
+  private static final long REPORT_CHANGED_INTERVAL_MS = 1000;
```
jasta: Javadoc here :)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDOMProvider.java
```diff
@@ -48,6 +50,24 @@
   private final InspectModeHandler mInspectModeHandler;
   private Listener mListener;
 
+  // We don't yet have an an implementation for reliably detecting fine-grained changes in the
+  // View tree. So, for now at least, we have a timer that runs every so often and just reports
+  // that we changed. Our listener will then read the entire DOM from us and transmit the changes to
+  // Chrome. Detecting, reporting, and traversing fine-grained changes is a future work item.
```
jasta: actually these docs belong on the 1000 constant.  The constants stand out more IMO and are better places to document unexpected behaviours like this

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDOMProvider.java
```diff
@@ -48,6 +50,24 @@
...
+  private boolean mIsReportChangesTimerPosted = false;
+  private final Runnable mReportChangesTimer = new Runnable() {
+    @Override
+    public void run() {
+      mIsReportChangesTimerPosted = false;
+
+      if (mListener != null) {
+        mListener.onChanged();
```
jasta: This is what computes the diff?  The semantics read very strangely to me, as if this code has already detected a change and is telling something about it.

Reworking the type/variable name of `mListener` and possibly modifying `onChanged` to be `onPossibleChange` might clear this up

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -292,6 +685,11 @@ protected void onUnmapped(Object object, int id) {
 
   private final class ProviderListener implements DOMProvider.Listener {
     @Override
+    public void onChanged() {
+      updateTree();
```
jasta: Wrap this with SystemClock timings and print the result using `LogUtil.d`.  It'd be useful for us to track this easily and also highlight to interested developers how this works.  Might also keep us honest about improving it :)

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -172,64 +212,245 @@ public void run() {
...
+    Queue<Object> queue = new ArrayDeque<>();
+    queue.add(mDOMProvider.getRootElement());
+
+    while (!queue.isEmpty()) {
+      final Object element = queue.remove();
+      NodeDescriptor descriptor = mDOMProvider.getNodeDescriptor(element);
+      mObjectIdMapper.putObject(element);
+      descriptor.getChildren(element, childrenAccumulator);
+      updateBuilder.setElementChildren(element, childrenAccumulator);
+      for (int i = 0, N = childrenAccumulator.size(); i < N; ++i) {
```
jasta: that's a neat pattern I've never seen before :)

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -172,64 +212,245 @@ public void run() {
...
+        queue.add(childrenAccumulator.get(i));
       }
+      childrenAccumulator.clear();
```
jasta: Seems weird that you don't just create a new accumulator that delegates to the queue then pass that in?

stetho/src/main/java/com/facebook/stetho/inspector/elements/ShadowDOM.java
```diff
@@ -0,0 +1,364 @@
...
+    if (mIsUpdating) {
+      throw new IllegalStateException();
+    }
+
+    mIsUpdating = true;
+
+    return new UpdateBuilderImpl();
+  }
+
+  public static interface UpdateBuilder {
```
jasta: Why hide the impl?  Can't you just use private to hide the parts of the impl you don't want folks to see?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -172,64 +212,245 @@ public void run() {
...
+
+  // TODO: organization of test code
+  private void VerifyShadowDOMMatchesDOMProvider() {
+    final List<Object> shadowDOMList = new ArrayList<>();
+    Queue<Object> shadowDOMQueue = new ArrayDeque<>();
+    shadowDOMQueue.add(mShadowDOM.getRootElement());
+    DOMViewUtil.TraverseBFS(
+        mShadowDOM,
+        shadowDOMQueue,
+        new Function<ElementInfo, TreeTraverseAction>() {
```
jasta: It's a bit strange that you generalized this all the way to `Function` but only have two concrete instances of it, both defined here, and both doing the same thing.  This generalization seems wasted...

stetho/src/main/java/com/facebook/stetho/common/ListAccumulator.java
```diff
@@ -0,0 +1,15 @@
...
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.common;
+
+import java.util.List;
+
+public interface ListAccumulator<E> extends List<E>, Accumulator<E> {
```
jasta: ...why not just go straight after the `ArrayListAccumulator`?  This diff appears to be way overrotating on generalization that isn't even used.  This generalization has a real cost to our released jar size which folks are sensitive to.

jasta:
The algorithm looks solid to me though I'll need to spend a bit more time reading through the details to make sure I actually understand it :)

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: That's the part that I need to extract into test code.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: I don't know :) I needed to ask you if there's a preferred or existing way of doing that

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: the for loop?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: Oh, I was just marking code I still wanted to delete / refactor. Like I mentioned before, there's still a small amount of "todo" left in this PR.

stetho/src/main/java/com/facebook/stetho/common/ListUtil.java
rickbrew: I mostly did this to save an array allocation on every List on every ElementInfo via Empty...One...FiveElementImmutableList (going up to 5, from inspection of the UI tree in another app, seemed to cover the vast majority of cases). It would also be possible to just store an array in ElementInfo, but that makes it more clumsy to compare and transfer the arrays around while maintaining immutability and encapsulation so I hadn't done it yet. We'll see what I can come up with ...

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewGroupDescriptor.java
rickbrew: This can be answered by looking at isChildVisible() :) isChildVisible()'s job is mostly to hide DOMHiddenViews (e.g. highlighting stuff). This code isn't semantically changed from before but because of the change in protocol (getChildCount/At vs. getChildren), it's much simpler.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDOMProvider.java
rickbrew: No this doesn't compute any diffs, it's telling DOM when it should do that (or is that what you mean? this is what kicks off the process, in other words). I don't really like the name onPossiblyChanged(), but another verb other than "changed" may be useful here. The intent of the interface is that the DOMProvider should be "smart" about raising the onChanged() event, and it's kind of an implementation detail of AndroidDOMProvider that it isn't smart enough to suppress it. 

My original thinking on this protocol lined it up with the "invalidate" pattern seen elsewhere, e.g. `InvalidateRect` https://msdn.microsoft.com/en-us/library/windows/desktop/dd145002(v=vs.85).aspx . "Tell me when and where you need to be repainted. I'll debounce it and be smart about how I prioritize calling your paint code, balancing quick updates to the user vs. bogging down the system with it. It's okay if your new output is bit-identical to our old output, but it's also to everyone's advantage if you're smart and only call this method when it's actually necessary."

I will scratch my chin for a bit on this...

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: `DOM.updateTree()` does that :)

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: I don't understand what you mean by this. Example code plz?

stetho/src/main/java/com/facebook/stetho/inspector/elements/ShadowDOM.java
rickbrew: Well, the interface defines the protocol that I want the client to use. But I wanted the flexibility of choosing how I implemented it, such as by inlining both interfaces into 1 class (UpdaterBuilderImpl.build() would then just "return this"), or even inlining them into the parent class.

stetho/src/main/java/com/facebook/stetho/common/ListAccumulator.java
rickbrew: Well, for one, because standard convention is to do stuff like `List<T> list = new ArrayList<>()` ? I was also experimenting with other ways to wrap both List and Accumulator, but I just happened to only end up utilizing ArrayListAccumulator.

I wouldn't say the diff is "way overrotating on generalization" ... I'm trying to strike a balance here. Sometimes having some generalization might cost 1 allocation and add 1 Class<?>, but save a ton of byte code elsewhere by removing the need for repeated code, which also has the benefit of improving maintainability.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: Where else is this type of casing repeated? I've also been hacking away at some C# code in my spare time, and the conventions are different there, so I may not have noticed it.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: Well, at this stage it's only being used here. But it was used a lot more while I was developing this. Most instances just got pruned away at the distillation phase (e.g. ShadowDOM.UpdateImpl.getGarbageElements() couldn't use it because it has *slightly* different needs). If I don't end up using it outside of test code then I'll toss it, but it has proven to be a solid generalization.

stetho/src/main/java/com/facebook/stetho/common/ListUtil.java
rickbrew: ... and after thinking on this a bit more, I still think this is a reasonable approach. It's small and to the point, and enables a lot of simplicity over in ElementInfo and ShadowDOM while also saving 1 allocation on almost every ElementInfo instance. The big problem I've seen with ImmutableList elsewhere (at least on Android) is how it's so casually used and ends up encourage lots of unnecessary copying (e.g. allocating an array, passing it to of(T...) which then copies the array even though just wrapping the original array would've worked fine because it isn't used thereafter). I'm not using this casually; I'm being very intentional.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDOMProvider.java
rickbrew: I'm not really sure why we'd need a javadoc for this. I do like the idea of moving this constant (and the comment) over so it's grouped with the associated boolean and the Runnable.

rickbrew:
Okay there's another round of changes. Deleted a bunch of code, added some significant comments explaining some important stuff, ...

stetho/src/main/java/com/facebook/stetho/common/ListUtil.java
jasta: Ack'd, though the problems with immutable list are more subtle than just encouraging copies.  One of the biggest issues is that they interoperate with List which is itself mutable.  Iterable, even, is mutable.

FWIW, I'd much rather have a strongly enforced annotation that prevents mutation through linting and runtime analysis but does nothing at runtime.  But Stetho isn't the place to solve this problem and that's why I prefer that Stetho just doesn't concern itself with immutability...

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewGroupDescriptor.java
jasta: Ack'd

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDOMProvider.java
jasta: Yes, I meant does this functionally cause the diffing.  I agree it's an implementation detail, but I disagree that it's _exclusively_ an implementation detail.  `onPossiblyChanged` suggests that callers will not be perfectly intelligent, and they never will be because if they could be so intelligent we wouldn't have used this design in the first place.  The idea that we have a master trigger that will rescan and diff the entire DOM suggests we are not smart about changes.  That's part of the API as much as it is part of the implementation detail.  Might as well make that clearer through the API.

Invalidate semantics also work, but then you should actually call it invalidate.  Saying "onChange" tells any reasonable reader that a change has been detected.  The containing code, however, clearly suggests otherwise :)

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
jasta: Ack'd, didn't notice that, sorry.  DOM.updateTree is fine.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
jasta: yeah, specifically the N initialization inside the initial statement.

stetho/src/main/java/com/facebook/stetho/inspector/elements/ShadowDOM.java
jasta: Strongly prefer treating IntelliJ as that flexibility (it can very trivially divorce concrete impls from their callers and create the appropriate interfaces).  It's much more powerful than you're assuming, I think...

Also, classes have real performance implications on Android and should always be avoided when developer tooling can be used to solve the problem.

stetho/src/main/java/com/facebook/stetho/common/ListAccumulator.java
jasta: Standard Java convention perhaps, but not necessarily Android convention.  One of the main motivators for this is the glaring performance implications of LinkedList's (and let's be honest, there are only two choices here *grin*).  LinkedList's, as they're implemented in the VM, are so incredibly memory inefficient that you should rule out their usage on mobile in all conceivable workloads.  In this way, you can assume all List instances in use are in fact ArrayList.  If the assumption fails, you should convert the code :)

My argument is fundamentally based on this fact: `ListAccumulator` can have only one reasonable implementation.

stetho/src/main/java/com/facebook/stetho/inspector/elements/ElementInfo.java
```diff
@@ -0,0 +1,46 @@
...
+  public final Object parentElement;
+  public final List<Object> children;
+
+  public ElementInfo(
+      Object element,
+      Object parentElement,
+      List<Object> children) {
+    this.element = Util.throwIfNull(element);
+    this.parentElement = parentElement;
+    this.children = ListUtil.copyToImmutableList(children);
```
jasta: The grating thing for me here is that I just can't tell if this is a copy or not.  Or if the copy is conditional (sometimes it's a copy but usually it's not?)  I find it considerably more clear to just enforce these contracts through design and documentation when they are _INTERNAL_ to your own module.

I'm fine to let it land as-is, but I can tell you from experience this is not a pattern that mobile wants.

stetho/src/main/java/com/facebook/stetho/inspector/elements/ShadowDOM.java
```diff
@@ -0,0 +1,397 @@
...
+    public void commit();
+  }
+
+  private final class UpdateBuilderImpl implements UpdateBuilder {
+    /**
+     * We use a {@link LinkedHashMap} to preserve ordering between
+     * {@link UpdateBuilder#setElementChildren(Object, List)} and
+     * {@link Update#getChangedElements(Accumulator)}. This isn't needed for correctness but it
+     * significantly improves performance.
+     *
```
jasta: `<p/>` -- check your javadocs with IntelliJ's View -> Quick Documentation feature.  They will format like hell as written :)

stetho/src/main/java/com/facebook/stetho/inspector/elements/ShadowDOM.java
```diff
@@ -0,0 +1,397 @@
...
+     * Transmitting DOM updates to Chrome works best if we can do it in BFS order because it allows
+     * us to skip processing (and, more importantly, transmission) of an element that was already
+     * transmitted in a previous DOM.childNodeInserted event (i.o.w. we can skip transmission of E2
+     * if it was already bundled up in E1's event, where E2 is any element in E1's sub-tree).
+     * DOM.childNodeInserted transmits the node being inserted by-value, so it takes time and space
+     * proportional to the size of that node's sub-tree. This means the difference between O(n^2)
+     * and O(n) time for transmitting updates to Chrome.
+     *
+     * Since {@link com.facebook.stetho.inspector.elements.android.AndroidDOMProvider} already
```
jasta: Tell IntelliJ to collapse this.  I haven't yet modified Stetho's proejct settings to do it for you :)

stetho/src/main/java/com/facebook/stetho/inspector/helper/ChromePeerManager.java
```diff
@@ -94,9 +102,12 @@ public synchronized boolean hasRegisteredPeers() {
     return !mReceivingPeers.isEmpty();
   }
 
-  private synchronized ArrayList<JsonRpcPeer> getReceivingPeersCopy() {
-    // This performance should be OK in practice because we rarely have more than 1 peer connected.
-    return new ArrayList<JsonRpcPeer>(mReceivingPeers.keySet());
+  private synchronized JsonRpcPeer[] getReceivingPeersSnapshot() {
+    if (mReceivingPeersSnapshot == null) {
```
jasta: Oops, accidentally commented on your commit before.  Moving here: prefer this in a separate commit (can be the same PR tho).  It seems unrelated generally.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -172,64 +208,196 @@ public void run() {
...
+    ShadowDOM.Update domUpdate = createShadowDOMUpdate();
+    boolean isEmpty = domUpdate.isEmpty();
+    if (isEmpty) {
+      domUpdate.abandon();
+    } else {
+      applyDOMUpdate(domUpdate);
+    }
+
+    long endTimeMs = SystemClock.elapsedRealtime();
+    long deltaMs = endTimeMs - startTimeMs;
```
jasta: nit: you can simplify this pattern with:

```java
long deltaMs = SystemClock.elapseRealtime() - startTimeMs;
```

It's quite rare that you actually care about the end time.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -172,64 +208,196 @@ public void run() {
...
+    LogUtil.e(String.format(
+        "DOM.updateTree() completed in %s ms%s",
+        Long.toString(deltaMs),
```
jasta: A few comments here:

1. Why use `Long.toString` when you can use use `%d`?  Java doesn't have any of that `%ld` stuff :)
2. `LogUtil.e` already supports formatting, so the extra `String.format` is unnecessary.
3. `LogUtil.e` is _way_ too verbose :)
4. Technically if you use `String.format` in debug code you should always write `String.format(Locale.US, ...`.  It's best not to assume you know how localization works even in this case it is safe.

jasta:
I'm fine to punt on the more philosophical issues (adding classes and the like), but please do take a pass through the feedback.  Otherwise this looks good to me and I see no major reason we can't land it.  Just remember to squash and merge when you're ready.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: oh ... I thought I got that from you

stetho/src/main/java/com/facebook/stetho/inspector/elements/ShadowDOM.java
rickbrew: Yeah having the interface and class separated just really helped me keep things organized while I was writing it. It isn't technically necessary. If I needed to toss over a different implementation depending on context, then interface would be great. I don't need that now, and if I do later then I can just change it.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: "Why" = using Long.toString() avoids one temporary boxing allocation.

stetho/src/main/java/com/facebook/stetho/inspector/elements/ShadowDOM.java
rickbrew: This didn't actually work because `AndroidDOMProvider` is package-private

PR #206, opened 2015-06-20
Fix memory leak in ViewGroupDescriptor

### LINKED ISSUE #203: Memory leak in ViewGroupDescriptor.ElementContext
We keep track of a mapping from View to either the View itself or the Fragment that owns it.

However, we never remove items from the map ... which means we're leaking Views and Fragments until the ViewGroup itself is unhooked.

### LINKED ISSUE #205: ViewGroupDescriptor.registerDecorView() should move to AndroidDescriptorHost
If you look at where this method is called, in WindowDescriptor, we have some uncomfortable coupling between WindowDescriptor and ViewGroupDescriptor:

```
  private void registerDecorView(View decorView) {
    if (decorView instanceof ViewGroup) {
      Descriptor descriptor = getHost().getDescriptor(decorView);
      if (descriptor instanceof ViewGroupDescriptor) {
        ((ViewGroupDescriptor) descriptor).registerDecorView((ViewGroup) decorView);
      }
    }
  }
```

This code is also coupled to the fact that decorView doesn't have a more specific Descriptor registered for it.

I believe a better way to do this would be something like:

```
  private void registerDecorView(View decorView) {
    if (getHost() is AndroidDescriptorHost) {
      ((AndroidDescriptorHost) getHost()).registerDecorView();
    }
  }
```

rickbrew:
This switches ViewGroupDescriptor to using a WeakHashMap for the View -> [View or Element] cache.

I also moved this cache out of ViewGroupDescriptor.ElementContext and into ViewGroupDescriptor because we don't really need to cache this per-ViewGroup. Why not save a few allocations per ViewGroup?

Closes #203

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewGroupDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewGroupDescriptor.java
@@ -20,10 +20,20 @@ import com.facebook.stetho.inspector.elements.ChainedDescriptor;
 import java.util.Collections;
 import java.util.IdentityHashMap;
 import java.util.Map;
+import java.util.WeakHashMap;
 
 final class ViewGroupDescriptor extends ChainedDescriptor<ViewGroup> {
   private final Map<ViewGroup, ElementContext> mElementToContextMap =
       Collections.synchronizedMap(new IdentityHashMap<ViewGroup, ElementContext>());
+
+  /**
+   * This is a cache that maps from a View to the Fragment that contains it. If the View isn't
+   * contained by a Fragment, then this maps the View to itself. For Views contained by Fragments,
+   * we emit the Fragment instead, and then let the Fragment's descriptor emit the View as its sole
+   * child. This allows us to see Fragments in the inspector as part of the UI tree.
+   */
+  private final Map<View, Object> mViewToElementMap =
+      Collections.synchronizedMap(new WeakHashMap<View, Object>());
 
   public ViewGroupDescriptor() {
   }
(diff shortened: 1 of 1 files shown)


stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewGroupDescriptor.java
```diff
@@ -20,11 +20,20 @@
 import java.util.Collections;
 import java.util.IdentityHashMap;
 import java.util.Map;
+import java.util.WeakHashMap;
 
 final class ViewGroupDescriptor extends ChainedDescriptor<ViewGroup> {
   private final Map<ViewGroup, ElementContext> mElementToContextMap =
       Collections.synchronizedMap(new IdentityHashMap<ViewGroup, ElementContext>());
 
+  // This is a cache that maps from a View to the Fragment that contains it. If the
```
jasta: nit: javadoc

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewGroupDescriptor.java
```diff
@@ -66,15 +75,30 @@ protected Object onGetChildAt(ViewGroup element, int index) {
...
 
-  private final class ElementContext implements ViewGroup.OnHierarchyChangeListener {
-    // This is a cache that maps from a View to the Fragment that contains it. If the
-    // View isn't contained by a Fragment, then this maps the View to itself.
-    // For Views contained by Fragments, we emit the Fragment instead, and then let
-    // the Fragment's descriptor emit the View as its sole child. This allows us to
-    // see Fragments in the inspector as part of the UI tree.
-    private final Map<View, Object> mViewToElementMap =
-        Collections.synchronizedMap(new IdentityHashMap<View, Object>());
+  private Object getElementForView(ViewGroup parentView, View view) {
```
jasta: `@Nullable` but actually, why allow null here?

jasta:
lgtm, merge at will

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewGroupDescriptor.java
rickbrew: Hmm it made sense at one point. I'll just remove the null check. I know why I did this but you'd probably disagree :) It doesn't match up with the usage anymore (a fact which isn't affected by this change btw).

PR #209, opened 2015-06-25
Improve ViewGroupDescriptor by removing unnecessary stuff

### LINKED ISSUE #205: ViewGroupDescriptor.registerDecorView() should move to AndroidDescriptorHost

rickbrew:
Marking a `ViewGroup` as a `decorView` was a necessary optimization due to the old `getChildCount` + `getChildAt` protocol. Now that we've transitioned to `getChildren(Accumulator)`, we don't need to do linear scans in order to answer those questions. Technically, the linear scan adds up to a quadratic scan for decor views, but they only have 1 child so that didn't matter. It did make the code a bit nasty.

And since we don't need the decor view marking anymore, and since the `View` --> `Element` cache was moved out of `ViewGroupDescriptor.ElementContext` into `ViewGroupDescriptor` ( #206 ), that means we can just delete `ViewGroupDescriptor.ElementContext`. This saves 1 allocation per `ViewGroup`. Hoo-ray :)

Closes #205

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewGroupDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewGroupDescriptor.java
@@ -13,19 +13,14 @@ import android.view.View;
 import android.view.ViewGroup;
 
 import com.facebook.stetho.common.Accumulator;
-import com.facebook.stetho.common.Util;
 import com.facebook.stetho.common.android.FragmentCompatUtil;
 import com.facebook.stetho.inspector.elements.ChainedDescriptor;
 
 import java.util.Collections;
-import java.util.IdentityHashMap;
 import java.util.Map;
 import java.util.WeakHashMap;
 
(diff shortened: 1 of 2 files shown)


jasta:
:+1:

PR #211, opened 2015-06-29
Implementation of DialogFragment support for Elements

### LINKED ISSUE #193: Dialogs and DialogFragment are not shown in Elements tab
Dialogs are invisible to the Elements tab which currently only considers the Activity window's decor view and all subviews.  This finds most normal fragments, but not dialogs due to the fact that dialogs used their own window and therefore have a separate decor view.

We can add support for this in a similar fashion to `adb shell dumpsys activity top` which visualizes both the Activity view tree and also all detached fragment's view trees separately.

rickbrew:
This adds support for `DialogFragment`s in the Elements tab. This code is actually pretty straightforward now that tree diffing is checked in ( #204 ), without which I had code with all sorts of clowney timers and stuff.

I refactored `ChainedDescriptor` to be an interface, with `AbstractChainedDescriptor` sitting on top of it as the canonical implementation. I did this because `DialogFragmentDescriptor` has the unique problem of needing to *not* pass-through to its super-descriptor's implementation of `getChildren()`. If it did, then it would emit both the `View` and the `Window` which contains the `View`. Not only would this be silly, but it causes an infinite loop and would break all sorts of stuff that assume each element can only appear once in the DOM tree. I chose this approach because the code was much simpler, and free of future bug and clown hazards, versus allowing classes that derive from `AbstractChainedDescriptor` to override the non-"on" methods (e.g. to implement a non-chained version of `getChildren`).

Note that this does *not* add support for `Dialog`s that aren't attached to a `DialogFragment`. If anyone needs that then we can go ahead and add support for it, but for now it's just so much simpler to just support `DialogFragment`.

Closes #193

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/common/android/FragmentCompatUtil.java
+++ b/stetho/src/main/java/com/facebook/stetho/common/android/FragmentCompatUtil.java
@@ -20,6 +20,22 @@ public final class FragmentCompatUtil {
   private FragmentCompatUtil() {
   }
 
+  public static boolean isDialogFragment(Object fragment) {
+    FragmentCompat supportLib = FragmentCompat.getSupportLibInstance();
+    if (supportLib != null &&
+        supportLib.getDialogFragmentClass().isInstance(fragment)) {
+      return true;
+    }
+
+    FragmentCompat framework = FragmentCompat.getFrameworkInstance();
+    if (framework != null &&
+        framework.getDialogFragmentClass().isInstance(fragment)) {
+      return true;
+    }
+
+    return false;
+  }
+
   @Nullable
   public static Object findFragmentForView(View view) {
     Activity activity = ViewUtil.tryGetActivity(view);
@@ -62,10 +78,12 @@ public final class FragmentCompatUtil {
       FragmentCompat compat,
       Activity activity,
       View view) {
-    return findFragmentForViewInFragmentManager(
-        compat,
-        compat.forFragmentActivity().getFragmentManager(activity),
-        view);
+    Object fragmentManager = compat.forFragmentActivity().getFragmentManager(activity);
+    if (fragmentManager != null) {
+      return findFragmentForViewInFragmentManager(compat, fragmentManager, view);
+    } else {
+      return null;
+    }
   }
 
   @Nullable
@@ -76,7 +94,7 @@ public final class FragmentCompatUtil {
     List<?> fragments = compat.forFragmentManager().getAddedFragments(fragmentManager);
 
     if (fragments != null) {
-      for (int i = 0; i < fragments.size(); ++i) {
+      for (int i = 0, N = fragments.size(); i < N; ++i) {
         Object fragment = fragments.get(i);
         Object result = findFragmentForViewInFragment(compat, fragment, view);
         if (result != null) {
--- a/stetho/src/main/java/com/facebook/stetho/common/android/ViewUtil.java
+++ b/stetho/src/main/java/com/facebook/stetho/common/android/ViewUtil.java
@@ -12,6 +12,7 @@ package com.facebook.stetho.common.android;
 import android.annotation.TargetApi;
 import android.app.Activity;
 import android.content.Context;
+import android.content.ContextWrapper;
 import android.graphics.PointF;
 import android.os.Build;
 import android.view.View;
@@ -148,14 +149,31 @@ public final class ViewUtil {
     }
 
     Context context = view.getContext();
-    if (context instanceof Activity) {
-      return (Activity)context;
+
+    Activity activityFromContext = tryGetActivity(context);
+    if (activityFromContext != null) {
+      return activityFromContext;
     }
 
     ViewParent parent = view.getParent();
     if (parent instanceof View) {
       View parentView = (View)parent;
       return tryGetActivity(parentView);
+    }
+
+    return null;
+  }
+
+  @Nullable
+  private static Activity tryGetActivity(Context context) {
+    while (context != null) {
+      if (context instanceof Activity) {
+        return (Activity) context;
+      } else if (context instanceof ContextWrapper) {
+        context = ((ContextWrapper) context).getBaseContext();
+      } else {
+        return null;
+      }
     }
 
     return null;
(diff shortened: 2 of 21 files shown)


stetho/src/main/java/com/facebook/stetho/common/android/FragmentCompatUtil.java
```diff
@@ -62,10 +78,12 @@ private static Object findFragmentForViewInActivity(
       FragmentCompat compat,
       Activity activity,
       View view) {
-    return findFragmentForViewInFragmentManager(
-        compat,
-        compat.forFragmentActivity().getFragmentManager(activity),
-        view);
+    Object fragmentManager = compat.forFragmentActivity().getFragmentManager(activity);
+    if (fragmentManager != null) {
```
jasta: Is this a separate issue being fixed?  I'm fine with it in context, just curious

stetho/src/main/java/com/facebook/stetho/common/android/ViewUtil.java
```diff
@@ -161,6 +164,23 @@ public static Activity tryGetActivity(View view) {
     return null;
   }
 
+  private static Activity tryGetActivity(Context context) {
+    if (context instanceof Activity) {
+      return (Activity) context;
+    }
+
+    if (context instanceof ContextWrapper) {
+      final ContextWrapper contextWrapper = (ContextWrapper) context;
```
jasta: Prefer a while loop over recursion.  Faster and simpler (IMO).

```java
while (context != null) {
  if (context instanceof Activity) {
    return (Activity)context;
  } else if (context instanceof ContextWrapper) {
    context = ((ContextWrapper)context).getBaseContext();
  } else {
    return null;
  }
}
return null;
```

stetho/src/main/java/com/facebook/stetho/inspector/elements/AbstractChainedDescriptor.java
```diff
@@ -0,0 +1,147 @@
...
+import com.facebook.stetho.common.Accumulator;
+import com.facebook.stetho.common.ThreadBound;
+import com.facebook.stetho.common.Util;
+
+import javax.annotation.Nullable;
+
+/**
+ * This class derives from {@link Descriptor} and provides a canonical implementation of
+ * {@link ChainedDescriptor}.
+ *
```
jasta: `<p/>`

stetho/src/main/java/com/facebook/stetho/inspector/elements/Descriptor.java
```diff
@@ -73,13 +73,13 @@ public final void removeCallbacks(Runnable r) {
    * @param text the text argument to be parsed
    * @return a map of attributes to their respective values to be set.
    */
-  protected Map<String, String> parseSetAttributesAsTextArg(String text) {
+  protected static Map<String, String> parseSetAttributesAsTextArg(String text) {
```
jasta: generally prefer this kind of change in a separate commit, but i'm fine with it for nwo.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ActivityDescriptor.java
```diff
@@ -44,4 +52,32 @@ public View getViewForHighlighting(Object element) {
 
     return null;
   }
+
+  private static void getDialogFragments(
+      FragmentCompat compat,
```
jasta: `@Nullable`

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/DialogDescriptor.java
```diff
@@ -0,0 +1,41 @@
...
+    if (window != null) {
+      children.store(window);
+    }
+  }
+
+  @Nullable
+  @Override
+  public View getViewForHighlighting(Object element) {
+    if (getHost() instanceof AndroidDescriptorHost) {
+      final AndroidDescriptorHost host = (AndroidDescriptorHost) getHost();
```
jasta: nit: store the result of `getHost()` and re-use the result.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/DialogFragmentDescriptor.java
```diff
@@ -0,0 +1,128 @@
...
+
+  private DialogFragmentDescriptor(FragmentCompat compat) {
+    mAccessor = compat.forDialogFragment();
+  }
+
+  @Override
+  public void setSuper(Descriptor superDescriptor) {
+    Util.throwIfNull(superDescriptor);
+
+    if (superDescriptor != mSuper) {
```
jasta: sigh, code duplication.  i understand why, but still, sigh :)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/DialogFragmentDescriptor.java
```diff
@@ -0,0 +1,128 @@
...
+  @Override
+  public void setAttributesAsText(Object element, String text) {
+    mSuper.setAttributesAsText(element, text);
+  }
+
+  @Nullable
+  @Override
+  public View getViewForHighlighting(Object element) {
+    if (getHost() instanceof AndroidDescriptorHost) {
+      final AndroidDescriptorHost host = (AndroidDescriptorHost) getHost();
```
jasta: nit about `getHost` again.

jasta:
lgtm less a few minor notes.  Would prefer a screenshot for the PR so I can see in context where the dialog fragments actually sit in the hierarchy.  Thanks so much for working on this!

stetho/src/main/java/com/facebook/stetho/common/android/FragmentCompatUtil.java
rickbrew: yeah. getFragmentManager()'s return type is @Nullable. Our implementation will never return null, but it used to when my code was in there that used reflection to peek at the field vs. calling the method which defer-creates the FragmentManager. We should fix that but I'm not sure of how going back to reflection here will affect performance (it's fairly low-pri, i.o.w.)

stetho/src/main/java/com/facebook/stetho/common/android/ViewUtil.java
rickbrew: sounds good

PR #219, opened 2015-07-09
Add async pretty printing binary payload

mengdilin:
This PR provides the necessary interfaces and abstract class that provide the asynchronous pretty printing mechanism for stetho.

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinter.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinter.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.network;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.PrintWriter;
+
+/**
+ * Interface that callers need to implement in order to pretty print binary payload received by Stetho
+ */
+public interface AsyncPrettyPrinter {
+  /**
+   * Prints the prettified version of payload to output. This method can block
+   * for a certain period of time. Note that Stetho may impose arbitrary
+   * time out on this method.
+   *
+   * @param output Writes the prettified version of payload
+   * @param payload Response stream that has the raw data to be prettified
+   * @throws IOException
+   */
+  public void printTo(PrintWriter output, InputStream payload) throws IOException;
+
+  /**
+   * Specifies the type of pretty printed content. Note that this method is called
+   * before the content is actually pretty printed. Stetho uses this
+   * method to make a hopeful guess of the type of prettified content
+   *
+   * @return an enum defined by PrettyPrinterDisplayType class
+   */
+  public PrettyPrinterDisplayType getPrettifiedType();
+}
--- a/stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterExecutorHolder.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterExecutorHolder.java
@@ -1,0 +1,20 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.network;
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+/**
+ * A holder class for the executor service used for pretty printing related tasks
+ */
+class AsyncPrettyPrinterExecutorHolder {
+  public static final ExecutorService sExecutorService = Executors.newCachedThreadPool();
+}
--- a/stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterFactory.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterFactory.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.network;
+
+import com.facebook.stetho.inspector.protocol.module.Page;
+
+/**
+ * Interface for creating a factory for asynchronous pretty printers
+ */
+public interface AsyncPrettyPrinterFactory {
+
+  /**
+   * Creates an asynchronous pretty printer. This method must not be blocking.
+   *
+   * @param headerName header name of a response which is used to associate
+   * with an asynchronous pretty printer
+   * @param headerValue header value of a response which contains the URI for
+   * the schema data needed to pretty print the response body
+   * @return an asynchronous pretty printer to prettify the response body
+   */
+  public AsyncPrettyPrinter getInstance(String headerName, String headerValue);
+}
--- a/stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterRegistry.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterRegistry.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.network;
+
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.ThreadSafe;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+@ThreadSafe
+public class AsyncPrettyPrinterRegistry {
+
+  private final Map<String, AsyncPrettyPrinterFactory> mRegistry = new HashMap<>();
+
+  public synchronized void register(String headerName, AsyncPrettyPrinterFactory factory) {
+    mRegistry.put(headerName, factory);
+  }
+
+  @Nullable
+  public synchronized AsyncPrettyPrinterFactory lookup(String headerName) {
+    return mRegistry.get(headerName);
+  }
+
+  public synchronized boolean unregister(String headerName) {
+    return mRegistry.remove(headerName) != null;
+  }
+}
--- a/stetho/src/main/java/com/facebook/stetho/inspector/network/DownloadingAsyncPrettyPrinterFactory.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/network/DownloadingAsyncPrettyPrinterFactory.java
@@ -1,0 +1,165 @@
+package com.facebook.stetho.inspector.network;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.PrintWriter;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+
+import com.facebook.stetho.common.ExceptionUtil;
+import com.facebook.stetho.common.Util;
+
+import javax.annotation.Nullable;
+
+/**
+ * Abstract class for pretty printer factory that asynchronously downloads schema needed for
+ * pretty printing the payload
+ */
+public abstract class DownloadingAsyncPrettyPrinterFactory implements AsyncPrettyPrinterFactory {
+
+  private final static ExecutorService sExecutorService =
+      AsyncPrettyPrinterExecutorHolder.sExecutorService;
+
+  @Override
+  public AsyncPrettyPrinter getInstance(final String headerName, final String headerValue) {
+
+    final MatchResult result = matchAndParseHeader(headerName, headerValue);
+    if (result == null) {
+      return null;
+    }
+    String uri = result.getSchemaUri();
+    URL schemaURL = parseURL(uri);
+    if (schemaURL == null) {
+      return getErrorAsyncPrettyPrinter(headerName, headerValue);
+    } else {
+      final Future<String> response = sExecutorService.submit(new Request(schemaURL));
+      return new AsyncPrettyPrinter() {
+        public void printTo(PrintWriter output, InputStream payload)
+            throws IOException {
+          try {
+            String schema;
+            try {
+              schema = response.get();
+            } catch (ExecutionException e) {
+              Throwable cause = e.getCause();
+              if (IOException.class.isInstance(cause)) {
+                doErrorPrint(
+                    output,
+                    payload,
+                    "Cannot successfully download schema: "
+                        + e.getMessage());
+                return;
+              } else {
+                throw e;
+              }
+            }
+            doPrint(output, payload, schema);
+          } catch (InterruptedException e) {
+            doErrorPrint(
+                output,
+                payload,
+                "Encountered spurious interrupt while downloading schema for pretty printing: "
+                    + e.getMessage());
+          } catch (ExecutionException e) {
+            Throwable cause = e.getCause();
+            throw ExceptionUtil.propagate(cause);
+          }
+        }
+
+        public PrettyPrinterDisplayType getPrettifiedType() {
+          return result.getDisplayType();
+        }
+      };
+    }
+  }
+
+  /**
+   * Match the correct header that contains information about the schema uri
+   * @param headerName header name of a response that needs to be pretty printed
+   * @param headerValue header value which contains the URI for
+   * the schema data needed to pretty print the response body
+   * @return MatchResult that has the schema uri and the type of prettified result. Null
+   * if there is no correct header match.
+   */
+  @Nullable
+  protected abstract MatchResult matchAndParseHeader(String headerName, String headerValue);
+
+  /**
+   * Note that the IOException thrown by this method will be propagated all the way up and
+   * yield an error to the chrome devtools
+   */
+  protected abstract void doPrint(PrintWriter output, InputStream payload, String schema)
+      throws IOException;
+
+  @Nullable
+  private static URL parseURL(String uri) {
+    try {
+      return new URL(uri);
+    } catch (MalformedURLException e) {
+      return null;
+    }
+  }
+
+  private static void doErrorPrint(PrintWriter output, InputStream payload, String errorMessage)
+      throws IOException {
+    output.print(errorMessage + "\n" + Util.readAsUTF8(payload));
+  }
+
+  private static AsyncPrettyPrinter getErrorAsyncPrettyPrinter(
+      final String headerName,
+      final String headerValue) {
+    return new AsyncPrettyPrinter() {
+      @Override
+      public void printTo(PrintWriter output, InputStream payload) throws IOException {
+        String errorMessage = "[Failed to parse header: "
+            + headerName + " : " + headerValue + " ]";
+        doErrorPrint(output, payload, errorMessage);
+      }
+
+      @Override
+      public PrettyPrinterDisplayType getPrettifiedType() {
+        return PrettyPrinterDisplayType.TEXT;
+      }
+    };
+  }
+
+
+  protected class MatchResult {
+    private final String mSchemaUri;
+    private final PrettyPrinterDisplayType mDisplayType;
+
+    public MatchResult(String schemaUri, PrettyPrinterDisplayType displayType) {
+      mSchemaUri = schemaUri;
+      mDisplayType = displayType;
+    }
+
+    public String getSchemaUri() {
+      return mSchemaUri;
+    }
+
+    public PrettyPrinterDisplayType getDisplayType() {
+      return mDisplayType;
+    }
+  }
+
+  private static class Request implements Callable<String> {
+    private URL url;
+
+    public Request(URL url) {
+      this.url = url;
+    }
+
+    @Override
+    public String call() throws IOException {
+      InputStream urlStream = url.openStream();
+      String result = Util.readAsUTF8(urlStream);
+      urlStream.close();
+      return result;
+    }
+  }
+}
+
--- a/stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkEventReporterImpl.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkEventReporterImpl.java
@@ -148,12 +148,56 @@ public class NetworkEventReporterImpl implements NetworkEventReporter {
       receivedParams.frameId = "1";
       receivedParams.loaderId = "1";
       receivedParams.timestamp = stethoNow() / 1000.0;
-      receivedParams.type = contentType != null ?
-          getResourceTypeHelper().determineResourceType(contentType) :
-          Page.ResourceType.OTHER;
       receivedParams.response = responseJSON;
+      AsyncPrettyPrinter asyncPrettyPrinter =
+          initAsyncPrettyPrinterForResponse(response, peerManager);
+      receivedParams.type =
+          determineResourceType(asyncPrettyPrinter, contentType, getResourceTypeHelper());
       peerManager.sendNotificationToPeers("Network.responseReceived", receivedParams);
     }
+  }
+
+  @Nullable
+   static AsyncPrettyPrinter initAsyncPrettyPrinterForResponse(
+      InspectorResponse response,
+      NetworkPeerManager peerManager) {
+    AsyncPrettyPrinterRegistry registry = peerManager.getAsyncPrettyPrinterRegistry();
+    AsyncPrettyPrinter asyncPrettyPrinter = createPrettyPrinterForResponse(response, registry);
+    if (asyncPrettyPrinter != null) {
+      peerManager.getResponseBodyFileManager().associateAsyncPrettyPrinterWithId(
+          response.requestId(),
+          asyncPrettyPrinter);
+    }
+     return asyncPrettyPrinter;
+  }
+
+  private static Page.ResourceType determineResourceType(
+      AsyncPrettyPrinter asyncPrettyPrinter,
+      String contentType,
+      ResourceTypeHelper resourceTypeHelper) {
+    if (asyncPrettyPrinter != null) {
+      return asyncPrettyPrinter.getPrettifiedType().getResourceType();
+    } else {
+      return contentType != null ?
+          resourceTypeHelper.determineResourceType(contentType) :
+          Page.ResourceType.OTHER;
+    }
+  }
+
+  @Nullable
+  private static AsyncPrettyPrinter createPrettyPrinterForResponse(
+      InspectorResponse response,
+      AsyncPrettyPrinterRegistry registry) {
+    for (int i = 0, count = response.headerCount(); i < count; i++) {
+      AsyncPrettyPrinterFactory factory = registry.lookup(response.headerName(i));
+      if (factory != null) {
+        AsyncPrettyPrinter asyncPrettyPrinter = factory.getInstance(
+            response.headerName(i),
+            response.headerValue(i));
+        return asyncPrettyPrinter;
+      }
+    }
+    return null;
   }
 
   @Override
--- a/stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkPeerManager.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkPeerManager.java
@@ -20,6 +20,7 @@ public class NetworkPeerManager extends ChromePeerManager {
   private static NetworkPeerManager sInstance;
 
   private final ResponseBodyFileManager mResponseBodyFileManager;
+  private final AsyncPrettyPrinterRegistry mAsyncPrettyPrinterRegistry;
 
   @Nullable
   public static synchronized NetworkPeerManager getInstanceOrNull() {
@@ -38,10 +39,15 @@ public class NetworkPeerManager extends ChromePeerManager {
   public NetworkPeerManager(ResponseBodyFileManager responseBodyFileManager) {
     mResponseBodyFileManager = responseBodyFileManager;
     setListener(mTempFileCleanup);
+    mAsyncPrettyPrinterRegistry = new AsyncPrettyPrinterRegistry();
   }
 
   public ResponseBodyFileManager getResponseBodyFileManager() {
     return mResponseBodyFileManager;
+  }
+
+  public AsyncPrettyPrinterRegistry getAsyncPrettyPrinterRegistry() {
+    return mAsyncPrettyPrinterRegistry;
   }
 
   private final PeersRegisteredListener mTempFileCleanup = new PeersRegisteredListener() {
@@ -53,6 +59,7 @@ public class NetworkPeerManager extends ChromePeerManager {
     @Override
     protected void onLastPeerUnregistered() {
       mResponseBodyFileManager.cleanupFiles();
+      AsyncPrettyPrinterExecutorHolder.sExecutorService.shutdownNow();
     }
   };
 }
--- a/stetho/src/main/java/com/facebook/stetho/inspector/network/PrettyPrinterDisplayType.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/network/PrettyPrinterDisplayType.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.network;
+
+import com.facebook.stetho.inspector.protocol.module.Page;
+
+public enum PrettyPrinterDisplayType {
+  JSON(Page.ResourceType.XHR),
+  HTML(Page.ResourceType.DOCUMENT),
+  TEXT(Page.ResourceType.DOCUMENT);
+
+  private final Page.ResourceType mResourceType;
+
+  private PrettyPrinterDisplayType(Page.ResourceType resourceType) {
+    mResourceType = resourceType;
+  }
+
+  /**
+   * Converts PrettyPrinterDisplayType values to the appropriate
+   *  {@link Page.ResourceType} values that Stetho understands
+   */
+  public Page.ResourceType getResourceType() {
+    return mResourceType;
+  }
+}
--- a/stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
@@ -15,11 +15,22 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.io.PrintWriter;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 import android.content.Context;
 import android.util.Base64;
 import android.util.Base64OutputStream;
 
+import com.facebook.stetho.common.ExceptionUtil;
 import com.facebook.stetho.common.LogRedirector;
 import com.facebook.stetho.common.Util;
 
@@ -29,8 +40,14 @@ import com.facebook.stetho.common.Util;
 public class ResponseBodyFileManager {
   private static final String TAG = "ResponseBodyFileManager";
   private static final String FILENAME_PREFIX = "network-response-body-";
+  private static final int PRETTY_PRINT_TIMEOUT_SEC = 10;
 
   private final Context mContext;
+
+  private final Map<String, AsyncPrettyPrinter> mRequestIdMap = Collections.synchronizedMap(
+      new HashMap<String, AsyncPrettyPrinter>());
+  private final ExecutorService sExecutorService =
+      AsyncPrettyPrinterExecutorHolder.sExecutorService;
 
   public ResponseBodyFileManager(Context context) {
     mContext = context;
@@ -56,17 +73,40 @@ public class ResponseBodyFileManager {
       }
       ResponseBodyData bodyData = new ResponseBodyData();
       bodyData.base64Encoded = firstByte != 0;
-      bodyData.data = readContentsAsUTF8(in);
+
+      AsyncPrettyPrinter asyncPrettyPrinter = mRequestIdMap.get(requestId);
+      if (asyncPrettyPrinter != null) {
+        // TODO: this line blocks for up to 10 seconds and create problems as described
+        // in issue #243 allow asynchronous dispatch for MethodDispatcher
+        bodyData.data = prettyPrintContentWithTimeOut(asyncPrettyPrinter, in);
+      } else {
+        bodyData.data = Util.readAsUTF8(in);
+      }
       return bodyData;
+
     } finally {
       in.close();
     }
   }
 
-  private static String readContentsAsUTF8(InputStream in) throws IOException {
-    ByteArrayOutputStream out = new ByteArrayOutputStream();
-    Util.copy(in, out, new byte[1024]);
-    return out.toString("UTF-8");
+  private String prettyPrintContentWithTimeOut(
+      AsyncPrettyPrinter asyncPrettyPrinter,
+      InputStream in) throws IOException {
+    AsyncPrettyPrintingCallable prettyPrintingCallable = new AsyncPrettyPrintingCallable(
+        in,
+        asyncPrettyPrinter);
+    Future<String> future = sExecutorService.submit(prettyPrintingCallable);
+    try {
+      return Util.getUninterruptibly(future, PRETTY_PRINT_TIMEOUT_SEC, TimeUnit.SECONDS);
+    } catch (TimeoutException e) {
+      future.cancel(true);
+      return "Time out after " + PRETTY_PRINT_TIMEOUT_SEC +
+          " seconds of attempting to pretty print\n" + Util.readAsUTF8(in);
+    }  catch (ExecutionException e) {
+      Throwable cause = e.getCause();
+      ExceptionUtil.propagateIfInstanceOf(cause, IOException.class);
+      throw ExceptionUtil.propagate(cause);
+    }
   }
 
   public OutputStream openResponseBodyFile(String requestId, boolean base64Encode)
@@ -82,5 +122,49 @@ public class ResponseBodyFileManager {
 
   private static String getFilename(String requestId) {
     return FILENAME_PREFIX + requestId;
+  }
+
+  /**
+   * Associates an asynchronous pretty printer with a response request id
+   * The pretty printer will be used to pretty print the response body that has
+   * the particular request id
+   *
+   * @param requestId Unique identifier for the response
+   * as per {@link NetworkEventReporter.InspectorResponse#requestId()}
+   * @param asyncPrettyPrinter Asynchronous Pretty Printer to pretty print the response body
+   */
+  public void associateAsyncPrettyPrinterWithId(
+      String requestId,
+      AsyncPrettyPrinter asyncPrettyPrinter) {
+    if (mRequestIdMap.put(requestId, asyncPrettyPrinter) != null) {
+      throw new IllegalArgumentException("cannot associate different " +
+          "pretty printers with the same request id: "+requestId);
+    }
+  }
+
+  private class AsyncPrettyPrintingCallable implements Callable<String> {
+    private final InputStream mInputStream;
+    private final AsyncPrettyPrinter mAsyncPrettyPrinter;
+
+    public AsyncPrettyPrintingCallable(
+        InputStream in,
+        AsyncPrettyPrinter asyncPrettyPrinter) {
+      mInputStream = in;
+      mAsyncPrettyPrinter = asyncPrettyPrinter;
+    }
+
+    @Override
+    public String call() throws IOException {
+      return prettyPrintContent(mInputStream, mAsyncPrettyPrinter);
+    }
+
+    private String prettyPrintContent(InputStream in, AsyncPrettyPrinter asyncPrettyPrinter)
+        throws IOException {
+      ByteArrayOutputStream out = new ByteArrayOutputStream();
+      PrintWriter writer = new PrintWriter(out);
+      asyncPrettyPrinter.printTo(writer, in);
+      writer.flush();
+      return out.toString("UTF-8");
+    }
   }
 }
--- a/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Network.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Network.java
@@ -19,6 +19,7 @@ import com.facebook.stetho.inspector.jsonrpc.JsonRpcPeer;
 import com.facebook.stetho.inspector.jsonrpc.JsonRpcResult;
 import com.facebook.stetho.inspector.jsonrpc.protocol.JsonRpcError;
 import com.facebook.stetho.inspector.network.NetworkPeerManager;
+import com.facebook.stetho.inspector.network.AsyncPrettyPrinterRegistry;
 import com.facebook.stetho.inspector.network.ResponseBodyData;
 import com.facebook.stetho.inspector.network.ResponseBodyFileManager;
 import com.facebook.stetho.inspector.protocol.ChromeDevtoolsDomain;
@@ -84,6 +85,10 @@ public class Network implements ChromeDevtoolsDomain {
     response.body = bodyData.data;
     response.base64Encoded = bodyData.base64Encoded;
     return response;
+  }
+
+  public AsyncPrettyPrinterRegistry getAsyncPrettyPrinterRegistry() {
+    return mNetworkPeerManager.getAsyncPrettyPrinterRegistry();
   }
 
   private static class GetResponseBodyResponse implements JsonRpcResult {
--- a/stetho/src/test/java/com/facebook/stetho/inspector/network/AsyncPrettyPrintResponseBodyTest.java
+++ b/stetho/src/test/java/com/facebook/stetho/inspector/network/AsyncPrettyPrintResponseBodyTest.java
@@ -1,0 +1,254 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.network;
+
+import com.facebook.stetho.common.Util;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.eq;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.never;
+import static org.mockito.Matchers.any;
+
+import com.facebook.stetho.inspector.network.AsyncPrettyPrinter;
+import dalvik.annotation.TestTargetClass;
+import org.junit.Before;
+import org.junit.Test;
+
+import javax.annotation.Nullable;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.lang.Override;
+import java.lang.String;
+import java.util.ArrayList;
+import java.util.Arrays;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+public class AsyncPrettyPrintResponseBodyTest {
+  private static final String TEST_REQUEST_ID = "1234";
+  private static final String TEST_HEADER_NAME = "header name";
+  private static final String TEST_HEADER_VALUE = "header value";
+  private static final String PRETTY_PRINT_PREFIX = "pretty printed result: ";
+  private static final String[] UNREGISTERED_HEADER_NAMES = {"unregistered header name 1",
+      "unregistered header name 2",
+      "unregistered header name 3"};
+  private static final String[] UNREGISTERED_HEADER_VALUES = {"unregistered header value 1",
+      "unregistered header value 2",
+      "unregistered header value 3"};
+  private static final byte[] TEST_RESPONSE_BODY;
+  private static final ByteArrayInputStream mInputStream;
+
+  static {
+    int responseBodyLength = 4096 * 2 + 2048; // span multiple buffers when tee-ing
+    TEST_RESPONSE_BODY = new byte[responseBodyLength];
+    for (int i = 0; i < responseBodyLength; i++) {
+      TEST_RESPONSE_BODY[i] = positionToByte(i);
+    }
+    mInputStream = new ByteArrayInputStream(TEST_RESPONSE_BODY);
+  }
+
+  private AsyncPrettyPrinterRegistry mAsyncPrettyPrinterRegistry;
+  private PrettyPrinterTestFactory mPrettyPrinterTestFactory;
+  private NetworkPeerManager mNetworkPeerManager;
+  private ResponseBodyFileManager mResponseBodyFileManager;
+
+  @Before
+  public void setup() {
+    mPrettyPrinterTestFactory = new PrettyPrinterTestFactory();
+    mResponseBodyFileManager = mock(ResponseBodyFileManager.class);
+    mNetworkPeerManager = new NetworkPeerManager(mResponseBodyFileManager);
+    mAsyncPrettyPrinterRegistry = mNetworkPeerManager.getAsyncPrettyPrinterRegistry();
+    mAsyncPrettyPrinterRegistry.register(TEST_HEADER_NAME, mPrettyPrinterTestFactory);
+  }
+
+  @Test
+  public void testAsyncPrettyPrinterResult() throws IOException {
+    StringWriter out = new StringWriter();
+    PrintWriter writer = new PrintWriter(out);
+    AsyncPrettyPrinter mAsyncPrettyPrinter = mPrettyPrinterTestFactory.getInstance(
+        TEST_HEADER_NAME,
+        TEST_HEADER_VALUE);
+    mAsyncPrettyPrinter.printTo(writer, mInputStream);
+    assertEquals(PRETTY_PRINT_PREFIX + Arrays.toString(TEST_RESPONSE_BODY), out.toString());
+  }
+
+  @Test
+  public void testInitAsyncPrettyPrinterForResponseWithRegisteredHeader() {
+    ArrayList<String> headerNames = new ArrayList<String>();
+    ArrayList<String> headerValues = new ArrayList<String>();
+
+    headerNames.add(UNREGISTERED_HEADER_NAMES[0]);
+    headerNames.add(UNREGISTERED_HEADER_NAMES[1]);
+    headerNames.add(TEST_HEADER_NAME);
+    headerValues.add(UNREGISTERED_HEADER_VALUES[0]);
+    headerValues.add(UNREGISTERED_HEADER_VALUES[1]);
+    headerValues.add(TEST_HEADER_VALUE);
+
+    TestInspectorResponse testResponse = new TestInspectorResponse(
+        headerNames,
+        headerValues,
+        TEST_REQUEST_ID
+    );
+    NetworkEventReporterImpl.initAsyncPrettyPrinterForResponse(
+        testResponse,
+        mNetworkPeerManager);
+    verify(mResponseBodyFileManager, times(1)).associateAsyncPrettyPrinterWithId(
+        eq(TEST_REQUEST_ID),
+        any(AsyncPrettyPrinter.class)
+    );
+  }
+
+  @Test
+  public void testInitAsyncPrettyPrinterForResponseWithUnregisteredHeader() {
+    ArrayList<String> headerNames = new ArrayList<String>();
+    ArrayList<String> headerValues = new ArrayList<String>();
+
+    headerNames.add(UNREGISTERED_HEADER_NAMES[0]);
+    headerNames.add(UNREGISTERED_HEADER_NAMES[1]);
+    headerNames.add(UNREGISTERED_HEADER_NAMES[2]);
+    headerValues.add(UNREGISTERED_HEADER_VALUES[0]);
+    headerValues.add(UNREGISTERED_HEADER_VALUES[1]);
+    headerValues.add(UNREGISTERED_HEADER_VALUES[2]);
+
+    TestInspectorResponse testResponse = new TestInspectorResponse(
+        headerNames,
+        headerValues,
+        TEST_REQUEST_ID
+    );
+    NetworkEventReporterImpl.initAsyncPrettyPrinterForResponse(
+        testResponse,
+        mNetworkPeerManager);
+    verify(mResponseBodyFileManager, never()).associateAsyncPrettyPrinterWithId(
+        eq(TEST_REQUEST_ID),
+        any(AsyncPrettyPrinter.class)
+    );
+  }
+
+  @Test
+  public void testGetInstanceWithUnmatchedHeader() {
+    ArrayList<String> headerNames = new ArrayList<String>();
+    ArrayList<String> headerValues = new ArrayList<String>();
+
+    headerNames.add(UNREGISTERED_HEADER_NAMES[0]);
+    headerNames.add(UNREGISTERED_HEADER_NAMES[1]);
+    headerNames.add(TEST_HEADER_NAME);
+    headerValues.add(UNREGISTERED_HEADER_VALUES[0]);
+    headerValues.add(UNREGISTERED_HEADER_VALUES[1]);
+    headerValues.add(UNREGISTERED_HEADER_VALUES[2]);
+
+    TestInspectorResponse testResponse = new TestInspectorResponse(
+        headerNames,
+        headerValues,
+        TEST_REQUEST_ID
+    );
+    AsyncPrettyPrinter asyncPrettyPrinter =
+        NetworkEventReporterImpl.initAsyncPrettyPrinterForResponse(
+        testResponse,
+        mNetworkPeerManager);
+    assertEquals(null, asyncPrettyPrinter);
+  }
+
+  private class PrettyPrinterTestFactory extends DownloadingAsyncPrettyPrinterFactory {
+    @Override
+    protected void doPrint(PrintWriter output, InputStream payload, String schema)
+        throws IOException {
+      ByteArrayOutputStream out = new ByteArrayOutputStream();
+      Util.copy(payload, out, new byte[1024]);
+      String prettifiedContent = PRETTY_PRINT_PREFIX + Arrays.toString(out.toByteArray());
+      output.write(prettifiedContent);
+      output.close();
+    }
+
+    @Override @Nullable
+    protected MatchResult matchAndParseHeader(String headerName, String headerValue) {
+      if (headerName.equals(TEST_HEADER_NAME) && headerValue.equals(TEST_HEADER_VALUE)) {
+        return new MatchResult("https://www.facebook.com", PrettyPrinterDisplayType.TEXT);
+      } else {
+        return null;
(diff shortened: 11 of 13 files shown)


stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkPeerManager.java
```diff
@@ -21,23 +21,31 @@
 
   private final ResponseBodyFileManager mResponseBodyFileManager;
 
+  private final AsyncPrettyPrinterRegistry mAsyncPrettyPrinterRegistry;
```
rickbrew: I'm not sure why this is being placed into Network's PeerManager ... what will it be used for here?

rickbrew:
This is a good start but we should get further along with the implementation before doing a pull request.

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkPeerManager.java
mengdilin: I need a way to access the registry in responseHeadersReceived method from NetworkEventReporterImpl, and it seems that I could access the registry through peerManager. I am also not so sure where the best place to place the registry. I have asked @jasta to take a look when he is available.

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
```diff
@@ -69,6 +79,13 @@ private static String readContentsAsUTF8(InputStream in) throws IOException {
     return out.toString("UTF-8");
   }
 
+  private String prettyPrintContent(InputStream in, AsyncPrettyPrinter asyncPrettyPrinter)
+      throws IOException {
+    ByteArrayOutputStream out = new ByteArrayOutputStream();
+    Util.copy(in, out, new byte[1024]);
```
rickbrew: is this going to happen frequently? if so then we really need to pool these byte arrays to avoid GC churn. can defer on this for now, but if so then make a note to definitely tackle this later. and I think there's a few other places where we do this and we should be pooling there as well, so solving it here could provide a solution in those other places as well

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
```diff
@@ -32,6 +34,8 @@
 
   private final Context mContext;
 
+  private Map<String, AsyncPrettyPrinter> mRequestIdMap = new HashMap<>();
```
rickbrew: this might be accessed by multiple threads? if so, let's make sure we `Collections.synchronizedMap()` it

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkEventReporterImpl.java
```diff
@@ -152,6 +152,22 @@ public void responseHeadersReceived(InspectorResponse response) {
           getResourceTypeHelper().determineResourceType(contentType) :
           Page.ResourceType.OTHER;
       receivedParams.response = responseJSON;
+
+      AsyncPrettyPrinterRegistry registry= peerManager.getAsyncPrettyPrinterRegistry();
+      AsyncPrettyPrinterFactory factory = null;
+      for (int i = 0; i < response.headerCount(); i++) {
```
rickbrew: recommend a for-loop pattern like this to reduce per-iteration overhead:
`for (int i = 0, count = response.headerCount(); i < count; i++) {`
(yeah, this can actually matter on android!)

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkEventReporterImpl.java
```diff
@@ -152,6 +152,22 @@ public void responseHeadersReceived(InspectorResponse response) {
           getResourceTypeHelper().determineResourceType(contentType) :
           Page.ResourceType.OTHER;
       receivedParams.response = responseJSON;
+
+      AsyncPrettyPrinterRegistry registry= peerManager.getAsyncPrettyPrinterRegistry();
+      AsyncPrettyPrinterFactory factory = null;
```
rickbrew: if we're not using this outside of the loop, then we can just declare it inside the loop, e.g.
`AsyncPrettyPrinterFactory factory = registry.lookup(response.headerName(i));`

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkEventReporterImpl.java
```diff
@@ -152,6 +152,22 @@ public void responseHeadersReceived(InspectorResponse response) {
           getResourceTypeHelper().determineResourceType(contentType) :
           Page.ResourceType.OTHER;
       receivedParams.response = responseJSON;
+
+      AsyncPrettyPrinterRegistry registry= peerManager.getAsyncPrettyPrinterRegistry();
```
rickbrew: nit: spaces surrounding operators, `registry = peerManager`

stetho/src/main/java/com/facebook/stetho/inspector/network/DefaultAsyncPrettyPrinterRegistry.java
```diff
@@ -0,0 +1,50 @@
...
+
+package com.facebook.stetho.inspector.network;
+
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.GuardedBy;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class DefaultAsyncPrettyPrinterRegistry implements AsyncPrettyPrinterRegistry {
```
rickbrew: Instead of having this be a static singleton that will stick around even after a client has disconnected, can we make it ... not be a singleton? I don't see any calls to `getInstance()` or `register()` yet, so it's not clear to me yet how this is intended to be used. In fact, we may not need `AsyncPrettyPrinterRegistry` to be an interface. I'm not sure this needs to be a pivot point for generalization; see `DescriptorMap` for an example of something like this which is *not* done via an interface. (If we go down this route, then we would remove the interface and rename this class to remove the word 'Default'.)

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterRegistry.java
```diff
@@ -0,0 +1,16 @@
...
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.network;
+
+public interface AsyncPrettyPrinterRegistry {
+  public boolean register (String headerName, AsyncPrettyPrinterFactory factory);
+  public AsyncPrettyPrinterFactory lookup (String headerName);
+  public void unregister(String headerName);
```
rickbrew: return type should probably be `boolean` to match the convention in Java: true means the thing was found and removed, false means it wasn't found and the collection is unchanged

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterRegistry.java
```diff
@@ -0,0 +1,16 @@
...
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.network;
+
+public interface AsyncPrettyPrinterRegistry {
+  public boolean register (String headerName, AsyncPrettyPrinterFactory factory);
+  public AsyncPrettyPrinterFactory lookup (String headerName);
```
rickbrew: should probably add a `@Nullable` to this because it can return 'null' if something isn't registered for headerName. right?

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterRegistry.java
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.network;
+
+public interface AsyncPrettyPrinterRegistry {
+  public boolean register (String headerName, AsyncPrettyPrinterFactory factory);
```
rickbrew: nit: no space after `register` and before `(` (and same for the next line with the `lookup` method)

rickbrew:
Alright there's my first round of feedback. Still need something implementing and using this so we can move forward more :)

jasta:
@mengdilin your commit message and title could use some work.  In particular, you're referring to the internal implementation details before you're describing what it does and what problem it addresses.

stetho/src/main/java/com/facebook/stetho/Stetho.java
```diff
@@ -141,7 +142,7 @@ public static InspectorModulesProvider defaultInspectorModulesProvider(final Con
         modules.add(new DOMStorage(context));
         modules.add(new HeapProfiler());
         modules.add(new Inspector());
-        modules.add(new Network(context));
+        modules.add(new Network(context, DefaultAsyncPrettyPrinterRegistry.getInstance()));
```
jasta: This is awkward as a singleton.  Strongly prefer `new ...`

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinter.java
```diff
@@ -0,0 +1,14 @@
...
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.network;
+
+public interface AsyncPrettyPrinter {
```
jasta: Javadoc.  Third party callers will be implementing this.

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterFactory.java
```diff
@@ -0,0 +1,14 @@
...
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.network;
+
+public interface AsyncPrettyPrinterFactory {
```
jasta: javadoc here as well

stetho/src/main/java/com/facebook/stetho/inspector/network/DefaultAsyncPrettyPrinterRegistry.java
```diff
@@ -0,0 +1,50 @@
...
+  public static synchronized DefaultAsyncPrettyPrinterRegistry getInstance() {
+    if (sInstance == null) {
+      sInstance = new DefaultAsyncPrettyPrinterRegistry();
+    }
+    return sInstance;
+  }
+
+  @Override
+  public synchronized boolean register(String headerName, AsyncPrettyPrinterFactory factory) {
+    if (mRegistry.containsKey(headerName)) {
```
jasta: Inefficient and a bad pattern.  If a caller wishes to register twice, you should either allow it or throw as a warning.  I prefer simply allowing it.

FWIW, `put` returns the old value which means you can do this without a perf hit in the common case.

stetho/src/main/java/com/facebook/stetho/inspector/network/DefaultAsyncPrettyPrinterRegistry.java
```diff
@@ -0,0 +1,50 @@
...
+      return false;
+    }
+    mRegistry.put(headerName, factory);
+    return true;
+  }
+
+
+  @Override
+  public synchronized @Nullable AsyncPrettyPrinterFactory lookup(String headerName) {
```
jasta: style nit: `@Nullable` goes next to `@Override`

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkEventReporterImpl.java
```diff
@@ -152,6 +152,22 @@ public void responseHeadersReceived(InspectorResponse response) {
...
+      AsyncPrettyPrinterFactory factory = null;
+      for (int i = 0; i < response.headerCount(); i++) {
+        factory = registry.lookup(response.headerName(i));
+        if (factory != null) {
+          AsyncPrettyPrinter asyncPrettyPrinter = factory.getInstance(
```
jasta: This all probably deserves a function to clean up the logic

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkPeerManager.java
jasta: @rickbrew the peer manager is the only context given to `NetworkEventReporterImpl`.  `Network` owns/constructs the peer manager.  `NetworkEventReporterImpl` accesses the instance that was constructed.

They talk to each other through a singleton because I didn'tw ant to ask third party callers to have to keep track of the default modules provider that they passed in through `Stetho.initialize`.

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinter.java
```diff
@@ -0,0 +1,14 @@
...
+  String print(byte[] payload);
```
jasta: Strongly prefer a more efficient API:

  void printTo(PrintWriter output, InputStream payload) throws IOException;

You can internally use a byte[] and String but this API lets you optimize later.

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
```diff
@@ -83,4 +100,12 @@ public OutputStream openResponseBodyFile(String requestId, boolean base64Encode)
   private static String getFilename(String requestId) {
     return FILENAME_PREFIX + requestId;
   }
+
+  public boolean addAsyncPrettyPrinter(String requestId, AsyncPrettyPrinter asyncPrettyPrinter) {
+    if (mRequestIdMap.containsKey(requestId)) {
```
jasta: This should throw.  I expect this to be a serious error if it occurs, no?

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkEventReporterImpl.java
```diff
@@ -152,6 +152,22 @@ public void responseHeadersReceived(InspectorResponse response) {
...
+              response.headerName(i),
+              response.headerValue(i));
+          peerManager.getResponseBodyFileManager().addAsyncPrettyPrinter(
```
jasta: The semantics don't look right here.  You aren't "adding" an async printer, you're associating it with a particular request.  Prefer `associateAsyncPrettyPrinter(...)`.

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
```diff
@@ -83,4 +100,12 @@ public OutputStream openResponseBodyFile(String requestId, boolean base64Encode)
   private static String getFilename(String requestId) {
     return FILENAME_PREFIX + requestId;
   }
+
+  public boolean addAsyncPrettyPrinter(String requestId, AsyncPrettyPrinter asyncPrettyPrinter) {
```
jasta: Prefer javadocs here.  It's unclear how this works at first glance.

jasta:
Have you tested this code in some capacity?  Without seeing a concrete implementation or unit tests I'm inclined to believe that it is entirely untested :)

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
```diff
@@ -56,7 +60,13 @@ public ResponseBodyData readFile(String requestId) throws IOException {
       }
       ResponseBodyData bodyData = new ResponseBodyData();
       bodyData.base64Encoded = firstByte != 0;
-      bodyData.data = readContentsAsUTF8(in);
+      if (mRequestIdMap.containsKey(requestId)) {
+        AsyncPrettyPrinter asyncPrettyPrinter = mRequestIdMap.get(requestId);
+        bodyData.data = prettyPrintContent(in, asyncPrettyPrinter);
```
jasta: We have a problem here.  This function can block for arbitrary lengths of time and in fact can time out.  Your code doesn't try to address that at all and could just wedge the UI.

Please consider a timeout here and also a test to see what the UI will do in the event that no response is given.

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
```diff
@@ -56,7 +60,13 @@ public ResponseBodyData readFile(String requestId) throws IOException {
...
+      }
+      else {
```
jasta: style nit: `} else {`

stetho/src/main/java/com/facebook/stetho/Stetho.java
mengdilin: To confirm I understand correctly with regard to the last suggestion, Network will have methods register and unregister which will internally ask async pretty printer registry to do the registering and unregistering?

stetho/src/main/java/com/facebook/stetho/inspector/network/DefaultAsyncPrettyPrinterRegistry.java
mengdilin: Thanks, I will change the registry to non static and remove the interface as well

mengdilin:
I have not tested my codes, but I will be adding unit tests soon!

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterFactory.java
```diff
@@ -0,0 +1,28 @@
...
+package com.facebook.stetho.inspector.network;
+
+/**
+ * Interface for creating a factory for asynchronous pretty printers
+ */
+public interface AsyncPrettyPrinterFactory {
+
+  /**
+   * Retrieves the Asynchronous Pretty Printer associated with the headerName
+   * and asynchronously download schema data from the URI provided in headerValue
```
jasta: This is an implementation detail.  Doesn't belong here...

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinter.java
```diff
@@ -0,0 +1,28 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.PrintWriter;
+
+/**
+ * Interface that callers need to implement in order to pretty print the payload received by Stetho
+ */
+public interface AsyncPrettyPrinter {
+  /**
+   * Prints the prettified version of payload to output
```
jasta: Explain that this method can block for a period of time, but warn implementors that the timeout may be arbitrarily imposed by stetho.

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterRegistry.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public class AsyncPrettyPrinterRegistry {
+
+  @GuardedBy("this")
+  private final Map<String, AsyncPrettyPrinterFactory> mRegistry = Collections.synchronizedMap(
+      new HashMap<String, AsyncPrettyPrinterFactory>());
```
jasta: `new HashMap<>();`

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterRegistry.java
```diff
@@ -0,0 +1,40 @@
...
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.GuardedBy;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public class AsyncPrettyPrinterRegistry {
+
+  @GuardedBy("this")
```
jasta: Err, it seems to me that it's guarded by an instance that you lost (`new HashMap<>()`).  `@GuardedBy` is intending to communicate the exact monitor lock used to ensure thread safety.  In this case, you can't specify that and are not explicitly enforcing the monitor locking yourself so the statement is invalid.

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterRegistry.java
```diff
@@ -0,0 +1,40 @@
...
+  private final Map<String, AsyncPrettyPrinterFactory> mRegistry = Collections.synchronizedMap(
+      new HashMap<String, AsyncPrettyPrinterFactory>());
+
+  public synchronized void register(String headerName, AsyncPrettyPrinterFactory factory) {
```
jasta: Redundant synchronization.  Pick which what you want and proceed with just 1 monitor lock.

stetho/src/main/java/com/facebook/stetho/inspector/network/BasePrettyPrinterFactory.java
```diff
@@ -0,0 +1,35 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.PrintWriter;
+
+import com.facebook.stetho.common.Util;
+import com.facebook.stetho.inspector.network.AsyncPrettyPrinter;
+import com.facebook.stetho.inspector.network.AsyncPrettyPrinterFactory;
+
+/**
+ * Implementation of {@link AsyncPrettyPrinterFactory} which generates asynchronous pretty printers
```
jasta: This documentation and the name is incorrect.  It should be `AbstractSchemaDownloadingPrettyPrinterFactory` or some such thing (maybe a shorter name would be wise *grin*)

stetho/src/main/java/com/facebook/stetho/inspector/network/BasePrettyPrinterFactory.java
```diff
@@ -0,0 +1,35 @@
...
+ * that allows callers to pretty print responses with specific header names
+ */
+public abstract class BasePrettyPrinterFactory implements AsyncPrettyPrinterFactory {
+
+  @Override
+  public AsyncPrettyPrinter getInstance(String headerName, String headerValue) {
+    //TODO: async download schema
```
jasta: This can't be landed as-is.  Are you intending to clean this up before the PR is merged?

stetho/src/main/java/com/facebook/stetho/inspector/network/BasePrettyPrinterFactory.java
```diff
@@ -0,0 +1,35 @@
...
+      public void printTo(PrintWriter output, InputStream payload) throws IOException {
+        ByteArrayOutputStream out = new ByteArrayOutputStream();
+        Util.copy(payload, out, new byte[1024]);
+        output.write(doPrint(out.toByteArray(), schema));
+        output.close();
+      }
+    };
+  }
+
+  protected abstract String doPrint(byte[] payload, String schema);
```
jasta: This dismantles the optimization I asked for.  Push the efficient interface all the way to the final implementor.

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterRegistry.java
```diff
@@ -0,0 +1,40 @@
...
+    mRegistry.put(headerName, factory);
+  }
+
+  @Nullable
+  public synchronized AsyncPrettyPrinterFactory lookup(String headerName) {
+    return mRegistry.get(headerName);
+  }
+
+  public synchronized boolean unregister(String headerName) {
+    if (mRegistry.remove(headerName) == null) {
```
jasta: `return mRegistry.remove(headerName) != null`

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkEventReporterImpl.java
```diff
@@ -152,10 +152,29 @@ public void responseHeadersReceived(InspectorResponse response) {
...
+    for (int i = 0, count = response.headerCount(); i < count; i++) {
+      AsyncPrettyPrinterFactory factory = registry.lookup(response.headerName(i));
+      if (factory != null) {
+        AsyncPrettyPrinter asyncPrettyPrinter = factory.getInstance(
+            response.headerName(i),
+            response.headerValue(i));
+        peerManager.getResponseBodyFileManager().associateAsyncPrettyPrinterWithId(
+            response.requestId(),
+            asyncPrettyPrinter);
+        break;
```
jasta: This is usually considered a bad code smell and can be corrected by dividing the logic into two methods:

```java
  public static void initAsyncPrettyPrineterForResponse(...) {
    prettyPrinter = matchPrettyPrinterToResponse(...);
    if (prettyPrinter != null) {
      associateAsyncPrettyPrinter(...);
    }
  }

  @Nullable
  public static AsyncPrettyPrinter matchPrettyPrinterToResponse(...);
```

Many engineers find this easier logic to follow because they can more easily predict the exact flow control semantics of `matchPrettyPrinterToResponse(...)` based on its name and signature.

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkPeerManager.java
```diff
@@ -21,29 +21,42 @@
 
   private final ResponseBodyFileManager mResponseBodyFileManager;
 
+  private AsyncPrettyPrinterRegistry mAsyncPrettyPrinterRegistry;
```
jasta: final, and remove the whitespace between this and `mResponseBodyFileManager`

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkPeerManager.java
```diff
@@ -21,29 +21,42 @@
...
+      Context context,
+      AsyncPrettyPrinterRegistry prettyPrinterRegistry) {
     if (sInstance == null) {
       sInstance = new NetworkPeerManager(
           new ResponseBodyFileManager(
-              context.getApplicationContext()));
+              context.getApplicationContext()),
+          prettyPrinterRegistry);
     }
+    sInstance.mAsyncPrettyPrinterRegistry = prettyPrinterRegistry;
```
jasta: Uhm, what?

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
jasta: @rickbrew is right.  This is dangerous.

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
```diff
@@ -69,6 +89,25 @@ private static String readContentsAsUTF8(InputStream in) throws IOException {
     return out.toString("UTF-8");
   }
 
+  private String prettyPrintContentWithTimeOut(
+      AsyncPrettyPrinter asyncPrettyPrinter,
+      InputStream in) throws IOException {
+    AsyncPrettyPrintingTask prettyPrintingTask = new AsyncPrettyPrintingTask(
+        in,
+        asyncPrettyPrinter);
+    ExecutorService executor = Executors.newSingleThreadExecutor();
```
jasta: Re-use this.

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
```diff
@@ -83,4 +122,48 @@ public OutputStream openResponseBodyFile(String requestId, boolean base64Encode)
...
+      String requestId,
+      AsyncPrettyPrinter asyncPrettyPrinter) {
+    if (mRequestIdMap.containsKey(requestId)) {
+      throw new IllegalArgumentException("cannot associate different " +
+          "pretty printers with the same request id: "+requestId);
+    }
+    mRequestIdMap.put(requestId, asyncPrettyPrinter);
+  }
+
+  private class AsyncPrettyPrintingTask implements Callable<String> {
```
jasta: `Task` is an overloaded term on Android and usually implies an `AsyncTask`.  Since you don't have this here, instead call this `Job` or `Callable`.

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
```diff
@@ -69,6 +89,25 @@ private static String readContentsAsUTF8(InputStream in) throws IOException {
...
+    AsyncPrettyPrintingTask prettyPrintingTask = new AsyncPrettyPrintingTask(
+        in,
+        asyncPrettyPrinter);
+    ExecutorService executor = Executors.newSingleThreadExecutor();
+    Future<String> future = executor.submit(prettyPrintingTask);
+    String prettifiedContent = "";
+    try {
+      prettifiedContent = future.get(10, TimeUnit.SECONDS);
+    } catch (TimeoutException | InterruptedException | ExecutionException e) {
+      future.cancel(true);
```
jasta: This is overly broad.  These 3 exceptions all represent very different problems and should be handled differently.  For starters, `InterruptedException` can occur spuriously and should be resumed.  For this, you will need a new method, `Util#getUninterruptibly` to match the others.

For `TimeoutException`, of course that's our "expected" case and can be handled as you've outlined here but with a slight tweak that you should explain to the caller in the message the reason that it failed (timeout after <n> seconds).

Finally, for `ExecutionException`, your `Callable` should not be throwing a checked exception so you can upgrade any exception that does occur to runtime exception and crash the app.  See `ExceptionUtil` in Stetho for tools to propagate when this happens.

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
```diff
@@ -69,6 +89,25 @@ private static String readContentsAsUTF8(InputStream in) throws IOException {
...
+      prettifiedContent = "Failed to pretty print data\n" + readContentsAsUTF8(in);
+    }
+    executor.shutdownNow();
```
jasta: You don't need this, but you can theoretically shut it down if you hook into `Network`'s cleanup code on the last peer disconnecting.

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
```diff
@@ -69,6 +89,25 @@ private static String readContentsAsUTF8(InputStream in) throws IOException {
...
 
+  private String prettyPrintContentWithTimeOut(
+      AsyncPrettyPrinter asyncPrettyPrinter,
+      InputStream in) throws IOException {
+    AsyncPrettyPrintingTask prettyPrintingTask = new AsyncPrettyPrintingTask(
+        in,
+        asyncPrettyPrinter);
+    ExecutorService executor = Executors.newSingleThreadExecutor();
+    Future<String> future = executor.submit(prettyPrintingTask);
+    String prettifiedContent = "";
```
jasta: This isn't needed and is also considered an anti-pattern.  Functions that return magic, undocumented values can be extremely frustrating to debug in large programs.  Avoid this whenever possible and instead use exceptions (even crashing the app can be better than this unpredictable behaviour).

For instance, imagine if you were a user and you were seeing empty payloads for some requests but not others and there were no error messages related to the problem.  What a frustrating problem to encounter...

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Network.java
```diff
@@ -33,8 +34,10 @@
   private final NetworkPeerManager mNetworkPeerManager;
   private final ResponseBodyFileManager mResponseBodyFileManager;
 
-  public Network(Context context) {
-    mNetworkPeerManager = NetworkPeerManager.getOrCreateInstance(context);
+  public Network(Context context, AsyncPrettyPrinterRegistry asyncPrettyPrinterRegistry) {
+    mNetworkPeerManager = NetworkPeerManager.getOrCreateInstance(
+        context,
+        asyncPrettyPrinterRegistry);
```
jasta: As I see this, it's clear to me that the `AsyncPrettyPrinterRegistry` should indeed be locally owned by `NetworkPeerManager`, not by `Network`.

stetho/src/test/java/com/facebook/stetho/inspector/network/AsyncPrettyPrintResponseBodyTest.java
```diff
@@ -0,0 +1,133 @@
...
+    try {
+      StringWriter out = new StringWriter();
+      PrintWriter writer = new PrintWriter(out);
+      AsyncPrettyPrinter mAsyncPrettyPrinter = mPrettyPrinterTestFactory.getInstance(
+          TEST_HEADER_NAME,
+          TEST_HEADER_VALUE);
+      mAsyncPrettyPrinter.printTo(writer, mInputStream);
+      assertEquals(PRETTY_PRINT_PREFIX + Arrays.toString(TEST_RESPONSE_BODY), out.toString());
+    }
+    catch (IOException e) {
```
jasta: unnecessary (also the extra line is wrong).  Instead, just declare `throws IOException`.  JUnit will auto-fail the test if this happens.

stetho/src/test/java/com/facebook/stetho/inspector/network/AsyncPrettyPrintResponseBodyTest.java
```diff
@@ -0,0 +1,133 @@
...
+      fail(e.getMessage());
+    }
+  }
+
+  @Test
+  public void testInitAsyncPrettyPrinterForResponseWithRegisteredHeader() {
+    NetworkEventReporter.InspectorResponse mockResponse = mock(
+        NetworkEventReporter.InspectorResponse.class);
```
jasta: These types of simple interfaces are usually poor candidates to use `mock`-style testing.  Instead, consider just making a `TestInspectorResponse` class which returns these dummy values.  Simpler and easier to reason about.

yamateh:
hi,mengdilin,Do your  pull request support AsyncHttpClient print?

mengdilin:
Hi Yamateh, the pull request is actually incomplete as of now (was looking for code reviews for what I already had from Rick and Josh). Asynchronous download of schema from the AbstractPrettyPrinterFactory class still needs to be implemented.

stetho/src/main/java/com/facebook/stetho/inspector/network/AbstractAsyncPrettyPrinterFactory.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+      };
+    } catch (MalformedURLException e) {
+      //cannot properly create async pretty printer without a correct url
+      return null;
+    }
+  }
+
+  @Nullable
+  protected abstract String parseHeaderValueToUri(String headerName, String headerValue);
```
jasta: This should not be abstract.  It's default implementation should be to return `headerValue`.  It also needs javadocs explaining this detail and what the function does as well.

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterFactory.java
```diff
@@ -0,0 +1,38 @@
...
+  public AsyncPrettyPrinter getInstance(String headerName, String headerValue);
+
+  /**
+   * Specifies the type of pretty printed content. Note that this method is called
+   * before the content is actually pretty printed. Stetho uses this
+   * method to make a hopeful guess of the type of prettified content
+   *
+   * @return an enum defined by ResourceType class
+   */
+  public Page.ResourceType getPrettifiedType();
```
jasta: This should be on the printer instance, not the factory per our discussion.

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterFactory.java
```diff
@@ -0,0 +1,38 @@
...
+   * @return an asynchronous pretty printer to prettify the response body
+   */
+  public AsyncPrettyPrinter getInstance(String headerName, String headerValue);
+
+  /**
+   * Specifies the type of pretty printed content. Note that this method is called
+   * before the content is actually pretty printed. Stetho uses this
+   * method to make a hopeful guess of the type of prettified content
+   *
+   * @return an enum defined by ResourceType class
```
jasta: On second thought, I actually think you should avoid leaking the `ResourceType` class since it is so hard to understand what the values actually translate to in terms of the UI.  Prefer instead adding your own enum:

```java
public enum PrettyPrinterDisplayType {
  JSON,
  HTML,
  TEXT,
}
```

stetho/src/main/java/com/facebook/stetho/inspector/network/AbstractAsyncPrettyPrinterFactory.java
```diff
@@ -0,0 +1,85 @@
...
+  public class Request implements Callable<Response> {
+    private URL url;
+
+    public Request(URL url) {
+      this.url = url;
+    }
+
+    @Override
+    public Response call() throws Exception {
+      return new Response(url.openStream());
```
jasta: Why do you defer reading the stream to the caller thread instead of having the background thread read the entire stream?  This seems like a serious mistake and oversight.  Also, you can simplify this considerably by making it a `Callable<String>`.

stetho/src/main/java/com/facebook/stetho/inspector/network/AbstractAsyncPrettyPrinterFactory.java
```diff
@@ -0,0 +1,85 @@
...
+            Response schemaResponse = Util.getUninterruptibly(response);
+            doPrint(output, payload, schemaResponse.getBody());
+          } catch (ExecutionException e) {
+            Throwable cause = e.getCause();
+            ExceptionUtil.propagateIfInstanceOf(cause, IOException.class);
+            throw ExceptionUtil.propagate(cause);
+          }
+        }
+      };
+    } catch (MalformedURLException e) {
```
jasta: Zen of try/catch management in Java, tip number 192: use isolate annoying checked exceptions into their own simple static functions to improved readability:

```java
@Nullable
private static URL parseURL(String uri) {
  try {
    return new URL(url);
  } catch (...) {
    return null;
  }
}
```

stetho/src/main/java/com/facebook/stetho/inspector/network/AbstractAsyncPrettyPrinterFactory.java
```diff
@@ -0,0 +1,85 @@
...
+      //cannot properly create async pretty printer without a correct url
+      return null;
```
jasta: This is a lame answer to the exception.  This would be very hard to debug/detect in the wild, especially if it happened spuriously.  The proper response is to bubble this information up somewhere that it would actually be noticed, for instance by still returning a pretty printer but pretty printing an error prefixing the payload:

[Failed to parse <HeaderName>: <Url>]
payloadhere...

stetho/src/main/java/com/facebook/stetho/inspector/network/AbstractAsyncPrettyPrinterFactory.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+  }
+
+  @Nullable
+  protected abstract String parseHeaderValueToUri(String headerName, String headerValue);
+
+  protected abstract void doPrint(PrintWriter output, InputStream payload, String schema)
+      throws IOException;
+
+  public class Request implements Callable<Response> {
```
jasta: Why is this and response public?  That makes no sense as it is not exported anywhere or generally useful to anyone else.

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinter.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package com.facebook.stetho.inspector.network;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.PrintWriter;
+
+/**
+ * Interface that callers need to implement in order to pretty print the payload received by Stetho
```
jasta: Specifically clarify that this is for pretty printing binary payloads.

stetho/src/main/java/com/facebook/stetho/inspector/network/AbstractAsyncPrettyPrinterFactory.java
```diff
@@ -0,0 +1,85 @@
...
+  @Override
+  public AsyncPrettyPrinter getInstance(String headerName, String headerValue) {
+    try {
+      String uri = parseHeaderValueToUri(headerName, headerValue);
+      final Future<Response> response = sExecutorService.submit(new Request(new URL(uri)));
+      return new AsyncPrettyPrinter() {
+        public void printTo(PrintWriter output, InputStream payload)
+            throws IOException {
+          try {
+            Response schemaResponse = Util.getUninterruptibly(response);
```
jasta: Actually, you do want to respond to interrupts.  See more later on...

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
```diff
@@ -83,4 +125,49 @@ public OutputStream openResponseBodyFile(String requestId, boolean base64Encode)
...
+    @Override
+    public String call() throws IOException {
+      return prettyPrintContent(mInputStream, mAsyncPrettyPrinter);
+    }
+
+    private String prettyPrintContent(InputStream in, AsyncPrettyPrinter asyncPrettyPrinter)
+        throws IOException {
+      ByteArrayOutputStream out = new ByteArrayOutputStream();
+      PrintWriter writer = new PrintWriter(out);
+      asyncPrettyPrinter.printTo(writer, in);
```
jasta: This needs to have timeout, and should use thread interrupt to attempt to unblock the caller.  A timeout can be implemented using an Android `Handler` and `sendMessageDelayed`.  It's safe to use the main thread for this since this is a debug tool:

```java
  private final Handler mPrettyPrinterTimeoutHandler = new Handler(Looper.mainLooper());

  ...

  final Thread myThread = Thread.currentThread();
  mHandler.postRunnableDelayed(() -> myThread.interrupt(), SOME_DELAY);
```

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
```diff
@@ -32,6 +45,10 @@
 
   private final Context mContext;
 
+  private Map<String, AsyncPrettyPrinter> mRequestIdMap = Collections.synchronizedMap(
```
jasta: You need to be using `final` in cases like this.  Please repeat this advice all over the diff.

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
jasta: My bad, I just noticed you have timeout logic above, but with more overhead than is needed.  A handler can do this cheaper than an executor, though I don't require that you use a handler in this case since it occurs only once per network request that is to be pretty printed.

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
```diff
@@ -83,4 +125,49 @@ public OutputStream openResponseBodyFile(String requestId, boolean base64Encode)
...
+   * the particular request id
+   *
+   * @param requestId Unique identifier for the response
+   * as per {@link NetworkEventReporter.InspectorResponse#requestId()}
+   * @param asyncPrettyPrinter Asynchronous Pretty Printer to pretty print the response body
+   */
+  public void associateAsyncPrettyPrinterWithId(
+      String requestId,
+      AsyncPrettyPrinter asyncPrettyPrinter) {
+    if (mRequestIdMap.containsKey(requestId)) {
```
jasta: This is not appropriately thread safe.  Whenever you require synchronization you must make sure all atomic operations are performed under the lock.  In this case, it would be possible for two threads to race associating the same async pretty printer with the request id and both would report that the key was not already there and then both would attempt to add it.

This is solved in one of two ways here: a) use a different locking object and synchronize on it for the entire method body or b) take advantage of the fact that `put` returns the old value if it was present and throw if it returns non-null.

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
```diff
@@ -56,19 +73,44 @@ public ResponseBodyData readFile(String requestId) throws IOException {
...
 
+  private String prettyPrintContentWithTimeOut(
+      AsyncPrettyPrinter asyncPrettyPrinter,
+      InputStream in) throws IOException {
+    AsyncPrettyPrintingCallable prettyPrintingCallable = new AsyncPrettyPrintingCallable(
+        in,
+        asyncPrettyPrinter);
+    Future<String> future = sExecutor.submit(prettyPrintingCallable);
+    try {
+      return Util.getUninterruptibly(future, 10, TimeUnit.SECONDS);
```
jasta: In general, always declare your constants as named variables.  In this case, with `private static final ... ` in the top of the class.  Likewise, below, replace your repeated use of the literal with the named constant so that there is only one place to change to keep the code correct and consistent.

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
jasta: EDIT2: you are calling interrupt as it turns out, via `future.cancel(true)`, however your actual Callable ignores the interrupt making it for naught.

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
```diff
@@ -56,19 +73,44 @@ public ResponseBodyData readFile(String requestId) throws IOException {
...
+        bodyData.data = readContentsAsUTF8(in);
+      }
       return bodyData;
     } finally {
       in.close();
     }
   }
 
-  private static String readContentsAsUTF8(InputStream in) throws IOException {
+  public static String readContentsAsUTF8(InputStream in) throws IOException {
```
jasta: This should be moved to `stetho.common` now that it is being used by more than one distinct caller.  Don't just mark private functions public because it is the least amount of typing :)

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
```diff
@@ -56,19 +73,44 @@ public ResponseBodyData readFile(String requestId) throws IOException {
       }
       ResponseBodyData bodyData = new ResponseBodyData();
       bodyData.base64Encoded = firstByte != 0;
-      bodyData.data = readContentsAsUTF8(in);
+
+      if (mRequestIdMap.containsKey(requestId)) {
```
jasta: This is also invalid synchronization.  `get` is all you need to call.  It is faster, simpler, and is thread safe.

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
```diff
@@ -56,19 +73,44 @@ public ResponseBodyData readFile(String requestId) throws IOException {
...
+        AsyncPrettyPrinter asyncPrettyPrinter = mRequestIdMap.get(requestId);
+        bodyData.data = prettyPrintContentWithTimeOut(asyncPrettyPrinter, in);
```
jasta: Hmm.  The way this blocks, even if just for 10s, is problematic.  The caller that invokes this is, I believe, the stetho network thread caller.  This means that the entire data portion of the DevTools UI will be wedged for up to 10s, possibly multiple times over.  You can fix this by asynchronously processing this event on an executor.

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
```diff
@@ -32,6 +45,10 @@
 
   private final Context mContext;
 
+  private Map<String, AsyncPrettyPrinter> mRequestIdMap = Collections.synchronizedMap(
+      new HashMap<String, AsyncPrettyPrinter>());
+  private ExecutorService sExecutor = Executors.newSingleThreadExecutor();
```
jasta: Prefer unifying all of your executors into one and documenting it in a holder class that stands on its own.  Also, I highly recommend you use parallel execution (not a single threaded executor) for better performance.

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkEventReporterImpl.java
```diff
@@ -152,10 +152,45 @@ public void responseHeadersReceived(InspectorResponse response) {
...
+
+  @Nullable
+  private static AsyncPrettyPrinter createPrettyPrinterForResponse(
+      InspectorResponse response,
+      AsyncPrettyPrinterRegistry registry,
+      Network.ResponseReceivedParams receivedParams) {
+    for (int i = 0, count = response.headerCount(); i < count; i++) {
+      AsyncPrettyPrinterFactory factory = registry.lookup(response.headerName(i));
+      if (factory != null) {
+        receivedParams.type = factory.getPrettifiedType();
```
jasta: This is ugly collusion.  Prefer not to mix unrelated concepts if you don't have to.  In this case, you can divide the method into `findAsyncPrettyPrinterFactory(...)` and then another function that calls `getInstance`.  By dividing functions into further granular logical components you can better express the logic you're after without uber functions that take large numbers ofa rguments (trust me, that's where you were headed *grin*)

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterFactory.java
mengdilin: I am not sure how I can accomplish this. AsyncPrettyPrinter is returned as an anonymous class inside the getInstance method of AbstractAsyncPrettyPrinterFactory. If we move this method to the AsyncPrettyPrinter interface, then we also need to define this method for getInstance in AbstractAsyncPrettyPrinterFactory; however, AbstractAsyncPrettyPrinterFactory knows nothing about the type of prettified content.

jasta: I don't grok any of this.  I'll explain further in `Abstract*`...

stetho/src/main/java/com/facebook/stetho/inspector/network/AbstractAsyncPrettyPrinterFactory.java
```diff
@@ -0,0 +1,85 @@
...
+public abstract class AbstractAsyncPrettyPrinterFactory implements AsyncPrettyPrinterFactory {
+
+  private static ExecutorService sExecutorService = Executors.newFixedThreadPool(1);
+
+  @Override
+  public AsyncPrettyPrinter getInstance(String headerName, String headerValue) {
+    try {
+      String uri = parseHeaderValueToUri(headerName, headerValue);
+      final Future<Response> response = sExecutorService.submit(new Request(new URL(uri)));
+      return new AsyncPrettyPrinter() {
```
jasta: This instance needs to know the type that will result, per your comments on `AsyncPrettyPrinterFactory`.  However, there are two ways you can get that:

1. Simple have the abstract factory define a new, abstract method, which gets the resource type and forward the invocation of the `AsyncPrettyPrinter` defined here to that outer class method.
2. Modify `parseHeader...` to supply more context on the matched header:

```java
MatchResult matchHeader(...) {
  return new MatchResult(
      parseUriFromValue(headerValue), 
      Page.ResourceType.XHR);
}
```

(2) is my preference because it is more general and easier to understand, all other things equal.

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterFactory.java
jasta: EDIT: I see what you mean, though I disagree.  Your concerns ignore the fact that the base interface and the simple abstract helper have to have a perfectly 1:1 matching API.  This isn't true and isn't even desirable :)

stetho/src/main/java/com/facebook/stetho/inspector/network/AbstractAsyncPrettyPrinterFactory.java
```diff
@@ -0,0 +1,85 @@
...
+import com.facebook.stetho.inspector.network.AsyncPrettyPrinter;
+import com.facebook.stetho.inspector.network.AsyncPrettyPrinterFactory;
+
+import javax.annotation.Nullable;
+
+/**
+ * Abstract class for pretty printer factory that asynchronously downloads schema needed for
+ * pretty printing the payload
+ */
+public abstract class AbstractAsyncPrettyPrinterFactory implements AsyncPrettyPrinterFactory {
```
jasta: This really should be renamed to something more deliberate like: `DownloadingAsyncPrettyPrinterFactory`.

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinter.java
```diff
@@ -0,0 +1,39 @@
...
+  public void printTo(PrintWriter output, InputStream payload) throws IOException;
+
+  /**
+   * Specifies the type of pretty printed content. Note that this method is called
+   * before the content is actually pretty printed. Stetho uses this
+   * method to make a hopeful guess of the type of prettified content
+   *
+   * @return an enum defined by PrettyPrinterDisplayType class
+   */
+  public DownloadingAsyncPrettyPrinterFactory.PrettyPrinterDisplayType getPrettifiedType();
```
jasta: This represents a layering problem.  `AsyncPrettyPrinter` and `AsyncPrettyPrinterFactory` exist at the most abstract layer and cannot refer to more concrete layers directly.  Prefer just using a new top-level class, `PrettyPrinterDisplayType`, to fix this.

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterExecutorHolder.java
```diff
@@ -0,0 +1,20 @@
...
+
+package com.facebook.stetho.inspector.network;
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+/**
+ * A holder class for the executor service used for pretty printing related tasks
+ */
+public class AsyncPrettyPrinterExecutorHolder {
```
jasta: Can you make this package private?

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterExecutorHolder.java
```diff
@@ -0,0 +1,20 @@
...
+  public final static ExecutorService sExecutorService = Executors.newCachedThreadPool();
```
jasta: style: `static final`

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterRegistry.java
```diff
@@ -0,0 +1,34 @@
...
+
+package com.facebook.stetho.inspector.network;
+
+import javax.annotation.Nullable;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public class AsyncPrettyPrinterRegistry {
```
jasta: nit: mark `@ThreadSafe`

stetho/src/main/java/com/facebook/stetho/inspector/network/DownloadingAsyncPrettyPrinterFactory.java
```diff
@@ -0,0 +1,170 @@
...
+  private static URL parseURL(String uri) {
+    try {
+      return new URL(uri);
+    } catch (MalformedURLException e) {
+      return null;
+    }
+  }
+
+  private static void doErrorPrint(PrintWriter output, InputStream payload, String errorMessage)
+    throws IOException {
```
jasta: style: 4 spaces indent for line wrapping

stetho/src/main/java/com/facebook/stetho/inspector/network/DownloadingAsyncPrettyPrinterFactory.java
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    @Override @Nullable
+    public String call() {
+      try {
+        InputStream urlStream = url.openStream();
+        String result = Util.readAsUTF8(urlStream);
+        urlStream.close();
+        return result;
+      }
```
jasta: style: catch on the same line as the closing `}`

stetho/src/main/java/com/facebook/stetho/inspector/network/DownloadingAsyncPrettyPrinterFactory.java
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  private class Request implements Callable<String> {
+    private URL url;
+
+    public Request(URL url) {
+      this.url = url;
+    }
+
+    @Override @Nullable
```
jasta: style: these go on separate lines

stetho/src/main/java/com/facebook/stetho/inspector/network/DownloadingAsyncPrettyPrinterFactory.java
```diff
@@ -0,0 +1,170 @@
...
+    public String getSchemaUri() {
+      return mSchemaUri;
+    }
+
+    public PrettyPrinterDisplayType getDisplayType() {
+      return mDisplayType;
+    }
+  }
+
+  private class Request implements Callable<String> {
```
jasta: Can this be a static class?  Classes and functions without side effects should be marked as such to help the reader easily understand their scope

stetho/src/main/java/com/facebook/stetho/inspector/network/DownloadingAsyncPrettyPrinterFactory.java
```diff
@@ -0,0 +1,170 @@
...
+      mDisplayType = displayType;
+    }
+
+    public String getDisplayType() {
+      return mDisplayType;
+    }
+  }
+
+  protected class MatchResult {
+    private String mSchemaUri;
```
jasta: `final`

stetho/src/main/java/com/facebook/stetho/inspector/network/DownloadingAsyncPrettyPrinterFactory.java
```diff
@@ -0,0 +1,170 @@
...
+      @Override
+      public PrettyPrinterDisplayType getPrettifiedType() {
+        return PrettyPrinterDisplayType.TEXT;
+      }
+    };
+  }
+
+
+  public enum PrettyPrinterDisplayType {
+    JSON("Json"),
```
jasta: Why do you have these string values?  What do they mean?

stetho/src/main/java/com/facebook/stetho/inspector/network/DownloadingAsyncPrettyPrinterFactory.java
```diff
@@ -0,0 +1,170 @@
...
+    @Override @Nullable
+    public String call() {
+      try {
+        InputStream urlStream = url.openStream();
+        String result = Util.readAsUTF8(urlStream);
+        urlStream.close();
+        return result;
+      }
+      catch (IOException e) {
+        return null;
```
jasta: Your code above pretends to handle this case, and yet you just return null instead, sacrificing all of the error information you previously had.  This is the kind of thing that makes engineers tear their hair out during multi-hour debug sessions trying to find this line.

Instead, declare that you throw an `IOException` and let it bubble up at least to the anonymous class in `getInstance`.  FWIW, I think the better design is to let it bubble all the way up to `Network#getResponseBody`  where it would actually yield an error to the chrome devtools, allowing it to potentially be retried if the user clicks again, but at this point that should be looked at as a follow-up to your base work.

stetho/src/main/java/com/facebook/stetho/inspector/network/DownloadingAsyncPrettyPrinterFactory.java
```diff
@@ -0,0 +1,170 @@
...
+            throws IOException {
+          try {
+            String schema = response.get();
+            if (schema != null) {
+              doPrint(output, payload, schema);
+            } else {
+              doErrorPrint(
+                  output,
+                  payload,
+                  "http get request for schema url fails to return a valid response");
```
jasta: Per my comment below, what sucks about this line is that you don't know what actually happened and can't give the caller any useful information.

stetho/src/main/java/com/facebook/stetho/inspector/network/DownloadingAsyncPrettyPrinterFactory.java
```diff
@@ -0,0 +1,170 @@
...
+                output,
+                payload,
+                "Error while downloading schema for pretty printing: " + e.getMessage());
+          } catch (ExecutionException e) {
+            Throwable cause = e.getCause();
+            ExceptionUtil.propagateIfInstanceOf(cause, IOException.class);
+            throw ExceptionUtil.propagate(cause);
+          }
+        }
+        public DownloadingAsyncPrettyPrinterFactory.PrettyPrinterDisplayType getPrettifiedType() {
```
jasta: style nit: blank line between methods...

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkEventReporterImpl.java
```diff
@@ -152,10 +152,52 @@ public void responseHeadersReceived(InspectorResponse response) {
...
+    AsyncPrettyPrinter asyncPrettyPrinter = createPrettyPrinterForResponse(response, registry);
+    if (asyncPrettyPrinter != null) {
+      peerManager.getResponseBodyFileManager().associateAsyncPrettyPrinterWithId(
+          response.requestId(),
+          asyncPrettyPrinter);
+    }
+     return asyncPrettyPrinter;
+  }
+
+  private void modifyReceivedParamsForAsyncPrettyPrinter(
```
jasta: This function has some readability challenges.  First, it claims to modify parameters but it really only conditionally does that and the condition isn't clear at a high level.  Second, there's no reason to allow the original assignment to proceed since you know that the pretty printer can override it at the very time that it was previously determined.  You can express this logic much more concisely as:

```java
private static Page.ResourceType determineResourceType(...) {
  if (asyncPrettyPrinter != null) {
    return ...
  } else {
    return ... getResourceTypeHelper().determineResourceType(...);
  }
}
```

stetho/src/main/java/com/facebook/stetho/inspector/network/ResourceTypeHelper.java
```diff
@@ -24,9 +24,18 @@ public ResourceTypeHelper() {
...
     mMimeMatcher.addRule("text/javascript", Page.ResourceType.XHR);
     mMimeMatcher.addRule("application/json", Page.ResourceType.XHR);
+    mMimeMatcher.addRule(
+        DownloadingAsyncPrettyPrinterFactory.PrettyPrinterDisplayType.JSON.getDisplayType(),
+        Page.ResourceType.XHR);
 
     // Everything else gets a lame, unformatted blob.
     mMimeMatcher.addRule("text/*", Page.ResourceType.DOCUMENT);
+    mMimeMatcher.addRule(
+        DownloadingAsyncPrettyPrinterFactory.PrettyPrinterDisplayType.HTML.getDisplayType(),
```
jasta: _VERY_ bad collusion between unrelated types!  You really must be careful about how you mix abstractions so that folks aren't surprised by two unrelated layers suddenly mixing together. 

In this case, you want an adapter that simply knows how to convert `PrettyPrinterDisplayType` values to `Page.ResourceType`.  `ResourceTypeHelper` was explicitly designed to convert MIME types to `Page.ResourceType`.  I'd recommend putting this adapter in `PrettyPrinterDisplayType` itself, with a `getResourceType` method instead of your `getDisplayType` method today.

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
```diff
@@ -29,9 +40,15 @@
 public class ResponseBodyFileManager {
   private static final String TAG = "ResponseBodyFileManager";
   private static final String FILENAME_PREFIX = "network-response-body-";
+  private static final int PRETTY_PRINT_TIMEOUT = 10;
```
jasta: Times are almost always better expressed in Java using milliseconds, and you should also clarify the unit in the variable to avoid confusio (`PRETTY_PRINT_TIMEOUT_MS`)

stetho/src/main/java/com/facebook/stetho/inspector/network/DownloadingAsyncPrettyPrinterFactory.java
```diff
@@ -0,0 +1,154 @@
...
+            String schema = response.get();
+            doPrint(output, payload, schema);
+          } catch (InterruptedException e) {
+            doErrorPrint(
+                output,
+                payload,
+                "Error while downloading schema for pretty printing: " + e.getMessage());
+          } catch (ExecutionException e) {
+            Throwable cause = e.getCause();
+            if (FileNotFoundException.class.isInstance(cause)) {
```
jasta: This is imprecise and problematic.  Let's discuss live.

stetho/src/main/java/com/facebook/stetho/inspector/network/DownloadingAsyncPrettyPrinterFactory.java
```diff
@@ -0,0 +1,154 @@
...
+
+  private final static ExecutorService sExecutorService =
+      AsyncPrettyPrinterExecutorHolder.sExecutorService;
+
+  @Override
+  public AsyncPrettyPrinter getInstance(final String headerName, final String headerValue) {
+
+    final MatchResult result = matchAndParseHeader(headerName, headerValue);
+    if (result == null) {
+      return getErrorAsyncPrettyPrinter(headerName, headerValue);
```
jasta: It's not an error if it is a false match.  This should just return null.

stetho/src/main/java/com/facebook/stetho/inspector/network/PrettyPrinterDisplayType.java
```diff
@@ -0,0 +1,31 @@
...
+  HTML(Page.ResourceType.DOCUMENT),
+  TEXT(Page.ResourceType.DOCUMENT);
+
+  private final Page.ResourceType mResourceType;
+
+  private PrettyPrinterDisplayType(Page.ResourceType resourceType) {
+    mResourceType = resourceType;
+  }
+
+  /** Converts PrettyPrinterDisplayType values to the appropriate
```
jasta: nit: newline.  Also use `{@link}` to make it easier to navigate in intellij.

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
```diff
@@ -56,17 +73,38 @@ public ResponseBodyData readFile(String requestId) throws IOException {
       }
       ResponseBodyData bodyData = new ResponseBodyData();
       bodyData.base64Encoded = firstByte != 0;
-      bodyData.data = readContentsAsUTF8(in);
+
+      AsyncPrettyPrinter asyncPrettyPrinter = mRequestIdMap.get(requestId);
+      if (asyncPrettyPrinter != null) {
+        bodyData.data = prettyPrintContentWithTimeOut(asyncPrettyPrinter, in);
```
jasta: Prefer that you reference the task you created in code so that if someone sees this they can get more context.

jasta:
Looks good overall, almost there

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
mengdilin: Blocking code that can be problematic as explained by #243

stetho/src/test/java/com/facebook/stetho/inspector/network/AsyncPrettyPrintResponseBodyTest.java
```diff
@@ -0,0 +1,243 @@
...
+      ByteArrayOutputStream out = new ByteArrayOutputStream();
+      Util.copy(payload, out, new byte[1024]);
+      String prettifiedContent = PRETTY_PRINT_PREFIX + Arrays.toString(out.toByteArray());
+      output.write(prettifiedContent);
+      output.close();
+    }
+
+    @Override @Nullable
+    protected MatchResult matchAndParseHeader(String headerName, String headerValue) {
+      return new MatchResult("https://www.facebook.com", PrettyPrinterDisplayType.TEXT);
```
longinoa: Your test should specify what `headerName` is before returning the `MatchResult` This way you can verify what happens if there are no matches.

stetho/src/test/java/com/facebook/stetho/inspector/network/AsyncPrettyPrintResponseBodyTest.java
```diff
@@ -0,0 +1,248 @@
...
+    mAsyncPrettyPrinter.printTo(writer, mInputStream);
+    assertEquals(PRETTY_PRINT_PREFIX + Arrays.toString(TEST_RESPONSE_BODY), out.toString());
+  }
+
+  @Test
+  public void testInitAsyncPrettyPrinterForResponseWithRegisteredHeader() {
+    ArrayList<String> headerNames = new ArrayList<String>();
+    ArrayList<String> headerValues = new ArrayList<String>();
+
+    headerNames.add("unregistered header name 1");
```
longinoa: Nit: These should be constants

longinoa:
This looks good go ahead and rebase and squash down, then ill merge.

PR #225, opened 2015-07-13
Added the ability to search the DOM tree

### LINKED ISSUE #179: Search View Hierarchy by Class Name
First off, great work guys! Is the search function I mentioned above supposed to be working? it it isn't for me. Android's monitor does support it, but the whole tool is flimsy at best. This is way more stable.

ichub:
The search searches node names, property names, and property values, mirroring the search in chrome inspect on an html webpage.
searchworking
Closes #179

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
@@ -44,9 +44,13 @@ import javax.annotation.Nullable;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Queue;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.regex.Pattern;
 
 public class DOM implements ChromeDevtoolsDomain {
   private final ChromePeerManager mPeerManager;
@@ -65,6 +69,9 @@ public class DOM implements ChromeDevtoolsDomain {
   private ChildNodeRemovedEvent mCachedChildNodeRemovedEvent;
   private ChildNodeInsertedEvent mCachedChildNodeInsertedEvent;
 
+  private final Map<String, List<Integer>> mSearchResults;
+  private final AtomicInteger mResultCounter;
+
   public DOM(DOMProvider.Factory providerFactory) {
     mDOMProviderFactory = Util.throwIfNull(providerFactory);
 
@@ -73,6 +80,11 @@ public class DOM implements ChromeDevtoolsDomain {
 
     mObjectMapper = new ObjectMapper();
     mObjectIdMapper = new DOMObjectIdMapper();
+
+    mSearchResults = Collections.synchronizedMap(
+      new HashMap<String, List<Integer>>());
+
+    mResultCounter = new AtomicInteger(0);
   }
 
   @ChromeDevtoolsMethod
@@ -206,6 +218,104 @@ public class DOM implements ChromeDevtoolsDomain {
         mDOMProvider.setInspectModeEnabled(request.enabled);
       }
     });
+  }
+
+  @ChromeDevtoolsMethod
+  public PerformSearchResponse performSearch(JsonRpcPeer peer, final JSONObject params) {
+    final PerformSearchRequest request = mObjectMapper.convertValue(
+        params,
+        PerformSearchRequest.class);
+
+    final Pattern queryPattern = Pattern.compile(Pattern.quote(request.query), Pattern.CASE_INSENSITIVE);
+
+    ArrayList<Integer> results = mDOMProvider.postAndWait(new UncheckedCallable<ArrayList<Integer>>() {
+      @Override
+      public ArrayList<Integer> call() {
+        final Object rootElement = mShadowDOM.getRootElement();
+
+        final ArrayList<Integer> results = new ArrayList<>();
+
+        findMatches(queryPattern, rootElement, results);
+
+        return results;
+      }
+    });
+
+    // Each search action has a unique ID so that
+    // it can be queried later.
+    final String id = String.valueOf(mResultCounter.getAndIncrement());
+
+    mSearchResults.put(id, results);
+
+    final PerformSearchResponse response = new PerformSearchResponse();
+    response.searchId = id;
+    response.resultCount = results.size();
+
+    return response;
+  }
+
+  private void findMatches(Pattern queryPattern, Object element, List<Integer> resultIds) {
+    final ElementInfo info = mShadowDOM.getElementInfo(element);
+
+    for (int i = 0, size = info.children.size(); i < size; i++) {
+      final Object childElement = info.children.get(i);
+
+      if (doesElementMatch(queryPattern, childElement)) {
+        resultIds.add(mObjectIdMapper.getIdForObject(childElement));
+      }
+
+      findMatches(queryPattern, childElement, resultIds);
+    }
+  }
+
+  private boolean doesElementMatch(final Pattern queryPattern, final Object element) {
+      final NodeDescriptor descriptor = mDOMProvider.getNodeDescriptor(element);
+
+      final AttributeListAccumulator accumulator = getCachedAttributeListAccumulator();
+
+      try {
+        descriptor.getAttributes(element, accumulator);
+
+        for (int i = 0, size = accumulator.size(); i < size; i++) {
+          if (queryPattern.matcher(accumulator.get(i)).find()) {
+            return true;
+          }
+        }
+      } finally {
+        returnCachedAttributeListAccumulator(accumulator);
+      }
+
+      return queryPattern.matcher(descriptor.getNodeName(element)).find();
+  }
+
+  @ChromeDevtoolsMethod
+  public GetSearchResultsResponse getSearchResults(JsonRpcPeer peer, JSONObject params) {
+    final GetSearchResultsRequest request = mObjectMapper.convertValue(
+        params,
+        GetSearchResultsRequest.class);
+
+    final List<Integer> results = mSearchResults.get(request.searchId);
+
+    if (results == null) {
+      LogUtil.w("\"" + request.searchId + "\" is not a valid reference to a search result");
+      return null;
+    }
+
+    final List<Integer> resultsRange = results.subList(request.fromIndex, request.toIndex);
+
+    final GetSearchResultsResponse response = new GetSearchResultsResponse();
+    response.nodeIds = resultsRange;
+
+    return response;
+  }
+
+  @ChromeDevtoolsMethod
+  public void discardSearchResults(JsonRpcPeer peer, JSONObject params) {
+    final DiscardSearchResultsRequest request = mObjectMapper.convertValue(
+      params,
+      DiscardSearchResultsRequest.class);
+
+    mSearchResults.remove(request.searchId);
   }
 
   private void updateTree() {
@@ -415,6 +525,7 @@ public class DOM implements ChromeDevtoolsDomain {
           mShadowDOM = null;
           mObjectIdMapper.clear();
           mDOMProvider.dispose();
+          mSearchResults.clear();
         }
       });
 
@@ -814,5 +925,42 @@ public class DOM implements ChromeDevtoolsDomain {
   private static class ResolveNodeResponse implements JsonRpcResult {
     @JsonProperty(required = true)
     public Runtime.RemoteObject object;
+  }
+
+  private static class PerformSearchRequest {
+    @JsonProperty(required = true)
+    public String query;
+
+    @JsonProperty
+    public Boolean includeUserAgentShadowDOM;
+  }
+
+  private static class PerformSearchResponse implements JsonRpcResult {
+    @JsonProperty(required = true)
+    public String searchId;
+
+    @JsonProperty(required = true)
+    public int resultCount;
+  }
+
+  private static class GetSearchResultsRequest {
+    @JsonProperty(required = true)
+    public String searchId;
+
+    @JsonProperty(required = true)
+    public int fromIndex;
+
+    @JsonProperty(required = true)
+    public int toIndex;
+  }
+
+  private static class GetSearchResultsResponse implements JsonRpcResult {
+    @JsonProperty(required = true)
+    public List<Integer> nodeIds;
+  }
+
+  private static class DiscardSearchResultsRequest {
+    @JsonProperty(required = true)
+    public String searchId;
   }
 }


stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -208,6 +214,96 @@ public void run() {
...
+    mSearchResults.put(id, results);
+
+    final PerformSearchResponse response = new PerformSearchResponse();
+    response.searchId = id;
+    response.resultCount = results.size();
+
+    return response;
+  }
+
+  private void FindMatches(String query, Object root, ArrayList<Integer> resultIds) {
```
rickbrew: method names should start with lowercase, e.g. `findMatches`. Saw this with `DoesElementMatch`, too.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -208,6 +214,96 @@ public void run() {
...
+    final ElementInfo info = mShadowDOM.getElementInfo(root);
+
+    for (int i = 0; i < info.children.size(); i++) {
```
rickbrew: recommend using a for-loop pattern like this:
`for (int i = 0, size = info.children.size(); i < size; i++) {`
(yeah, optimizations like this actually matter on Android!)

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -208,6 +214,96 @@ public void run() {
...
+      if (DoesElementMatch(query, element)) {
+        resultIds.add(mObjectIdMapper.getIdForObject(element));
+      }
+
+      FindMatches(query, element, resultIds);
+    }
+  }
+
+  private boolean DoesElementMatch(String query, Object element) {
+    final String queryParam = query;
```
rickbrew: you can avoid copying into a local variable just for the sake of adding `final` ... by adding `final` to the parameter, e.g.
```
private boolean doesElementMatch(final String query, final Object element) {
```

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -208,6 +214,96 @@ public void run() {
...
+      @Override
+      public Boolean call() {
+        final NodeDescriptor descriptor = mDOMProvider.getNodeDescriptor(elemParam);
+
+        final AttributeListAccumulator accumulator = getCachedAttributeListAccumulator();
+        descriptor.getAttributes(elemParam, accumulator);
+
+        for (int i = 0; i < accumulator.size(); i++) {
+          if (accumulator.get(i).toLowerCase().contains(queryParam)) {
+            return true;
```
rickbrew: when you return here you're not returning the accumulator. this is easy to fix up via try/finally though:
```
final AttributeListAccumulator accumulator = getCachedAttributeListAccumulator();
try {
  ...
} finally {
  returnCachedAttributeListAccumulator(accumulator);
}
```

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -208,6 +214,96 @@ public void run() {
...
 
+  @ChromeDevtoolsMethod
+  public PerformSearchResponse performSearch(JsonRpcPeer peer, JSONObject params) {
+    final PerformSearchRequest request = mObjectMapper.convertValue(
+        params,
+        PerformSearchRequest.class);
+
+    // Searches are case insensitive so convert
+    // the query to lower case for simpler comparisons.
+    request.query = request.query.toLowerCase();
```
rickbrew: prefer not overwriting the query field in the request. instead, bounce to a local variable and use that, e.g.
```
final String requestQuery = request.query.toLowerCase();
```
(fuller explanation available upon request!)

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -208,6 +214,96 @@ public void run() {
...
+
+      FindMatches(query, element, resultIds);
+    }
+  }
+
+  private boolean DoesElementMatch(String query, Object element) {
+    final String queryParam = query;
+    final Object elemParam = element;
+
+    return mDOMProvider.postAndWait(new UncheckedCallable<Boolean>() {
```
rickbrew: I recommend putting the entire search into a `postAndWait` callback, as there's some decent overhead to this thing. To do this, remove the post-and-waiting from `doesElementMatch` and hoist it into the calling method, `findMatches` (that whole method should be inside a post-and-wait callback, from the looks of it). That way you'll have 1 post-and-wait for the whole search instead of 1 per element.

rickbrew:
Okay there's my first round of feedback. Mostly just have some nits and thread safety concerns that should be really easy to fix up. Looks pretty awesome so far, and there's much less code than I was expecting (which is a really good thing).

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
ichub: Should all request params classes be marked as immutable then?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -65,6 +67,8 @@
   private ChildNodeRemovedEvent mCachedChildNodeRemovedEvent;
   private ChildNodeInsertedEvent mCachedChildNodeInsertedEvent;
 
+  private HashMap<String, List<Integer>> mSearchResults;
```
rickbrew: because this is read and written from different threads, we need to synchronize it, e.g.
```
private final Map<String, List<Integer>> mSearchResults = Collections.synchronizedMap(
    new HashMap<String, List<Integer>>());
```
And by assigning to it here instead of in the constructor we can tag it with 'final'.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -208,6 +214,97 @@ public void run() {
...
+      return descriptor.getNodeName(element).toLowerCase().contains(query);
+  }
+
+  @ChromeDevtoolsMethod
+  public GetSearchResultsResponse getSearchResults(JsonRpcPeer peer, JSONObject params) {
+    final GetSearchResultsRequest request = mObjectMapper.convertValue(
+        params,
+        GetSearchResultsRequest.class);
+
+    final List<Integer> results = mSearchResults.get(request.searchId).subList(request.fromIndex, request.toIndex);
```
rickbrew: We shouldn't trust that the `searchId` from the client is valid. If it's not then `mSearchResults.get()` will return null and we'll crash with an NPE. So let's make sure to check the return value from `get()` before using it. If it returns null then ... I guess we just return null? You can test that by just returning null from this method and making sure Chrome doesn't crash or do something weird (I doubt it will, but always good to double-check).

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -208,6 +214,97 @@ public void run() {
...
+        final PerformSearchResponse response = new PerformSearchResponse();
+        response.searchId = id;
+        response.resultCount = results.size();
+
+        return response;
+      }
+    });
+  }
+
+  private void findMatches(String query, Object root, ArrayList<Integer> resultIds) {
```
rickbrew: Prefer calling it 'element' instead of 'root'. We need to ensure we include the word 'element' because we don't really have another way to express in Java that this needs to be "an element" (and element means something very specific in this corner of the code) (and this also provides consistency with our other code that works with "elements"). Since this is a recursive method, calling it 'root' might be a bit misleading too. The `element` in the loop below can be called `childElement`.

Also prefer having the `resultIds` be of type `List<Integer>` since it's the convention in Java to strip away concrete types in favor of interfaces, especially across boundaries.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -208,6 +214,97 @@ public void run() {
...
+    }
+  }
+
+  private boolean doesElementMatch(final String query, final Object element) {
+      final NodeDescriptor descriptor = mDOMProvider.getNodeDescriptor(element);
+
+      final AttributeListAccumulator accumulator = getCachedAttributeListAccumulator();
+
+      try {
+        descriptor.getAttributes(element, accumulator);
```
rickbrew: We'll need to update this code later, re: #228 that I just opened. Just a heads up :)

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
rickbrew: Also, we can probably go ahead and `clear()` this in `onLastPeerUnregistered()`.

rickbrew:
Okay second round of feedback complete

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -208,6 +216,103 @@ public void run() {
...
+    final List<Integer> resultsRange = results.subList(request.fromIndex, request.toIndex);
+
+    final GetSearchResultsResponse response = new GetSearchResultsResponse();
+    response.nodeIds = resultsRange;
+
+    return response;
+  }
+
+  @ChromeDevtoolsMethod
+  public void discardSearchResults(JsonRpcPeer peer, JSONObject params) {
```
jasta: When is this method called?  I'm curious how much we can depend on this for our garbage collection.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -65,6 +68,8 @@
   private ChildNodeRemovedEvent mCachedChildNodeRemovedEvent;
   private ChildNodeInsertedEvent mCachedChildNodeInsertedEvent;
 
+  private final Map<String, List<Integer>> mSearchResults;
```
jasta: Strongly prefer `Map<String, ManagedIntArray>` for memory efficiency.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -73,6 +78,9 @@ public DOM(DOMProvider.Factory providerFactory) {
 
     mObjectMapper = new ObjectMapper();
     mObjectIdMapper = new DOMObjectIdMapper();
+
+    mSearchResults = Collections.synchronizedMap(
+      new HashMap<String, List<Integer>>());
```
jasta: Pretty sure you can just write `new HashMap<>();`

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -208,6 +216,103 @@ public void run() {
...
+    return mDOMProvider.postAndWait(new UncheckedCallable<PerformSearchResponse>() {
+      @Override
+      public PerformSearchResponse call() {
+        final PerformSearchRequest request = mObjectMapper.convertValue(
+            params,
+            PerformSearchRequest.class);
+
+        // Searches are case insensitive so convert
+        // the query to lower case for simpler comparisons.
+        final String lowerCaseQuery = request.query.toLowerCase();
```
jasta: Fun fact about localization: `a.toLowerCase().equals(b.toLowerCase())` is not the same as `a.equalsIgnoreCase(b)`.  I believe you are making this mistake here.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -208,6 +216,103 @@ public void run() {
...
+
+        final Object root = mShadowDOM.getRootElement();
+
+        // Each search action has a unique ID so that
+        // it can be queried later.
+        final String id = UUID.randomUUID().toString();
```
jasta: Strongly prefer monotonically increasing identifiers.  This isn't a security implication and therefore the extra cost and complexity of a UUID is not warranted.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -208,6 +216,103 @@ public void run() {
...
+
+        final ArrayList<Integer> results = new ArrayList<>();
+
+        findMatches(lowerCaseQuery, root, results);
+
+        mSearchResults.put(id, results);
```
jasta: This is a bit awkward to be doing on the main thread.  Strongly prefer logically dividing the work that has to be on the main thread with the rest:

```java
  List<> results = postAndWait(() -> {
    findMatches(...);
    return results;
  });

  mSearchResults.put(id, results);
  PerformSearchResponse response = ...;
  return response;
```java

This is not only a perf issue in general, it also helps clarify your intention for moving to the main thread for some of your work.

File: stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java [hunk 17]
```diff
@@ -208,6 +216,103 @@ public void run() {
...
+
+  @ChromeDevtoolsMethod
+  public GetSearchResultsResponse getSearchResults(JsonRpcPeer peer, JSONObject params) {
+    final GetSearchResultsRequest request = mObjectMapper.convertValue(
+        params,
+        GetSearchResultsRequest.class);
+
+    final List<Integer> results = mSearchResults.get(request.searchId);
+
+    if (results == null) {
```
jasta: At least use `LogUtil.w` to indicate this error.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
jasta: As an aside, you can pick up an interesting perf win by simply addressing an array instead of using a map here:

  synchronized (...) {
    String id = String.valueOf(mSearchResults.size());
    mSearchResults.add(results);
  }

Then when the request id comes back in you'd simply do:

  int index = Integer.parseInt(requestId);
  synchronized (...) {
    ManagedIntArray results = mSearchResults.get(index);
  }

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
ichub: It's called either when you hide the search box by pressing escape, or when you search from something else (eg. add or remove a character from the existing query).

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
ichub: Rick specified that it should be this way. Are you sure? ref: https://github.com/facebook/stetho/pull/225#discussion_r34513807

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
ichub: Doesn't seem like ManagedIntArray is used anywhere else in the project. Is it part of the Java standard library? Is [this](http://www.blazegraph.com/docs/api/com/bigdata/rdf/graph/impl/util/ManagedIntArray.html) the class?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
ichub: Are you sure? Rick commented on this too: https://github.com/facebook/stetho/pull/225#discussion_r34513807

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
ichub: Also, when passing back a collection from Stetho back to chrome, the Json serializer expects a `List`, so using an array and then converting back to a list in the end seems redundant.

rickbrew: I think we're fine to use `List<Integer>` in this case, although it's a little circumstantial. We're pulling the `Integer`s from `mObjectIdMapper.getIdForObject()` which is already storing them as `Integer` in its mObjectToIdMap ... in other words, we're not adding any extra overhead.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
jasta: Ok, fair enough, looks like your usage of Map and this implementation is the right strategy then.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
jasta: Ack'd, I think this is fine as well.

jasta:
lgtm, please squash and rebase, and make sure the commit message matches the PR.

PR #229, opened 2015-07-14
Remove redundant modifiers from interfaces

VEINHORN:


jasta:
Hmm, I actually prefer the redundant keywords :\  Helps me to not assume that things are package private, non-static, etc residing inside a regular class.

felipecsl:
i think it makes sense to remove them since it's just redundant information/noise

jasta:
@felipecsl alright, you convinced me :)

felipecsl:
Whoohoo thanks!!!

PR #234, opened 2015-07-21
Provide a better customization API for Stetho.initialize

### LINKED ISSUE #195: Make it easier to change the DatabaseFilesProvider
In 1.1.1, it's very hard to use a different DatabaseFilesProvider than the DefaultDFP, because the instance is private final and it's created at the end of a long, non-configurable chain.

An easier method in the builder (or after build) of changing it would be welcome.

The purpose of changing it is to support applications that do not store their databases in the default location (/data/data/app_id/databases).

jasta:
Now that many more configuration options are becoming available it's
important to provide a more streamlined API for overriding plugins or
adding new ones, especially for dumpapp.

Closes #195

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,11 +52,7 @@ which occurs in your `Application` class:
 public class MyApplication extends Application {
   public void onCreate() {
     super.onCreate();
-    Stetho.initialize(
-        Stetho.newInitializerBuilder(this)
-            .enableDumpapp(Stetho.defaultDumperPluginsProvider(this))
-            .enableWebKitInspector(Stetho.defaultInspectorModulesProvider(this))
-            .build());
+    Stetho.initializeWithDefaults(this);
   }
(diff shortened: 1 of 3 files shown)


jasta:
@rightparen I actually took a look at doing it using a `Map` per your out-of-band suggestion but I honestly think this approach is still better.  It's more concise and doesn't require callers to repeat the name of the plugin in two places (one in registration and one in the plugin itself).  Technically that can be modified so the plugin no longer self identifies but since this is not ABI compatible it cannot be done quickly.

Also I noticed that the flexibility desired by your approach is still possibly by simply manipulating this design and saying `new DefaultDumperPluginsBuilder(...).build()`, yielding a list which can be trivially converted to a map yourself.  Few callers would need this, however, but it is important to observe that it is still _possible_.

PR #235, opened 2015-07-22
Load External databases if Present

Aexyn:
Load External database If present

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/database/DefaultDatabaseFilesProvider.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/database/DefaultDatabaseFilesProvider.java
@@ -10,9 +10,11 @@
 package com.facebook.stetho.inspector.database;
 
 import android.content.Context;
+import android.os.Environment;
+
 import java.io.File;
+import java.io.FilenameFilter;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 
 /**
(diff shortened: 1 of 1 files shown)


stetho/src/main/java/com/facebook/stetho/inspector/database/DefaultDatabaseFilesProvider.java
```diff
@@ -32,6 +34,45 @@ public DefaultDatabaseFilesProvider(Context context) {
...
+        databaseFiles.add(new File(absoluteFile,db));
+      }
+    }
     return databaseFiles;
   }
+
+  /**
+   * @return true if External Storage Present
+   */
+  private static boolean isExternalAvailable() {
```
jasta: This doesn't strictly seem necessary to me.  You can just test for the file path's existence, no?  If it is not there, then `getExternalFilesDir()` would yield a path that didn't exist right?

stetho/src/main/java/com/facebook/stetho/inspector/database/DefaultDatabaseFilesProvider.java
```diff
@@ -32,6 +34,45 @@ public DefaultDatabaseFilesProvider(Context context) {
...
+      public boolean accept(File dir, String filename) {
+        return filename.endsWith(".sql")
+            || filename.endsWith(".sqlite")
+            || filename.endsWith(".db");
+      }
+    };
+
+    //Add External Databases if Present
+    if(isExternalAvailable()){
+      final File absoluteFile = mContext.getExternalFilesDir(null).getAbsoluteFile();
```
jasta: I'm uncomfortable with this being the default.  These files have dubious security implications and are not necessarily accessible without special permissions pre-KitKat.  Perhaps we can make this an option to the DefaultDatabaseFilesProvider and use our extensibility API (which is being improved BTW) to let callers tweak how they want this to be brought in?

jasta:
I believe Stetho 1.2.0 addressed this with a superior API that allows you to declare where your database files live.  See `DefaultInspectorModulesBuilder#databaseFiles`.

PR #237, opened 2015-07-28
Split DOM into two classes: Document and DOM

ichub:
DOM now handles only the communication between Chrome and Stetho, whereas Document is responsible for accessing, modifying, and maintaining information about elements on the Android side. The motivation for this change is that DOM is not the only class that needs access to information about elements - CSS also needs this information. By splitting these classes up it is possible for both DOM and CSS to be able to get and set element data.


stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
```diff
@@ -0,0 +1,473 @@
...
+  private ShadowDOM mShadowDOM;
+  private ObjectIdMapper mObjectIdMapper;
+  private UpdateListener mUpdateListener;
+
+  private AttributeListAccumulator mCachedAttributeAccumulator;
+  private ChildEventingList mCachedChildEventingList;
+  private ArrayListAccumulator<Object> mCachedChildrenAccumulator;
+
+  public Document(DOMProvider.Factory factory) {
+    this.mFactory = factory;
```
jasta: style: drop `this.`

stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
```diff
@@ -0,0 +1,473 @@
...
+    this.mObjectIdMapper = new DOMObjectIdMapper();
+  }
+
+  // dom provider should be created only once, and if multiple
+  // domains use the class Document, errors could arise if this
+  // was merely a `createDomProvider` method
+  public void createDomProviderIfNoneExists() {
```
jasta: Uhm, what is the `IfNoneExists` part of this?  I don't see any such enforcement.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -21,7 +22,10 @@
 import org.json.JSONObject;
 
 public class CSS implements ChromeDevtoolsDomain {
-  public CSS() {
+  private Document mDocument;
```
jasta: final

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -616,166 +335,73 @@ private void releaseChildNodeRemovedEvent(ChildNodeRemovedEvent childNodeRemoved
...
-      mDOMProvider.verifyThreadAccess();
+    protected synchronized void onFirstPeerRegistered() {
+      // We don't call mDOMProvider.setListener() until the first DOM.getDocument request.
 
-      AttributeModifiedEvent message = new AttributeModifiedEvent();
-      message.nodeId = mObjectIdMapper.getIdForObject(element);
-      message.name = name;
-      message.value = value;
-      mPeerManager.sendNotificationToPeers("DOM.attributeModified", message);
+      mDocument.createDomProviderIfNoneExists();
```
jasta: This looks funky to me.  Why is the document not self managing?  Also, why is there not a document peer manager now?  It would seem that now that DOM and CSS are being intertwined together then the concept of them being enabled and the related peers also seem to be interconnected.  In other words, what would your code do if a caller invoked CSS.enable but not DOM.enable?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -616,166 +335,73 @@ private void releaseChildNodeRemovedEvent(ChildNodeRemovedEvent childNodeRemoved
...
-      message.nodeId = mObjectIdMapper.getIdForObject(element);
-      message.name = name;
-      mPeerManager.sendNotificationToPeers("DOM.attributeRemoved", message);
-    }
-
-    @Override
-    public void onInspectRequested(Object element) {
-      mDOMProvider.verifyThreadAccess();
+    protected synchronized void onLastPeerUnregistered() {
+      mDocument.cleanUp();
```
jasta: Same as above.  DOM is doing the state management for Document, even though CSS also uses it.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
ichub: Prior to the refactoring, `DOM` created an instance of `DOMProvider` inside `OnFirstPeerRegistered`, and removed it in `OnLastPeerUnregistered`. `DOM` and `CSS` have unique peer managers, and `Document` does not/should not have a peer manager (since it is not a domain class). 

Now, since `Document` owns `DOMProvider`, but has no peer manager, it has to delegate the responsibility of creating and destroying DOMProvider to the domain classes (DOM, CSS, etc.). 

As of this moment, the only class that makes use of `Document` is `DOM`, which is why this logic remained in `DOM`. 

A possible solution to this would be to use reference counting, and take care of creating/destroying `DOMProvider` inside of `Document`.

Another solution is to just create a `DOMProvider` when the app starts, without worrying about whether or not there are peers.

jasta: Letting the `DOMProvider` leak isn't an option, so let's just scratch that off now (it uses way too much memory)...

As far as lifecycle management, your assertion that only domain modules should have a peer manager is reasonable, although I don't fully accept this argument given that CSS and DOM are being made codependent and your code, as designed here, assumes erroneously that DOM is somehow the "real" owner but nothing reasonably supports that in how the code or the protocol is laid out.

I think reference counting is the right compromise here.

stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
```diff
@@ -0,0 +1,483 @@
...
+  public Object getElement(int id) {
+    return mObjectIdMapper.getObjectForId(id);
+  }
+
+  public String[] getElementAttributes(Object element) {
+    final Document.AttributeListAccumulator accumulator = getCachedAttributeListAccumulator();
+    NodeDescriptor descriptor = mDOMProvider.getNodeDescriptor(element);
+    descriptor.getAttributes(element, accumulator);
+
+    String[] result = accumulator.toArray(new String[accumulator.size()]);
```
rickbrew: we should not be returning an array. let the caller send in an AttributeAccumulator.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -228,10 +186,10 @@ public PerformSearchResponse performSearch(JsonRpcPeer peer, final JSONObject pa
 
     final Pattern queryPattern = Pattern.compile(Pattern.quote(request.query), Pattern.CASE_INSENSITIVE);
```
rickbrew: like we discussed, let's go ahead and move Search into Document

stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
```diff
@@ -0,0 +1,483 @@
...
+
+  private int mReferenceCounter;
+
+  public Document(DOMProvider.Factory factory) {
+    mFactory = factory;
+    mObjectIdMapper = new DOMObjectIdMapper();
+    mReferenceCounter = 0;
+  }
+
+  public synchronized void AddRef() {
```
rickbrew: capitalization, e.g. addRef()

stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
```diff
@@ -0,0 +1,483 @@
...
+    if (mReferenceCounter++ == 0) {
+      mDOMProvider = mFactory.create();
+    }
+  }
+
+  public synchronized void Release() {
```
rickbrew: capitalization, e.g. release()

stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
```diff
@@ -0,0 +1,483 @@
...
+      extends ArrayList<String> implements AttributeAccumulator {
+
+    @Override
+    public void store(String name, String value) {
+      add(name);
+      add(value);
+    }
+  }
+
+  private void cleanUp() {
```
rickbrew: move this lifecycle method up with addref/release so that they're all grouped together

stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
```diff
@@ -0,0 +1,483 @@
...
+import com.facebook.stetho.inspector.helper.ObjectIdMapper;
+
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Queue;
+
+public class Document {
```
rickbrew: final

stetho/src/main/java/com/facebook/stetho/Stetho.java
```diff
@@ -133,10 +134,12 @@ public static InspectorModulesProvider defaultInspectorModulesProvider(final Con
...
         ArrayList<ChromeDevtoolsDomain> modules = new ArrayList<ChromeDevtoolsDomain>();
         modules.add(new Console());
-        modules.add(new CSS());
         modules.add(new Debugger());
         if (Build.VERSION.SDK_INT >= AndroidDOMConstants.MIN_API_LEVEL) {
-          modules.add(new DOM(new AndroidDOMProviderFactory((Application)context.getApplicationContext())));
+          Document doc = new Document(new AndroidDOMProviderFactory((Application)context.getApplicationContext()));
+
+          modules.add(new DOM(doc));
+          modules.add(new CSS(doc));
```
rickbrew: This makes it so that the CSS module also requires the same minimum API level as DOM. @jasta should comment on this too.

stetho/src/main/java/com/facebook/stetho/inspector/helper/ThreadBoundProxy.java
```diff
@@ -0,0 +1,56 @@
...
+import com.facebook.stetho.common.ThreadBound;
+import com.facebook.stetho.common.UncheckedCallable;
+import com.facebook.stetho.common.Util;
+
+/*
+This class is for those cases when is a class' threading
+policy is determined by one of its member variables.
+ */
+public abstract class ThreadBoundProxy implements ThreadBound {
+  private ThreadBound mEnforcer;
```
rickbrew: this can be made 'final'

stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
```diff
@@ -0,0 +1,602 @@
...
+  private DocumentProvider mDocumentProvider;
+  private ShadowDocument mShadowDOM;
+  private ObjectIdMapper mObjectIdMapper;
+
+  private UpdateListenerCollection mUpdateListeners;
+
+  private ChildEventingList mCachedChildEventingList;
+  private ArrayListAccumulator<Object> mCachedChildrenAccumulator;
+
+  private int mReferenceCounter;
```
rickbrew: because this is only accessed by `addRef` and `release`, and because they're synchronized, we can add a `GuardedBy("this")` :

```
@GuardedBy("this")
private int mReferenceCounter;
```

stetho/src/main/java/com/facebook/stetho/inspector/helper/ThreadBoundProxy.java
```diff
@@ -0,0 +1,56 @@
...
+
+  public ThreadBoundProxy(ThreadBound enforcer) {
+    mEnforcer = Util.throwIfNull(enforcer);
+  }
+
+  @Override
+  public boolean checkThreadAccess() {
```
rickbrew: let's mark all of the `ThreadBound` methods as `final`, so long as nobody else is overriding them. The whole point of deriving from this class is to allow the base class to handle them, so why not make sure derived classes can't even accidentally override them :)

stetho/src/main/java/com/facebook/stetho/inspector/helper/DocumentAccessRegisteredListener.java
```diff
@@ -0,0 +1,30 @@
...
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.helper;
+
+import com.facebook.stetho.inspector.elements.Document;
+
+public class DocumentAccessRegisteredListener extends PeersRegisteredListener {
```
rickbrew: Please add a javadoc explaining why you'd want to use this class

stetho/src/main/java/com/facebook/stetho/inspector/helper/ThreadBoundProxy.java
```diff
@@ -0,0 +1,56 @@
...
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.helper;
+
+import com.facebook.stetho.common.ThreadBound;
+import com.facebook.stetho.common.UncheckedCallable;
+import com.facebook.stetho.common.Util;
+
+/*
```
rickbrew: The standard commenting form for this, "javadoc", goes like this:

```
/**
 * Lines of text go after the first line of slash-star-star.
 * Each line of text starts with space-star-space.
 * The closing line (the next line) is space-star-slash
 */
```

stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
```diff
@@ -0,0 +1,602 @@
...
+
+  public void addUpdateListener(UpdateListener updateListener) {
+    mUpdateListeners.add(updateListener);
+  }
+
+  public void removeUpdateListener(UpdateListener updateListener) {
+    mUpdateListeners.remove(updateListener);
+  }
+
+  public NodeDescriptor getNodeDescriptor(Object element) {
```
rickbrew: because `DocumentProvider.getNodeDescriptor` is Nullable, we should mark this method with that also.

stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
```diff
@@ -0,0 +1,602 @@
...
+    verifyThreadAccess();
+    mDocumentProvider.setInspectModeEnabled(enabled);
+  }
+
+  public int getNodeIdForElement(Object element) {
+    verifyThreadAccess();
+    return mObjectIdMapper.getIdForObject(element);
+  }
+
+  public Object getElementForNodeId(int id) {
```
rickbrew: `ObjectIdMapper.getObjectForId` can return null and is marked as `@Nullable`, so we should mark this method as `@Nullable` too

stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
```diff
@@ -0,0 +1,602 @@
...
+    verifyThreadAccess();
+    return mObjectIdMapper.getObjectForId(id);
+  }
+
+  public DocumentView getDocumentView() {
```
rickbrew: One thing we're going to have to be careful with here is the fact that `DocumentView` seems like it should be providing an immutable "view". However, it's not. `ShadowDOM` can and will change if you access it across tree updates.

There are a few things we should do here. First, we should add documentation to this method that says you should only use the `DocumentView` "temporarily" ... don't stash it in a field and access it later and expect it to be unchanged! Only use it for the duration of the calling method. Second, let's file a task to go and make this area less prone to interpretation errors (it's a potential bug hazard). That may mean we rename some things, or it may mean that `ShadowDOM` doesn't implement `DocumentView` directly and instead can hand out a `DocumentView` which is, in fact, immutable.

stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
```diff
@@ -0,0 +1,602 @@
...
+
+      if (doesElementMatch(queryPattern, childElement)) {
+        matchedIds.store(mObjectIdMapper.getIdForObject(childElement));
+      }
+
+      findMatches(queryPattern, childElement, matchedIds);
+    }
+  }
+
+  private boolean doesElementMatch(Pattern queryPattern, Object element) {
```
rickbrew: let's switch the order of the arguments here (we probably should've done this with the initial check-in for Search). generally when you do an operation on an object, the object is the first parameter to the method and the parameters that describe the operation are placed afterward.

stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
```diff
@@ -0,0 +1,602 @@
...
+      }
+    }
+
+    return queryPattern.matcher(descriptor.getNodeName(element)).find();
+  }
+
+  private ChildEventingList acquireChildEventingList(
+      Object parentElement,
+      DocumentView documentView) {
+    ChildEventingList childEventingList = mCachedChildEventingList;
```
rickbrew: This method has a bug in it. We need to set `mCachedChildEventingList` to null! Otherwise we could hand it out to multiple call sites in the event of some recursive algorithm or reentrant code block.

stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
```diff
@@ -0,0 +1,602 @@
...
+      if (doesElementMatch(queryPattern, childElement)) {
+        matchedIds.store(mObjectIdMapper.getIdForObject(childElement));
+      }
+
+      findMatches(queryPattern, childElement, matchedIds);
+    }
+  }
+
+  private boolean doesElementMatch(Pattern queryPattern, Object element) {
+    AttributeListAccumulator accumulator = new AttributeListAccumulator();
```
rickbrew: for now we should have a `// TODO: cache this` attached to this line of code. Then on our final round of review we can Ctrl+F and find all the `// TODO`'s and clean them up.

stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
```diff
@@ -0,0 +1,602 @@
...
+    verifyThreadAccess();
+
+    if (mDocumentProvider.getRootElement() != mShadowDOM.getRootElement()) {
+      throw new IllegalStateException();
+    }
+
+    ArrayListAccumulator<Object> childrenAccumulator = acquireChildrenAccumulator();
+
+    ShadowDocument.UpdateBuilder updateBuilder = mShadowDOM.beginUpdate();
+    Queue<Object> queue = new ArrayDeque<>();
```
rickbrew: I think we can add a `// TODO: cache this` on here too, and then resolve this before we're done with this PR.

stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
```diff
@@ -0,0 +1,602 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Queue;
+import java.util.regex.Pattern;
+
+public final class Document extends ThreadBoundProxy {
+  private final DocumentProvider.Factory mFactory;
+
+  private DocumentProvider mDocumentProvider;
+  private ShadowDocument mShadowDOM;
```
rickbrew: perhaps this should be called `mShadowDocument` ? :)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDOMProviderFactory.java
```diff
@@ -11,18 +11,59 @@
...
   public AndroidDOMProviderFactory(Application application) {
     mApplication = Util.throwIfNull(application);
+    mHandler = new Handler(Looper.getMainLooper());
   }
 
   @Override
-  public DOMProvider create() {
-    return new AndroidDOMProvider(mApplication);
+  public DocumentProvider create() {
+    return new AndroidDocumentProvider(mApplication, mHandler);
```
rickbrew: hmm. If `AndroidDocumentProvider` only uses the `Handler` for the sake of implementing `ThreadBound`, then we can probably just pass `this` as the 2nd parameter, which could be re-typed to be of type `ThreadBound`.

```
public AndroidDocumentProvider(Application application, ThreadBound threadBound) { ...
```

This would then allow `AndroidDocumentProvider` to derive from your new `ThreadBoundProxy`.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -196,14 +154,20 @@ public ResolveNodeResponse resolveNode(JsonRpcPeer peer, JSONObject params)
...
         params, SetAttributesAsTextRequest.class);
-    final Object object = mObjectIdMapper.getObjectForId(request.nodeId);
 
-    mDOMProvider.postAndWait(new Runnable() {
-      @Override
-      public void run() {
-        mDOMProvider.setAttributesAsText(object, request.text);
-      }
-    });
+    final Object element = mDocument.getElementForNodeId(request.nodeId);
```
rickbrew: Why was the implementation of `setAttributesAsText` completely changed? It's not related to what we're doing here. Plus, this new code here in `DOM` is incorrect! All these methods on `mDocument` need to be called on the thread dictated by its `ThreadBound` implementation. We should really put this all back the way it was unless there's a really good reason for all this.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDOMProviderFactory.java
ichub: I tried doing this, but there is an issue since we shouldn't override methods of `ThreadBoundProxy`, while the implementation of `postDelayed(...)` in `AndroidDocumentProvider` is different from the one in `ThreadBoundProxy`.

ichub: In `AndroidDocumentProvider`:

```
@Override
public void postDelayed(Runnable r, long delayMillis) {
  if (!mHandler.postDelayed(r, delayMillis)) {
    throw new RuntimeException("Handler.postDelayed() returned false");
  }
}
```

Whereas in `ThreadBoundProxy`:
```
@Override
public final void postDelayed(Runnable r, long delayMillis) {
  mEnforcer.postDelayed(r, delayMillis);
}
```

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -226,18 +190,12 @@ public PerformSearchResponse performSearch(JsonRpcPeer peer, final JSONObject pa
         params,
         PerformSearchRequest.class);
 
-    final Pattern queryPattern = Pattern.compile(Pattern.quote(request.query), Pattern.CASE_INSENSITIVE);
+    final ArrayListAccumulator<Integer> results = new ArrayListAccumulator<>();
```
rickbrew: I think there's an opportunity to increase clarity here ... what do the integers inside of `results` represent? Are they node IDs? If so, calling this `nodeIdResults` might help to illuminate this fact.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
ichub: Before, `DomProvider#setAttributesAsText(...)` just fell through to `NodeDescriptor#setAttributesAsText(...)`. Now we just set it through an instance of `NodeDescriptor` because that way there is only one place where we can set attributes.

stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
```diff
@@ -0,0 +1,602 @@
...
+        mShadowDOM = null;
+        mObjectIdMapper.clear();
+        mDocumentProvider.dispose();
+        mUpdateListeners.clear();
+        mDocumentProvider = null;
+      }
+    });
+  }
+
+  public void addUpdateListener(UpdateListener updateListener) {
```
rickbrew: We need to figure out the right way to make sure this add/remove listener stuff is thread safe. Notice how we call things like `Document.removeUpdateListener` during `DOM`'s `lastPeerUnregistered`, which will happen on a different thread.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
ichub: See this diff: c1fb7343c3479a6e524a21d18d3d2578ec1c2478

rickbrew:
Okay there's another round of feedback. This is shaping up well, despite how confusing it all is :)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDOMProviderFactory.java
```diff
@@ -11,18 +11,59 @@
...
+import android.os.Looper;
+import com.facebook.stetho.common.ThreadBound;
+import com.facebook.stetho.common.UncheckedCallable;
 import com.facebook.stetho.common.Util;
-import com.facebook.stetho.inspector.elements.DOMProvider;
+import com.facebook.stetho.common.android.HandlerUtil;
+import com.facebook.stetho.inspector.elements.DocumentProvider;
 
-public class AndroidDOMProviderFactory implements DOMProvider.Factory {
+public final class AndroidDOMProviderFactory
```
rickbrew: rename class to `AndroidDocumentProviderFactory`

stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
```diff
@@ -0,0 +1,637 @@
...
+    }
+
+    mCachedChildEventingList = null;
+
+    childEventingList.acquire(parentElement, documentView);
+    return childEventingList;
+  }
+
+  private AttributeListAccumulator acquireCachedAttributeAccumulator() {
+    if (mCachedAttributeAccumulator == null) {
```
rickbrew: This method has a bug. We're using a singleton caching system here, but this implementation will hand out the same `AttributeListAccumulator` more than once if its called twice without a call to `releaseCachedAttributeAccumulator` in-between.

ichub: If it's being cleared, why does that matter?

ichub:
```
else if (mCachedAttributeAccumulator.size() > 0) {
    mCachedAttributeAccumulator.clear();
 }
```

ichub: Oh I see, because it is legal to call it more than once before calling return.

stetho/src/main/java/com/facebook/stetho/inspector/helper/DocumentAccessRegisteredListener.java
```diff
@@ -0,0 +1,35 @@
...
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.helper;
+
+import com.facebook.stetho.inspector.elements.Document;
+
+/*
```
rickbrew: javadocs starts with slash-star-star, not slash-star. Saw a few other cases of this too

rickbrew: This class only does 1 little thing for clients ... it calls `addRef` and `release`. Do we need to put this into a base class? Seems like `DOM` and `CSS` could probably handle this by themselves without being required to derive their `PeersRegisteredListener` from this?

stetho/src/main/java/com/facebook/stetho/inspector/helper/ThreadBoundProxy.java
```diff
@@ -0,0 +1,107 @@
...
+package com.facebook.stetho.inspector.helper;
+
+import android.os.Handler;
+import com.facebook.stetho.common.ThreadBound;
+import com.facebook.stetho.common.UncheckedCallable;
+import com.facebook.stetho.common.Util;
+import com.facebook.stetho.common.android.HandlerUtil;
+
+/*
+ * This class is for those cases when is a class' threading
```
rickbrew: grammar ... "when is a class" :)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDocumentProvider.java
```diff
@@ -71,15 +72,17 @@ public void run() {
     }
   };
 
-  public AndroidDOMProvider(Application application) {
+  public AndroidDocumentProvider(Application application, Handler handler) {
```
rickbrew: why not just have the 2nd parameter be a `ThreadBound` ? we don't use the handler; we only used it to implement `ThreadBound`. But the factory does that now and we're deferring to it ... so why not have our ctor here take a `ThreadBound`, pass it to the super class, and don't even bother storing it in a field? That will also allow us to remove the ctor of `ThreadBoundProxy` which takes a `Handler`, and then we can remove the `HandlerThreadBoundWrapper`. (this is what I intended when I made earlier comments about this :) )

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -616,166 +307,85 @@ private void releaseChildNodeRemovedEvent(ChildNodeRemovedEvent childNodeRemoved
...
-    public void onAttributeRemoved(Object element, String name) {
-      mDOMProvider.verifyThreadAccess();
-
-      AttributeRemovedEvent message = new AttributeRemovedEvent();
-      message.nodeId = mObjectIdMapper.getIdForObject(element);
-      message.name = name;
-      mPeerManager.sendNotificationToPeers("DOM.attributeRemoved", message);
+    protected synchronized void onFirstPeerRegistered() {
+      super.onFirstPeerRegistered();
+      mDocument.addRef();
```
rickbrew: my comment about `DocumentAccessRegisteredListener` ... this listener is deriving from that and then calls `mDocument.addRef` and `mDocument.release` anyway, so we end up calling these twice. I think we should just get rid of `DocumentAccessRegisteredListener`

stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
```diff
@@ -0,0 +1,637 @@
...
+
+package com.facebook.stetho.inspector.elements;
+
+import android.os.SystemClock;
+import com.facebook.stetho.common.Accumulator;
+import com.facebook.stetho.common.ArrayListAccumulator;
+import com.facebook.stetho.common.LogUtil;
+import com.facebook.stetho.inspector.helper.ObjectIdMapper;
+import com.facebook.stetho.inspector.helper.ThreadBoundProxy;
+import org.w3c.dom.Attr;
```
rickbrew: wut?

stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
rickbrew: Yeah our singleton caching is a "good faith" optimization. If you have recursive or reentrant code, it will trip over this

stetho/src/main/java/com/facebook/stetho/inspector/helper/DocumentAccessRegisteredListener.java
ichub: The functionality is the exact same in DOM and CSS, so I think it makes sense to put it in one place.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDocumentProvider.java
```diff
@@ -197,7 +166,7 @@ public void setInspectModeEnabled(boolean enabled) {
     }
   }
 
-  @Override
+   @Override
```
rickbrew: errant bit of whitespace?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -245,49 +199,15 @@ public PerformSearchResponse performSearch(JsonRpcPeer peer, final JSONObject pa
     // it can be queried later.
     final String id = String.valueOf(mResultCounter.getAndIncrement());
```
rickbrew: renaming this from `id` to `queryId` might make things a little more clear

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -318,267 +238,38 @@ public void discardSearchResults(JsonRpcPeer peer, JSONObject params) {
...
+    node.nodeType = descriptor.getNodeType(element);
+    node.nodeName = descriptor.getNodeName(element);
+    node.localName = descriptor.getLocalName(element);
+    node.nodeValue = descriptor.getNodeValue(element);
 
-    if (mDOMProvider.getRootElement() != mShadowDOM.getRootElement()) {
-      throw new IllegalStateException();
-    }
+    Document.AttributeListAccumulator accumulator = new Document.AttributeListAccumulator();
+    mDocument.getNodeDescriptor(element).getAttributes(element, accumulator);
```
rickbrew: why are we retrieving the node descriptor again? we already have it

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -36,55 +29,35 @@
...
-    mPeerManager = new ChromePeerManager();
-    mPeerManager.setListener(new PeerManagerListener());
+  private ChildNodeRemovedEvent mCachedChildNodeRemovedEvent;
+  private ChildNodeInsertedEvent mCachedChildNodeInsertedEvent;
 
+  public DOM(Document document) {
     mObjectMapper = new ObjectMapper();
-    mObjectIdMapper = new DOMObjectIdMapper();
-
+    mDocument = document;
```
rickbrew: `mDocument = Util.throwIfNull(document);`

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -21,7 +24,13 @@
 import org.json.JSONObject;
 
 public class CSS implements ChromeDevtoolsDomain {
-  public CSS() {
+  private final ChromePeerManager mPeerManager;
+  private final Document mDocument;
+
+  public CSS(Document document) {
```
rickbrew: since `CSS` isn't actually using `Document` yet, let's not modify `CSS` in this PR

rickbrew:
Looks like we're almost done here. You have some conflicts to resolve, then need to squash, and then I think we can merge!

PR #238, opened 2015-07-29
Add support for EXPLAIN queries run via Web SQL console.

dougsparling:
Hello,

I noticed that `EXPLAIN` queries run via the Web SQL console were failing with the error: *Queries can be performed using SQLiteDatabase query or rawQuery methods only.*

I tried to add `EXPLAIN` support by going through the path for `SELECT`, and sure enough, I got back a usable result.

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/database/DatabasePeerManager.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/database/DatabasePeerManager.java
@@ -9,7 +9,6 @@
 
 package com.facebook.stetho.inspector.database;
 
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


stetho/src/main/java/com/facebook/stetho/inspector/database/DatabasePeerManager.java
```diff
@@ -41,6 +41,11 @@
       "-uid",
       "-wal"
   };
+  private static final Set<String> QUERY_TYPES_WITH_RESULT = new HashSet<>(Arrays.asList(
```
jasta: We can do this better and faster using string switches.  We already require java 7 so this is safe.  See example below.

stetho/src/main/java/com/facebook/stetho/inspector/database/DatabasePeerManager.java
```diff
@@ -143,14 +148,13 @@ private static String removeSuffix(String str, String[] suffixesToRemove) {
     Util.throwIfNull(handler);
     SQLiteDatabase database = openDatabase(databaseName);
     try {
-      String firstWord = getFirstWord(query);
+      String firstWordUpperCase = getFirstWord(query).toUpperCase();
 
-      if (firstWord.equalsIgnoreCase("UPDATE") || firstWord.equalsIgnoreCase("DELETE")) {
+      if (firstWordUpperCase.equals("UPDATE") || firstWordUpperCase.equals("DELETE")) {
```
jasta:
```java
switch (firstWordUpperCase) {
  case "UPDATE":
  case "DELETE":
    ...
    break;
  case "INSERT":
    ...
    break;
  case "SELECT":
  case "PRAGMA":
  case "EXPLAIN":
    ...
    break;
}
```

jasta:
Prefer the switch, otherwise lgtm.  Please make sure to squash your update so that it remains a single commit.

dougsparling:
I agree, the switch is much easier to read. I've replaced both the set and if blocks with a switch.

PR #241, opened 2015-08-03
Add checkstyle rules

jasta:
This is intended to run on Travis-CI to fail builds which don't adhere
to the style.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,3 +12,9 @@ allprojects {
         jcenter()
     }
 }
+
+apply plugin: 'checkstyle'
+
+checkstyle {
+    configFile = new File("${project.projectDir}", "checkstyle.xml")
+}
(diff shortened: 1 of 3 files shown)


jasta:
Oops, didn't work for some reason.

build.gradle
```diff
@@ -12,3 +12,9 @@ allprojects {
         jcenter()
     }
 }
+
+apply plugin: 'checkstyle'
+
+checkstyle {
+    configFile = new File("${project.projectDir}", "checkstyle.xml")
+}
```
Turbo87: - the `configFile` should be configured using the `rootDir` project variable
- you should explicitly set the checkstyle version used by gradle via `toolVersion` (note that some Gradle versions are incompatible with same checkstyle versions. e.g. Gradle v2.5 isn't playing nice with checkstyle v6.8)

````groovy
checkstyle {
    configFile = new File(rootDir, 'checkstyle.xml')
    toolVersion = '6.7'
}
````

Turbo87: also you should probably put this block into the `allprojects` block above to apply the plugin to all subprojects

PR #245, opened 2015-08-04
Autodetect stetho-js-rhino and use it

jasta:
This diff focuses on making stetho-js-rhino a bit more of a first-party
citizen in Stetho, with the default implementation both detecting it
more conveniently and automatically using it if available or
recommending users add it to the classpath so that it will be
autodetected.

### DIFF
--- a/stetho-js-rhino/README.md
+++ b/stetho-js-rhino/README.md
@@ -21,26 +21,8 @@ or Maven:
 Make sure that you depend on the main `stetho` dependency too.
 
 ### Putting it together
-The JavaScript integration is similar to standard Stetho integration.
-The main difference is with the WebKitInspector used.
-There is a simple initialization step which occurs in your `Application` class:
-
-```java
-public class MyApplication extends Application {
-  public void onCreate() {
-    super.onCreate();
-+   JsRuntimeBuilder jsRuntimeBuilder = new JsRuntimeBuilder(this);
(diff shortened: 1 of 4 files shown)


jasta:
@potyl working on making this a bit easier to discover in preparation for the v1.2.0 release which will be highlighting this as a main feature.

stetho-sample/build.gradle
```diff
@@ -31,6 +31,7 @@ android {
 dependencies {
     debugCompile project(':stetho')
     debugCompile project(':stetho-urlconnection')
+    debugCompile project(':stetho-js-rhino')
```
jasta: Oops, this should not have been committed.

potyl: I am sorry. That one slipped through.

jasta:
@potyl understood re configuration.  This diff is designed to simplify the "default" configuration, which I'm assuming to be the common case for just trying it out.  If you want to configure it, your old API is still used except I removed the `InspectorModulesProvider` method at the end of the builder.  It's no longer needed with #234 (providing a better API to creating a custom modules provider means we don't need to build one for you).

What we could do here is make it warn in the console when the "default" configuration is used, as a kind of self-documenting measure so that folks can see how things can be expanded from there.  My goal with this diff is merely to get the system more discoverable and easier to "try out".

PR #248, opened 2015-08-06
Implement transparent request decompression

### LINKED ISSUE #111: Requests with "Content-Encoding: deflate" don't seem to get re-inflated
Looking at #7, it looks like support for inflation was added there. It hasn't worked for us in our testing though. Using okhttp + interceptor, confirmed that the `Content-Encoding` header is there and correctly set to `deflate`. Is there any other configuration that needs to be done in order for this to be handled correctly?

jasta:
Request decompression is not officially part of the standard but many
mobile implementations choose to do it because payloads tend to be quite
large on clients and servers these days are easily made to support it.
This diff simply makes it so that Stetho will show the decompressed
payload, but still report the compressed payloads size to the UI.

Note that this only affects stetho-urlconnection and stetho-okhttp.
Other modules must manually implement this as it is not possible given
the existing API contract to do it transparently.

Closes #111

### DIFF
--- a/stetho-okhttp/src/test/java/com/facebook/stetho/okhttp/StethoInterceptorTest.java
+++ b/stetho-okhttp/src/test/java/com/facebook/stetho/okhttp/StethoInterceptorTest.java
@@ -15,9 +15,18 @@ import com.facebook.stetho.inspector.network.DecompressionHelper;
 import com.facebook.stetho.inspector.network.NetworkEventReporter;
 import com.facebook.stetho.inspector.network.NetworkEventReporterImpl;
 import com.facebook.stetho.inspector.network.ResponseHandler;
-import com.squareup.okhttp.*;
+import com.squareup.okhttp.Connection;
+import com.squareup.okhttp.Interceptor;
+import com.squareup.okhttp.MediaType;
+import com.squareup.okhttp.OkHttpClient;
+import com.squareup.okhttp.Protocol;
+import com.squareup.okhttp.Request;
+import com.squareup.okhttp.RequestBody;
+import com.squareup.okhttp.Response;
+import com.squareup.okhttp.ResponseBody;
 import com.squareup.okhttp.mockwebserver.MockResponse;
 import com.squareup.okhttp.mockwebserver.MockWebServer;
+import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
@@ -33,13 +42,14 @@ import org.robolectric.RobolectricTestRunner;
 import org.robolectric.annotation.Config;
 
 import javax.annotation.Nullable;
+
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.zip.GZIPOutputStream;
 
-import static org.junit.Assert.assertArrayEquals;
-import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.*;
 import static org.mockito.Matchers.*;
 
 @Config(emulateSdk = Build.VERSION_CODES.JELLY_BEAN)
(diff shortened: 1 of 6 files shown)


stetho-okhttp/src/test/java/com/facebook/stetho/okhttp/StethoInterceptorTest.java
```diff
@@ -50,24 +60,37 @@
...
+    PowerMockito.when(NetworkEventReporterImpl.get()).thenReturn(mMockEventReporter);
 
-    StethoInterceptor interceptor = new StethoInterceptor();
+    mInterceptor = new StethoInterceptor();
+    mClientWithInterceptor = new OkHttpClient();
+    mClientWithInterceptor.networkInterceptors().add(mInterceptor);
+  }
+
+  @Test
+  public void testHappyPath() throws IOException {
```
longinoa: rename this to `testWithoutRequestCompression`

stetho/src/main/java/com/facebook/stetho/inspector/network/RequestBodyHelper.java
```diff
@@ -0,0 +1,80 @@
...
+
+  private ByteArrayOutputStream mDeflatedOutput;
+  private CountingOutputStream mDeflatingOutput;
+
+  public RequestBodyHelper(NetworkEventReporter eventReporter, String requestId) {
+    mEventReporter = eventReporter;
+    mRequestId = requestId;
+  }
+
+  public OutputStream createBodySink(@Nullable String contentEncoding) throws IOException {
```
longinoa: I wish we could actually return a `Sink` here but understand that urlConnection doesn't have a notion of that. May want to change the name of this to reflect what is actually being created here.

stetho/src/main/java/com/facebook/stetho/inspector/network/DecompressionHelper.java
```diff
@@ -20,8 +20,8 @@
 
 // @VisibleForTest
 public class DecompressionHelper {
```
longinoa: Given how we are doing this in multiple places now it may make sense to create an interface to allow people to specify additional compression/decompression methods in a more generic way.

For example if someone wants to use bzip they would have to re-implement `NetworkEventReporterImpl` as well as the `RequestBodyHelper`. If we had a simple registry system from Content-Encoding to an interface that exposes a 
compress and decompress function it would enable this.

stetho/src/main/java/com/facebook/stetho/inspector/network/RequestBodyHelper.java
```diff
@@ -0,0 +1,80 @@
...
+
+    mDeflatingOutput = new CountingOutputStream(deflatingOutput);
+    mDeflatedOutput = deflatedOutput;
+
+    return mDeflatingOutput;
+  }
+
+  public byte[] getDisplayBody() {
+    throwIfNoBody();
+    return mDeflatedOutput.toByteArray();
```
longinoa: Boo now we have to create 2 byte[] to satisfy the counting :(

stetho-urlconnection/src/main/java/com/facebook/stetho/urlconnection/URLConnectionInspectorRequest.java
```diff
@@ -22,21 +24,21 @@
   private final String mRequestId;
   private final String mFriendlyName;
   @Nullable private final SimpleRequestEntity mRequestEntity;
+  private final RequestBodyHelper mRequestBodyHelper;
```
longinoa: `@Nullable`

longinoa:
The only thing that is kind of dubious about this is the fact that `NetworkEventReporter#body()` has now changed. It is no longer the faithful body, I wonder if it is worth adding a `decompressedBody()` function so that there will be no change to callers who may be depending on the fact that `body` is the compressed version

stetho-okhttp/src/test/java/com/facebook/stetho/okhttp/StethoInterceptorTest.java
jasta: Ack'd, but it should be `testManualInterceptorNoCompression` -- there is already a `testWithResponseCompression`...

stetho/src/main/java/com/facebook/stetho/inspector/network/RequestBodyHelper.java
jasta: Unfortunately sink feels like the best name for the job :)

stetho/src/main/java/com/facebook/stetho/inspector/network/DecompressionHelper.java
jasta: Ack'd, though there are other, deeper problems which prevent this.  The API would need a lot of work

stetho/src/main/java/com/facebook/stetho/inspector/network/RequestBodyHelper.java
jasta: Yeah, lame for okhttp.  Perhaps in the future we can refactor further to allow for subclassing...

jasta:
Discussed out of band re `#body` changing.  It remains the same from Stetho's point of view and only changes if the caller chooses for it to change, as is the case with only `stetho-urlconnection` and `stetho-okhttp`

PR #249, opened 2015-08-07
Fix RequestBodyHelper body inflation

hzsweers:
Following up from #111 and testing #248 

This was using the wrong class (Deflate is the compression, inflate is the reverse). `InflaterOutputStream` properly inflates the request body, and works for our project when I tested out the change.

CC @jasta


hzsweers:
Sent an email about the CLA... I changed my username from pandanomic to hzsweers after I signed it before, so I need to get that updated.

PR #253, opened 2015-08-12
Rhino autodetect singleton (based on jasta:rhino-autodetect)

potyl:
This is an enhancement to the branch jasta:rhino-autodetect that provides an easy way for apps to configure the JavaScript runtime while using the new built-in autodtection.

Sorry if I don't reply in the next two weeks, I'll be in holidays and I might not have access to network (or even a laptop).

### DIFF
--- a/stetho-js-rhino/src/main/java/com/facebook/stetho/rhino/JsRuntimeReplFactoryBuilder.java
+++ b/stetho-js-rhino/src/main/java/com/facebook/stetho/rhino/JsRuntimeReplFactoryBuilder.java
@@ -49,6 +49,11 @@ public class JsRuntimeReplFactoryBuilder {
   private static final String SOURCE_NAME = "chrome";
 
   /**
+   * Singleton instance.
+   */
+  private static volatile JsRuntimeReplFactoryBuilder sInstance;
+
+  /**
    * Android application context.
    */
   private final android.content.Context mContext;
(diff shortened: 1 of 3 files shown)


jasta:
Sorry was on vacation these past few weeks and didn't see this.  I'd like to look at manually merging this in before we cut the release I was supposed to do before I left :)

potyl:
I was too on holidays and I couldn't comment too much on the patches. I rebased my branch with master and gave it a try. Things are not as I expected them.

It seems that by default we create a Runtime without a context thus running in legacy mode. Is this intended?
In legacy mode there's no JavaScript configuration/autodetection done.

Wihtout changing `Stetho.DefaultInspectorModulesBuilder.finish()` to invoke `new Runtime(context)`. An app can't no longer just plug-and-play stetho-js-rhino (which was the original intent on my patch). The app would need to provide a full modules builder with all modules just to enable JavaScript.

Without the change then each time that we type an expression in the console we get: **"Not supported with legacy Runtime module"**

Assuming that we do change Stetho.DefaultInspectorModulesBuilder.finish() to invokes new Runtime(context).
Then the following happens:
 - We don't load stetho-js-rhino in gradle and the console shows the following after evaluating an expression: **"Not supported without stetho-js-rhino dependency"**
 - We load stetho-js-rhino in gradle, everything works fine. JavaScript is now available.

I think that the easiest / friendliest thing is to build a Runtime with a Context.

So in this branch I change the default Runtime to be constructed with a Context and I enhance the JavaScript builder to be a singleton thus giving apps an easy access to the JavaScript runtime.

jasta:
Yeah I noticed that I made some mistakes in my earlier implementation.  I am working on some small fixes and incorporating your changes as well.

stetho-js-rhino/src/main/java/com/facebook/stetho/rhino/JsRuntimeReplFactoryBuilder.java
```diff
@@ -74,11 +79,18 @@
    */
   private final Map<String, Function> mFunctions = new HashMap<>();
 
-  public static RuntimeReplFactory defaultFactory(@NonNull android.content.Context context) {
-    return new JsRuntimeReplFactoryBuilder(context).build();
+  public static synchronized JsRuntimeReplFactoryBuilder getInstance(@NonNull android.content.Context context) {
```
jasta: I think that #265 makes this unnecessary.  The new API would be:
```java
new DefaultInspectorModulesProvider(context)
    .runtimeRepl(
        new JsRuntimeReplFactoryBuilder(context)
            .addStuff(...)
            .finish())
    .finish();
```

jasta:
See #268 and #269

PR #254, opened 2015-08-13
Add schema caching functionality for async pretty printing

mengdilin:
Right now, one schema is downloaded for every pretty printing instantiated. Often times, async pretty printers use the same schema.

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterSchemaManager.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterSchemaManager.java
@@ -17,33 +17,33 @@ import java.util.Map;
 
 public final class AsyncPrettyPrinterSchemaManager {
   private static AsyncPrettyPrinterSchemaManager sSchemaManager;
-  private Map<URL, String> sSchemaCache;
+  private Map<URL, String> mSchemaCache;
 
-  public static void ensureInitialized() {
+  private static void ensureInitialized() {
     if (sSchemaManager == null) {
       sSchemaManager = new AsyncPrettyPrinterSchemaManager(new HashMap<URL, String>());
     }
   }
 
   private AsyncPrettyPrinterSchemaManager(Map<URL, String> schemaCache) {
-    sSchemaCache = schemaCache;
+    mSchemaCache = schemaCache;
   }
 
   public synchronized void put(URL schemaUrl, String schema) {
-      sSchemaCache.put(schemaUrl, schema);
+    mSchemaCache.put(schemaUrl, schema);
   }
 
   @Nullable
   public synchronized String get(URL schemaUrl) {
-    return sSchemaCache.get(schemaUrl);
+    return mSchemaCache.get(schemaUrl);
   }
 
-  @Nullable
   public static AsyncPrettyPrinterSchemaManager getInstance() {
+    ensureInitialized();
     return sSchemaManager;
   }
 
   public void clearCache() {
-    sSchemaCache = new HashMap<>();
+    mSchemaCache = new HashMap<>();
   }
 }
--- a/stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterSchemaManager.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterSchemaManager.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.network;
+
+import javax.annotation.Nullable;
+
+import java.net.URL;
+import java.util.HashMap;
+import java.util.Map;
+
+public final class AsyncPrettyPrinterSchemaManager {
+  private static AsyncPrettyPrinterSchemaManager sSchemaManager;
+  private Map<URL, String> sSchemaCache;
+
+  public static void ensureInitialized() {
+    if (sSchemaManager == null) {
+      sSchemaManager = new AsyncPrettyPrinterSchemaManager(new HashMap<URL, String>());
+    }
+  }
+
+  private AsyncPrettyPrinterSchemaManager(Map<URL, String> schemaCache) {
+    sSchemaCache = schemaCache;
+  }
+
+  public synchronized void put(URL schemaUrl, String schema) {
+      sSchemaCache.put(schemaUrl, schema);
+  }
+
+  @Nullable
+  public synchronized String get(URL schemaUrl) {
+    return sSchemaCache.get(schemaUrl);
+  }
+
+  @Nullable
+  public static AsyncPrettyPrinterSchemaManager getInstance() {
+    return sSchemaManager;
+  }
+
+  public void clearCache() {
+    sSchemaCache = new HashMap<>();
+  }
+}
--- a/stetho/src/main/java/com/facebook/stetho/inspector/network/DownloadingAsyncPrettyPrinterFactory.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/network/DownloadingAsyncPrettyPrinterFactory.java
@@ -11,6 +11,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 
+import android.util.Log;
 import com.facebook.stetho.common.ExceptionUtil;
 import com.facebook.stetho.common.Util;
 
--- a/stetho/src/main/java/com/facebook/stetho/inspector/network/DownloadingAsyncPrettyPrinterFactory.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/network/DownloadingAsyncPrettyPrinterFactory.java
@@ -30,53 +30,77 @@ public abstract class DownloadingAsyncPrettyPrinterFactory implements AsyncPrett
       return null;
     }
     String uri = result.getSchemaUri();
-    URL schemaURL = parseURL(uri);
+    final URL schemaURL = parseURL(uri);
     if (schemaURL == null) {
-      return getErrorAsyncPrettyPrinter(headerName, headerValue);
+      String errorMessage = "[Failed to parse header: "
+          + headerName + " : " + headerValue + " ]";
+      return getErrorAsyncPrettyPrinter(errorMessage);
     } else {
-      ExecutorService  executorService = AsyncPrettyPrinterExecutorHolder.getExecutorService();
-      if (executorService == null) {
-        //last peer is unregistered...
-        return null;
-      }
-      final Future<String> response = executorService.submit(new Request(schemaURL));
-      return new AsyncPrettyPrinter() {
-        public void printTo(PrintWriter output, InputStream payload)
-            throws IOException {
-          try {
-            String schema;
+      final String schemaCache;
+      AsyncPrettyPrinterSchemaManager schemaFileManager
+          = AsyncPrettyPrinterSchemaManager.getInstance();
+      schemaCache = schemaFileManager.get(schemaURL);
+      if (schemaCache != null) {
+        return new AsyncPrettyPrinter() {
+          @Override
+          public void printTo(PrintWriter output, InputStream payload) throws IOException {
+            doPrint(output, payload, schemaCache);
+          }
+
+          @Override
+          public PrettyPrinterDisplayType getPrettifiedType() {
+            return result.getDisplayType();
+          }
+        };
+      } else {
+        ExecutorService  executorService = AsyncPrettyPrinterExecutorHolder.getExecutorService();
+        if (executorService == null) {
+          //last peer is unregistered...
+          return null;
+        }
+        final Future<String> response = executorService.submit(new Request(schemaURL));
+        return new AsyncPrettyPrinter() {
+
+          @Override
+          public void printTo(PrintWriter output, InputStream payload)
+              throws IOException {
             try {
-              schema = response.get();
+              String schema;
+              try {
+                schema = response.get();
+              } catch (ExecutionException e) {
+                Throwable cause = e.getCause();
+                if (IOException.class.isInstance(cause) ||
+                    IllegalArgumentException.class.isInstance(cause)) {
+                  doErrorPrint(
+                      output,
+                      payload,
+                      "Cannot successfully download schema: "
+                          + e.getMessage());
+                  return;
+                } else {
+                  throw e;
+                }
+              }
+              doPrint(output, payload, schema);
+            } catch (InterruptedException e) {
+              doErrorPrint(
+                  output,
+                  payload,
+                  "Encountered spurious interrupt while downloading schema for pretty printing: "
+                      + e.getMessage());
             } catch (ExecutionException e) {
               Throwable cause = e.getCause();
-              if (IOException.class.isInstance(cause)) {
-                doErrorPrint(
-                    output,
-                    payload,
-                    "Cannot successfully download schema: "
-                        + e.getMessage());
-                return;
-              } else {
-                throw e;
-              }
+              throw ExceptionUtil.propagate(cause);
             }
-            doPrint(output, payload, schema);
-          } catch (InterruptedException e) {
-            doErrorPrint(
-                output,
-                payload,
-                "Encountered spurious interrupt while downloading schema for pretty printing: "
-                    + e.getMessage());
-          } catch (ExecutionException e) {
-            Throwable cause = e.getCause();
-            throw ExceptionUtil.propagate(cause);
           }
-        }
 
-        public PrettyPrinterDisplayType getPrettifiedType() {
-          return result.getDisplayType();
-        }
-      };
+          @Override
+          public PrettyPrinterDisplayType getPrettifiedType() {
+            return result.getDisplayType();
+          }
+        };
+      }
     }
   }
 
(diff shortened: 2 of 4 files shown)


stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterSchemaFileManager.java
```diff
@@ -0,0 +1,66 @@
...
+
+  private final Context mContext;
+
+  public AsyncPrettyPrinterSchemaFileManager(Context context) {
+    mContext = context;
+  }
+
+  public synchronized OutputStream openNewSchemaCache(URL schemaUrl)
+      throws IOException {
+    String filename = FILENAME_PREFIX + cacheFileCounter++;
```
rickbrew: Let's put the `cacheFileCounter++` on the next line. Prefix and postfix operators are easy to misread when mixed with other expressions, and even though this case is pretty simple it is good to be in the habit of not using them much in that way. (in other words, prefer using them by themselves)

```
String filename = FILENAME_PREFIX + cacheFileCounter;
cacheFileCounter++;
```

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterSchemaFileManager.java
```diff
@@ -0,0 +1,66 @@
...
+import java.io.OutputStream;
+import java.net.URL;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public class AsyncPrettyPrinterSchemaFileManager {
+  private static final String TAG = "AsyncPrettyPrinterSchemaCacheManager";
+  private static final String FILENAME_PREFIX = "async-pretty-printer-schema-";
+  private static int cacheFileCounter = 0;
```
rickbrew: This should probably have an `s` prefix on it, right?

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterSchemaFileManager.java
```diff
@@ -0,0 +1,66 @@
...
+  private final Map<URL,String> mFilenameMap = Collections.synchronizedMap(
+      new HashMap<URL,String>());
+
+  private final Context mContext;
+
+  public AsyncPrettyPrinterSchemaFileManager(Context context) {
+    mContext = context;
+  }
+
+  public synchronized OutputStream openNewSchemaCache(URL schemaUrl)
```
rickbrew: we can avoid making this method synchronized, as well as `readCache`, if we instead use an AtomicInteger for the cacheFileCounter

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterSchemaFileManager.java
```diff
@@ -0,0 +1,66 @@
...
+import java.net.URL;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public class AsyncPrettyPrinterSchemaFileManager {
+  private static final String TAG = "AsyncPrettyPrinterSchemaCacheManager";
+  private static final String FILENAME_PREFIX = "async-pretty-printer-schema-";
+  private static int cacheFileCounter = 0;
+  private final Map<URL,String> mFilenameMap = Collections.synchronizedMap(
```
rickbrew: space after commas plz

```
private final Map<URL, String> ...
```

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterSchemaFileManager.java
mengdilin: Sounds good! While fixing this, I also noticed that there may be a race condition going on with mFilenameMap. Specifically, openNewSchemaCache calls put on the map while readCache calls get. These two methods are called in a block in DownloadingAsyncPrettyPrinterFactory. I am not very familiar with concurrency, so please verify if this is correct: There can be a race condition where some threads can call these two methods at the same time.

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterSchemaFileManager.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+    LogRedirector.i(TAG, "Cleaned up cache pretty printing schema files.");
+  }
+
+  public String readCache(URL schemaUrl) throws IOException {
+    synchronized (mFilenameMap) {
+      String filename = mFilenameMap.get(schemaUrl);
+      if (filename != null) {
+        InputStream in = mContext.openFileInput(filename);
+        String cacheSchema = Util.readAsUTF8(in);
```
rickbrew: we should try/finally the stream, e.g.
```
InputStream in = mContext.openFileInput(filename);
try {
  return Util.readAsUTF8(in);
} finally {
  in.close();
}
```

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterSchemaFileManager.java
```diff
@@ -0,0 +1,71 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.net.URL;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class AsyncPrettyPrinterSchemaFileManager {
```
rickbrew: can this class be made `final` ? (in other words, there are no classes deriving from it, nor do we intend for that to happen?)

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterSchemaFileManager.java
```diff
@@ -0,0 +1,71 @@
...
+  private static final String TAG = "AsyncPrettyPrinterSchemaCacheManager";
+  private static final String FILENAME_PREFIX = "async-pretty-printer-schema-";
+  private static AtomicInteger sCacheFileCounter = new AtomicInteger(0);
```
rickbrew: this should be `final`

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterSchemaFileManager.java
```diff
@@ -0,0 +1,71 @@
...
+  private final Map<URL, String> mFilenameMap = Collections.synchronizedMap(
+      new HashMap<URL,String>());
+
+  private final Context mContext;
+
+  public AsyncPrettyPrinterSchemaFileManager(Context context) {
+    mContext = context;
```
rickbrew: let's throw if null, e.g. `mContext = Util.throwIfNull(context);`

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterSchemaFileManager.java
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  public OutputStream openNewSchemaCache(URL schemaUrl)
```
rickbrew: throw if schemaUrl is null

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterSchemaFileManager.java
```diff
@@ -0,0 +1,71 @@
...
+      throws IOException {
+    synchronized (mFilenameMap) {
```
rickbrew: We don't need the `synchronized` anymore: `sCacheFileCounter` is atomic, `mFilenameMap` is already synchronized, and other than that the remaining statements don't appear to need group atomicity.

However .............. what happens if `schemaUrl` is already present in the `mFilenameMap` ? that seems like an error, right? If so, then we should 1) not bother with `AtomicInteger` for `sCacheFileCounter`, 2) not bother with making `mFilenameMap` synchronized, and 3) do our own synchronization (yeah that can mean making the methods synchronized again :-) ), and 4) check if `mFilenameMap.containsKey(schemaUrl)` before blindly putting it in there

stetho/src/main/java/com/facebook/stetho/inspector/network/DownloadingAsyncPrettyPrinterFactory.java
```diff
@@ -155,10 +156,21 @@ public Request(URL url) {
 
     @Override
     public String call() throws IOException {
-      InputStream urlStream = url.openStream();
-      String result = Util.readAsUTF8(urlStream);
-      urlStream.close();
-      return result;
+      NetworkPeerManager peerManager = NetworkPeerManager.getInstanceOrNull();
+      AsyncPrettyPrinterSchemaFileManager schemaFileManager
+          = peerManager.getAsyncPrettyPrinterSchemaFileManager();
```
rickbrew: what if `peerManager` is null ?!

longinoa:
Do we need to actually put these on disk? Why not just download them on demand and hold them in memory? Seems like it could be much simpler, especially for a debug tool.

mengdilin:
We don't need to put the schema files on disk. It is not very likely that we will be downloading a lot of different schemas. In that case, I will keep a hash map of schema url to schema string.

longinoa:
Yeah I really think putting them on disk is the wrong thing to do. by putting them on disk you are making an assumption that shouldn't be made which is that they will never change. While this may be true for one implementation you are assuming that all developers will want to keep the schema file locked.

stetho/src/main/java/com/facebook/stetho/inspector/network/DownloadingAsyncPrettyPrinterFactory.java
```diff
@@ -149,18 +150,31 @@ public PrettyPrinterDisplayType getDisplayType() {
...
-      urlStream.close();
-      return result;
+    public String call() throws IOException, IllegalArgumentException {
+      NetworkPeerManager peerManager = NetworkPeerManager.getInstanceOrNull();
+      if (peerManager == null) {
+        throw new IllegalArgumentException("NetworkPeerManager cannot be null");
+      }
+      AsyncPrettyPrinterSchemaManager schemaFileManager
+          = peerManager.getPrettyPrinterSchemaManager();
+      String cacheSchema = schemaFileManager.readCache(url);
```
longinoa: Reading the cache should be done before invoking the `Executor` The reason you need the executor model is because of the fact that you are scheduling the pull from the network on the executor. Now that you are pulling in from an in memory cache there is no need to thread hop.

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterSchemaManager.java
```diff
@@ -0,0 +1,42 @@
...
+
+import javax.annotation.Nullable;
+
+import java.io.IOException;
+import java.net.URL;
+import java.util.HashMap;
+import java.util.Map;
+
+public final class AsyncPrettyPrinterSchemaManager {
+  private final Map<URL, String> mSchemaCache = new HashMap<>();
```
longinoa: should we clear when the client disconnects?

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkPeerManager.java
```diff
@@ -57,6 +58,10 @@ public void setPrettyPrinterInitializer(AsyncPrettyPrinterInitializer initialize
     mPrettyPrinterInitializer = Util.throwIfNull(initializer);
   }
 
+  public AsyncPrettyPrinterSchemaManager getPrettyPrinterSchemaManager() {
```
longinoa: Why not make this a singleton instance and just have `AsyncPrettyPrinterSchemaManager.getInstance()`

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterSchemaManager.java
```diff
@@ -0,0 +1,42 @@
...
+
+  public synchronized void storeNewSchemaCache(URL schemaUrl, String schema) {
+    if (schemaUrl == null || schema == null) {
```
longinoa: why is this true, why not just expose a `Map` like interface?

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterSchemaManager.java
```diff
@@ -0,0 +1,49 @@
...
+      sSchemaManager = new AsyncPrettyPrinterSchemaManager(new HashMap<URL, String>());
+    }
+  }
+
+  private AsyncPrettyPrinterSchemaManager(Map<URL, String> schemaCache) {
+    sSchemaCache = schemaCache;
+  }
+
+  public synchronized void put(URL schemaUrl, String schema) {
+      sSchemaCache.put(schemaUrl, schema);
```
longinoa: whitespace

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterSchemaManager.java
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  @Nullable
+  public synchronized String get(URL schemaUrl) {
+    return sSchemaCache.get(schemaUrl);
+  }
+
+  @Nullable
```
longinoa: Why is this `@Nullable` does anyone need to get a null instance? having a static `ensureInitialized` and `getInstance()` is very bizarre.

I would either 
1) make all functions in here public static and internally call `getInstance`
2) Just use `getInstance` and just ensure that it is non null

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterSchemaManager.java
```diff
@@ -0,0 +1,49 @@
...
+
+import javax.annotation.Nullable;
+
+import java.net.URL;
+import java.util.HashMap;
+import java.util.Map;
+
+public final class AsyncPrettyPrinterSchemaManager {
+  private static AsyncPrettyPrinterSchemaManager sSchemaManager;
+  private Map<URL, String> mSchemaCache;
```
longinoa: `@GuardedBy("this")`

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterSchemaManager.java
```diff
@@ -0,0 +1,49 @@
...
+  public synchronized String get(URL schemaUrl) {
+    return mSchemaCache.get(schemaUrl);
+  }
+
+  public static AsyncPrettyPrinterSchemaManager getInstance() {
+    ensureInitialized();
+    return sSchemaManager;
+  }
+
+  public void clearCache() {
```
longinoa: `synchronized`

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterSchemaManager.java
longinoa: I just realized you are threading the `URL` all the way through. `URL#equals()` will actually perform dns lookup to verify the hosts resolve to the same address. It would probably better to take the String representation of the url and use that. Its okay to keep passing in URL but I would make this a `Map<String, String>` and use `URL.toString()` when calling `put` or `get`

mengdilin: wow! This is good to know. I will change it to String :D

stetho/src/main/java/com/facebook/stetho/inspector/network/DownloadingAsyncPrettyPrinterFactory.java
```diff
@@ -30,53 +30,77 @@ public AsyncPrettyPrinter getInstance(final String headerName, final String head
...
-        return null;
-      }
-      final Future<String> response = executorService.submit(new Request(schemaURL));
-      return new AsyncPrettyPrinter() {
-        public void printTo(PrintWriter output, InputStream payload)
-            throws IOException {
-          try {
-            String schema;
+      final String schemaCache;
+      AsyncPrettyPrinterSchemaManager schemaFileManager
```
longinoa: long style note:

You should only create the `AsyncPrettyPrinter` once. Imagine you want to add a prefix in `doPrint`. Now you have to know that you need to modify it in 2 places. Prefer a paradigm like the following

```
String schema = AsyncPretyPrinterSchemaManager.getInstance().get(schemaUrl);
ExecutorService  executorService = AsyncPrettyPrinterExecutorHolder.getExecutorService();
if (schema == null && executorService != null) {
  final Future<String> response = executorService.submit(new Request(schemaURL));
  try {
    schema = response.get();
  } catch (...) { ... }
}
if (schema != null) {
  return new AsyncPrettyPrinter() {...};
} else {
  return null;
}
```

This really drops your line count down, makes it more readable, and leaves less room for bugs. The other option is to keep a `String errorMessage` object as null and send that in the  `doErrorPrint` like you are doing already.

mengdilin: I agree that returning different pretty printers throughout this method is not the best idea. I thought about consolidating all the returned pretty printers into one. The major problem that prevented me from being able to do this is response.get() must be called inside PrettyPrinter's printTo method. By design, only printTo is allowed to block for a specified amount of time. This condition somewhat ties my hands to refactor the codes...

rickbrew:
So what's the word on this?

mengdilin:
Sorry for the late response! Been busy with school stuff! I can finish this PR, but it will take a while for me to push new updates and respond to code reviews given I have school works. Does that sound good or is this feature urgently needed?

jasta:
This isn't urgent, no.  We can commandeer as well if you're too busy.  Just let us know.

mengdilin:
Hi guys, I have been preoccupied with school works, and I was hoping I could finish this over fall break and Thanksgiving break. Unfortunately, there are stuff I have to do during the breaks as well, so I can no longer follow through with this :( I am sorry

jasta:
Actually I don't even think this caching is meaningful for developer builds.  It really has never caused an issue for me and I use the feature all the time...

Issue #255, opened 2015-08-13
NetworkPeerManager needs to start AsyncPrettyPrinterExecutorHolder


longinoa:
Currently in `NetworkPeerManager.PeersRegisteredListener#onLastPeerUnregistered` we call       `AsyncPrettyPrinterExecutorHolder.sExecutorService.shutdownNow();` but this is never created in `onFirstPeerRegistered`

PR #256, opened 2015-08-14
Fix registration flow for Async Pretty Printer

mengdilin:
Right now, pretty printers are instantiated and registered at the start up path. This PR moves these steps to onFirstPeerRegistered() for NetworkPeerManager's listener.

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterInitializer.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterInitializer.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.network;
+
+/**
+ * Interface that is called if AsyncPrettyPrinterRegistry is unpopulated when
+ * the first peer connects to Stetho. It is responsible for registering header
+ * names and their corresponding pretty printers
+ */
+public interface AsyncPrettyPrinterInitializer {
+
+  /**
+   * Populates AsyncPrettyPrinterRegistry with header names and their corresponding pretty
+   * printers. This is responsible for registering all {@link AsyncPrettyPrinter} to the
+   * provided registry.
+   * @param registry
+   */
+  void populatePrettyPrinters(AsyncPrettyPrinterRegistry registry);
+
+}
--- a/stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkEventReporterImpl.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkEventReporterImpl.java
@@ -158,7 +158,7 @@ public class NetworkEventReporterImpl implements NetworkEventReporter {
   }
 
   @Nullable
-   static AsyncPrettyPrinter initAsyncPrettyPrinterForResponse(
+  private static AsyncPrettyPrinter initAsyncPrettyPrinterForResponse(
       InspectorResponse response,
       NetworkPeerManager peerManager) {
     AsyncPrettyPrinterRegistry registry = peerManager.getAsyncPrettyPrinterRegistry();
@@ -184,17 +184,20 @@ public class NetworkEventReporterImpl implements NetworkEventReporter {
     }
   }
 
+  //@VisibleForTesting
   @Nullable
-  private static AsyncPrettyPrinter createPrettyPrinterForResponse(
+  static AsyncPrettyPrinter createPrettyPrinterForResponse(
       InspectorResponse response,
-      AsyncPrettyPrinterRegistry registry) {
-    for (int i = 0, count = response.headerCount(); i < count; i++) {
-      AsyncPrettyPrinterFactory factory = registry.lookup(response.headerName(i));
-      if (factory != null) {
-        AsyncPrettyPrinter asyncPrettyPrinter = factory.getInstance(
-            response.headerName(i),
-            response.headerValue(i));
-        return asyncPrettyPrinter;
+      @Nullable AsyncPrettyPrinterRegistry registry) {
+    if (registry != null) {
+      for (int i = 0, count = response.headerCount(); i < count; i++) {
+        AsyncPrettyPrinterFactory factory = registry.lookup(response.headerName(i));
+        if (factory != null) {
+          AsyncPrettyPrinter asyncPrettyPrinter = factory.getInstance(
+              response.headerName(i),
+              response.headerValue(i));
+          return asyncPrettyPrinter;
+        }
       }
     }
     return null;
--- a/stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkPeerManager.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkPeerManager.java
@@ -12,7 +12,7 @@ package com.facebook.stetho.inspector.network;
 import javax.annotation.Nullable;
 
 import android.content.Context;
-
+import com.facebook.stetho.common.Util;
 import com.facebook.stetho.inspector.helper.ChromePeerManager;
 import com.facebook.stetho.inspector.helper.PeersRegisteredListener;
 
@@ -20,7 +20,8 @@ public class NetworkPeerManager extends ChromePeerManager {
   private static NetworkPeerManager sInstance;
 
   private final ResponseBodyFileManager mResponseBodyFileManager;
-  private final AsyncPrettyPrinterRegistry mAsyncPrettyPrinterRegistry;
+  private AsyncPrettyPrinterInitializer mPrettyPrinterInitializer;
+  private AsyncPrettyPrinterRegistry mAsyncPrettyPrinterRegistry;
 
   @Nullable
   public static synchronized NetworkPeerManager getInstanceOrNull() {
@@ -36,23 +37,33 @@ public class NetworkPeerManager extends ChromePeerManager {
     return sInstance;
   }
 
-  public NetworkPeerManager(ResponseBodyFileManager responseBodyFileManager) {
+  public NetworkPeerManager(
+      ResponseBodyFileManager responseBodyFileManager) {
     mResponseBodyFileManager = responseBodyFileManager;
     setListener(mTempFileCleanup);
-    mAsyncPrettyPrinterRegistry = new AsyncPrettyPrinterRegistry();
   }
 
   public ResponseBodyFileManager getResponseBodyFileManager() {
     return mResponseBodyFileManager;
   }
 
+  @Nullable
   public AsyncPrettyPrinterRegistry getAsyncPrettyPrinterRegistry() {
     return mAsyncPrettyPrinterRegistry;
+  }
+
+  public void setPrettyPrinterInitializer(AsyncPrettyPrinterInitializer initializer) {
+    Util.throwIfNotNull(mPrettyPrinterInitializer);
+    mPrettyPrinterInitializer = Util.throwIfNull(initializer);
   }
 
   private final PeersRegisteredListener mTempFileCleanup = new PeersRegisteredListener() {
     @Override
     protected void onFirstPeerRegistered() {
+      if (mAsyncPrettyPrinterRegistry == null && mPrettyPrinterInitializer != null) {
+        mAsyncPrettyPrinterRegistry = new AsyncPrettyPrinterRegistry();
+        mPrettyPrinterInitializer.populatePrettyPrinters(mAsyncPrettyPrinterRegistry);
+      }
       mResponseBodyFileManager.cleanupFiles();
     }
 
--- a/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Network.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Network.java
@@ -14,12 +14,13 @@ import java.util.List;
 
 import android.content.Context;
 
+import com.facebook.stetho.common.Util;
 import com.facebook.stetho.inspector.jsonrpc.JsonRpcException;
 import com.facebook.stetho.inspector.jsonrpc.JsonRpcPeer;
 import com.facebook.stetho.inspector.jsonrpc.JsonRpcResult;
 import com.facebook.stetho.inspector.jsonrpc.protocol.JsonRpcError;
+import com.facebook.stetho.inspector.network.AsyncPrettyPrinterInitializer;
 import com.facebook.stetho.inspector.network.NetworkPeerManager;
-import com.facebook.stetho.inspector.network.AsyncPrettyPrinterRegistry;
 import com.facebook.stetho.inspector.network.ResponseBodyData;
 import com.facebook.stetho.inspector.network.ResponseBodyFileManager;
 import com.facebook.stetho.inspector.protocol.ChromeDevtoolsDomain;
@@ -87,8 +88,16 @@ public class Network implements ChromeDevtoolsDomain {
     return response;
   }
 
-  public AsyncPrettyPrinterRegistry getAsyncPrettyPrinterRegistry() {
-    return mNetworkPeerManager.getAsyncPrettyPrinterRegistry();
+  /**
+   * Method that allows callers to provide an {@link AsyncPrettyPrinterInitializer} that is
+   * responsible for registering all
+   * {@link com.facebook.stetho.inspector.network.AsyncPrettyPrinter}.
+   * Note that AsyncPrettyPrinterInitializer cannot be null and can only be set once.
+   * @param initializer
+   */
+  public void setPrettyPrinterInitializer(AsyncPrettyPrinterInitializer initializer) {
+    Util.throwIfNull(initializer);
+    mNetworkPeerManager.setPrettyPrinterInitializer(initializer);
   }
 
   private static class GetResponseBodyResponse implements JsonRpcResult {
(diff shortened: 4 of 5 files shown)


stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkPeerManager.java
```diff
@@ -50,16 +53,23 @@ public AsyncPrettyPrinterRegistry getAsyncPrettyPrinterRegistry() {
...
+      mAsyncPrettyPrinterRegistry = new AsyncPrettyPrinterRegistry();
+      mPrettyPrinterRegister.registerPrettyPrinters(mAsyncPrettyPrinterRegistry);
       mResponseBodyFileManager.cleanupFiles();
     }
 
     @Override
     protected void onLastPeerUnregistered() {
       mResponseBodyFileManager.cleanupFiles();
-      AsyncPrettyPrinterExecutorHolder.sExecutorService.shutdownNow();
+      AsyncPrettyPrinterExecutorHolder.getExecutorService().shutdownNow();
```
longinoa: This should be a first order function on `AsyncPrettyPrinterExecutorHolder` calling `shutdownNow` will not set the executor to null in `AsyncPrettyPrinterExecutorHolder`.

stetho/src/main/java/com/facebook/stetho/inspector/network/DownloadingAsyncPrettyPrinterFactory.java
```diff
@@ -22,7 +22,7 @@
 public abstract class DownloadingAsyncPrettyPrinterFactory implements AsyncPrettyPrinterFactory {
 
   private final static ExecutorService sExecutorService =
-      AsyncPrettyPrinterExecutorHolder.sExecutorService;
+      AsyncPrettyPrinterExecutorHolder.getExecutorService();
```
longinoa: This doesn't work you are caching the first value of `getExecutorService()`. If you are calling `shutdownNow()` and the re-assigning the value in `AsyncPrettyPrinterExecutorHolder` then you will not have the new executor here.

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkEventReporterImpl.java
```diff
@@ -185,7 +185,7 @@ static AsyncPrettyPrinter initAsyncPrettyPrinterForResponse(
   }
 
   @Nullable
-  private static AsyncPrettyPrinter createPrettyPrinterForResponse(
+  static AsyncPrettyPrinter createPrettyPrinterForResponse(
```
longinoa: `//@VisibleForTesting`

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
```diff
@@ -47,7 +47,7 @@
   private final Map<String, AsyncPrettyPrinter> mRequestIdMap = Collections.synchronizedMap(
       new HashMap<String, AsyncPrettyPrinter>());
   private final ExecutorService sExecutorService =
-      AsyncPrettyPrinterExecutorHolder.sExecutorService;
+      AsyncPrettyPrinterExecutorHolder.getExecutorService();
```
longinoa: same issue with caching here.

longinoa:
Can you split this out into 2 PRs one that solves #255 the other that fixes the registration flow.

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterRegister.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.network;
+
+/**
+ * Interface to register header names and their corresponding pretty printers
```
longinoa: You should state a bit more about when and how this is called. Something like:
Interface that is called each time the first peer connects to stetho.

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterRegister.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+public interface AsyncPrettyPrinterRegister {
+
+  /**
+   * Populates AsyncPrettyPrinterRegistry with header names and their corresponding pretty printers
```
longinoa: This is responsible for registering all {@link AsyncPrettyPrinter} to the provided registry.

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterRegister.java
```diff
@@ -0,0 +1,23 @@
...
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.network;
+
+/**
+ * Interface to register header names and their corresponding pretty printers
+ */
+public interface AsyncPrettyPrinterRegister {
```
longinoa: `AsyncPrettyPrinterInitializer`

the hardest problem in computer science is naming things ...

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterRegister.java
```diff
@@ -0,0 +1,23 @@
...
+
+  /**
+   * Populates AsyncPrettyPrinterRegistry with header names and their corresponding pretty printers
+   * @param registry
+   */
+  void registerPrettyPrinters(AsyncPrettyPrinterRegistry registry);
```
longinoa: `populatePrettyPrinterRegistry`

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkPeerManager.java
```diff
@@ -50,9 +53,15 @@ public AsyncPrettyPrinterRegistry getAsyncPrettyPrinterRegistry() {
...
   }
 
+  public void setPrettyPrinterRegister(@NonNull AsyncPrettyPrinterRegister register) {
+    mPrettyPrinterRegister = register;
+  }
+
   private final PeersRegisteredListener mTempFileCleanup = new PeersRegisteredListener() {
     @Override
     protected void onFirstPeerRegistered() {
+      mAsyncPrettyPrinterRegistry = new AsyncPrettyPrinterRegistry();
```
longinoa: Reading this a second time this is a bit bizarre. Why is this done every time the first client connects? Should this be done once per process invocation? 

ie something like
```
if (mAsyncPrettyPrinterRegistry == null) { 
  mAsyncPrettyPrinterRegistry = new AsyncPrettyPrinterRegistry();
  mPrettyPrinterRegister.registerPrettyPrinters(mAsyncPrettyPrinterRegistry);
}
```

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkPeerManager.java
```diff
@@ -50,9 +53,15 @@ public AsyncPrettyPrinterRegistry getAsyncPrettyPrinterRegistry() {
     return mAsyncPrettyPrinterRegistry;
   }
 
+  public void setPrettyPrinterRegister(@NonNull AsyncPrettyPrinterRegister register) {
+    mPrettyPrinterRegister = register;
```
longinoa: throw if `mPrettyPrinterRegister` is not null

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkPeerManager.java
```diff
@@ -50,9 +53,20 @@ public AsyncPrettyPrinterRegistry getAsyncPrettyPrinterRegistry() {
     return mAsyncPrettyPrinterRegistry;
   }
 
+  public void setPrettyPrinterInitializer(@NonNull AsyncPrettyPrinterInitializer initializer) {
+    if (initializer == null) {
```
rickbrew: you can just use `Util.throwIfNull(initializer)`. Don't really need the `@NonNull` annotation.

stetho/src/main/java/com/facebook/stetho/inspector/network/DefaultPrettyPrinterInitializer.java
```diff
@@ -0,0 +1,20 @@
...
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.network;
+
+public class DefaultPrettyPrinterInitializer implements AsyncPrettyPrinterInitializer {
```
rickbrew: This implementation doesn't do anything at all... why not just use `null` later on instead of instantiating this? Then we can remove this class entirely

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkPeerManager.java
```diff
@@ -36,10 +38,11 @@ public static synchronized NetworkPeerManager getOrCreateInstance(Context contex
     return sInstance;
   }
 
-  public NetworkPeerManager(ResponseBodyFileManager responseBodyFileManager) {
+  public NetworkPeerManager(
+      ResponseBodyFileManager responseBodyFileManager) {
     mResponseBodyFileManager = responseBodyFileManager;
+    mPrettyPrinterInitializer = new DefaultPrettyPrinterInitializer();
```
rickbrew: related to my comment about removing the `DefaultPrettyPrinterInitializer` class: Another reason why I'd rather see us use `null` is that we're just allocating an object that will either be replaced (via `setPrettyPrinterInitializer`) or that will do nothing. So why not just remove it entirely and use `null` instead :)

stetho/src/main/java/com/facebook/stetho/inspector/network/DefaultPrettyPrinterInitializer.java
mengdilin: Sounds good. I will instantiate pretty printer registry onFirstPeerRegistered and if PrettyPrinterInitializer is not null, I will populate the pretty printer registry

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkPeerManager.java
```diff
@@ -50,9 +51,19 @@ public AsyncPrettyPrinterRegistry getAsyncPrettyPrinterRegistry() {
     return mAsyncPrettyPrinterRegistry;
   }
 
+  public void setPrettyPrinterInitializer(AsyncPrettyPrinterInitializer initializer) {
+    mPrettyPrinterInitializer = Util.throwIfNull(initializer);
```
longinoa: throw if `mPrettyPrinterInitializer` is null. Basically only allow `setPrettyPrinterInitializer` to be called once.

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkPeerManager.java
```diff
@@ -50,9 +51,19 @@ public AsyncPrettyPrinterRegistry getAsyncPrettyPrinterRegistry() {
...
+  }
+
   private final PeersRegisteredListener mTempFileCleanup = new PeersRegisteredListener() {
     @Override
     protected void onFirstPeerRegistered() {
+      if (mAsyncPrettyPrinterRegistry == null) {
```
longinoa: you should shortcut this. 
`if (mAsyncPrettyPrinterRegistry == null && mPrettyPrinterInitializer != null)`

There is no need to reinitialize if you have already done it. See my note in `#setPrettyPrinterInitializer()`

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkEventReporterImpl.java
```diff
@@ -184,8 +184,9 @@ static AsyncPrettyPrinter initAsyncPrettyPrinterForResponse(
     }
   }
 
+  //@VisibleForTesting
   @Nullable
-  private static AsyncPrettyPrinter createPrettyPrinterForResponse(
+  static AsyncPrettyPrinter createPrettyPrinterForResponse(
       InspectorResponse response,
       AsyncPrettyPrinterRegistry registry) {
```
longinoa: `@Nullable`

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Network.java
```diff
@@ -87,8 +88,8 @@ private GetResponseBodyResponse readResponseBody(String requestId)
     return response;
   }
 
-  public AsyncPrettyPrinterRegistry getAsyncPrettyPrinterRegistry() {
-    return mNetworkPeerManager.getAsyncPrettyPrinterRegistry();
+  public void setPrettyPrinterInitializer(@NonNull AsyncPrettyPrinterInitializer initializer) {
```
longinoa: No real need to use `NonNull` here just do the `Util.throwIfNull(initializer);`

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkPeerManager.java
```diff
@@ -36,10 +37,10 @@ public static synchronized NetworkPeerManager getOrCreateInstance(Context contex
...
 
-  public NetworkPeerManager(ResponseBodyFileManager responseBodyFileManager) {
+  public NetworkPeerManager(
+      ResponseBodyFileManager responseBodyFileManager) {
     mResponseBodyFileManager = responseBodyFileManager;
     setListener(mTempFileCleanup);
-    mAsyncPrettyPrinterRegistry = new AsyncPrettyPrinterRegistry();
   }
 
   public ResponseBodyFileManager getResponseBodyFileManager() {
```
longinoa: `@Nullable` after the change to `onFirstPeerRegistered`

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkPeerManager.java
```diff
@@ -50,9 +51,19 @@ public AsyncPrettyPrinterRegistry getAsyncPrettyPrinterRegistry() {
     return mAsyncPrettyPrinterRegistry;
```
longinoa: `@Nullable` after the change to `onFirstPeerRegistered`

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkPeerManager.java
mengdilin: Slightly confused here. Do you mean throw if mPrettyPrinterInitializer is not null?

longinoa: yeah basically make the set call non idempotent. This way you can do things like only fire the initializer once. Unless im mistaken there is no real need for the client process to change what pretty printers are registered during a single process invocation?

mengdilin: That makes sense :)  There is no real need to change pretty printers during a single process invocation.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Network.java
```diff
@@ -87,8 +89,9 @@ private GetResponseBodyResponse readResponseBody(String requestId)
     return response;
   }
 
-  public AsyncPrettyPrinterRegistry getAsyncPrettyPrinterRegistry() {
-    return mNetworkPeerManager.getAsyncPrettyPrinterRegistry();
+  public void setPrettyPrinterInitializer(AsyncPrettyPrinterInitializer initializer) {
```
longinoa: please add javadoc here

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterInitializer.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.network;
+
+/**
+ * Interface that is called each time the first peer connects
```
longinoa: update this based on the latest change

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Network.java
```diff
@@ -14,12 +14,14 @@
 
 import android.content.Context;
 
+import android.support.annotation.NonNull;
```
rickbrew: also, if we're not using this annotation anymore then we can remove it

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Network.java
```diff
@@ -87,8 +89,17 @@ private GetResponseBodyResponse readResponseBody(String requestId)
     return response;
   }
 
-  public AsyncPrettyPrinterRegistry getAsyncPrettyPrinterRegistry() {
-    return mNetworkPeerManager.getAsyncPrettyPrinterRegistry();
+  /**
+   * Method that allows callers to provide an
+   * {@link com.facebook.stetho.inspector.network.AsyncPrettyPrinterInitializer} that is
```
rickbrew: fully qualified class name isn't needed because we have an import for it. So the javadoc can say, `Method that allows callers to provide an {@link AsyncPrettyPrinterInitialier} that is...`

longinoa:
This looks good to me, @rickbrew anything else you want to add?

rickbrew:
Yeah I think we're good here. Squash 'n merge

mengdilin:
yayayay!

PR #257, opened 2015-08-14
Init executor service for pretty printers on first peer registered

mengdilin:


stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterExecutorHolder.java
```diff
@@ -16,5 +16,21 @@
  * A holder class for the executor service used for pretty printing related tasks
  */
 class AsyncPrettyPrinterExecutorHolder {
-  public static final ExecutorService sExecutorService = Executors.newCachedThreadPool();
+  private static ExecutorService sExecutorService;
+
+  public static void init() {
```
rickbrew: This shouldn't be public. I would also name it something like `ensureInitialized`

rickbrew: In fact, I'm not sure we should have this stashed into a static class. Right now, `ResponseBodyFileManager` has a dependency on the fact that `DownloadingAsyncPrettyPrinterFactory` has called `init` before it calls `getExecutorService`. Why not just have the owner of `ResponseBodyFileManager` create the `ExecutorService` and pass it down?

mengdilin: The ExecutorService for async pretty printing is being used at two instances. First instance occurs when Stetho receives a network response. At that point, Stetho will try to create a pretty printer and ExecutorService will start downloading schema if pretty printer exists for that response (from DownloadingAsyncPrettyPrinterFactory's getInstance method). Second instance occurs when the user clicks on the Response tab in Network to view the response body. Then ExecutorService is used again in ResponseBodyFileManager to asynchronously pretty print response. The first instance occurs before the second instance. I can't pass the executor service from ResponeBodyFileManager to DownloadingAsyncPrettyPrinterFactory.

stetho/src/main/java/com/facebook/stetho/inspector/network/DownloadingAsyncPrettyPrinterFactory.java
```diff
@@ -21,8 +21,7 @@
  */
 public abstract class DownloadingAsyncPrettyPrinterFactory implements AsyncPrettyPrinterFactory {
 
-  private final static ExecutorService sExecutorService =
-      AsyncPrettyPrinterExecutorHolder.sExecutorService;
+  private static ExecutorService sExecutorService;
```
longinoa: delete this, and always call `AsyncPrettyPrinterExecutorHolder.getExecutorService()`

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterExecutorHolder.java
```diff
@@ -16,5 +16,21 @@
...
-  public static final ExecutorService sExecutorService = Executors.newCachedThreadPool();
+  private static ExecutorService sExecutorService;
+
+  static void ensureInitialized() {
+    if (sExecutorService == null) {
+      sExecutorService = Executors.newCachedThreadPool();
+    }
+  }
+
+  public static ExecutorService getExecutorService() {
```
longinoa: `@Nullable`

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
```diff
@@ -43,11 +43,10 @@
   private static final int PRETTY_PRINT_TIMEOUT_SEC = 10;
 
   private final Context mContext;
-
   private final Map<String, AsyncPrettyPrinter> mRequestIdMap = Collections.synchronizedMap(
       new HashMap<String, AsyncPrettyPrinter>());
-  private final ExecutorService sExecutorService =
-      AsyncPrettyPrinterExecutorHolder.sExecutorService;
+
+  private ExecutorService sExecutorService;
```
longinoa: Same here, no need to keep a static reference to something you can just call a getter for. Especially when it is to a mutable object

longinoa:
Please make your commit message more descriptive. The top line should be what the diff does in one sentence, not referencing the issue.

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterExecutorHolder.java
mengdilin: the return value cannot be null here since I called "ensureInitialized" before returning an executor service.

longinoa: It actually technically can be null. There is no synchronization on `sExecutorService`, and thus someone can call `shutdown` immediatley after the call to `ensureInitialized()` This will result in `sExecutorService` being null and crashing the app where it is used.

Additionally the correct thing to do, given the current design, is to not call `ensureInitialized` here for the case I mentioned above. If the last peer has disconnected you do not want to re-initialize the executor.

stetho/src/main/java/com/facebook/stetho/inspector/network/ResponseBodyFileManager.java
```diff
@@ -95,6 +92,7 @@ private String prettyPrintContentWithTimeOut(
     AsyncPrettyPrintingCallable prettyPrintingCallable = new AsyncPrettyPrintingCallable(
         in,
         asyncPrettyPrinter);
+    ExecutorService sExecutorService = AsyncPrettyPrinterExecutorHolder.getExecutorService();
```
longinoa: nit `executorService` this is no longer static.

stetho/src/main/java/com/facebook/stetho/inspector/network/DownloadingAsyncPrettyPrinterFactory.java
```diff
@@ -36,6 +33,7 @@ public AsyncPrettyPrinter getInstance(final String headerName, final String head
     if (schemaURL == null) {
       return getErrorAsyncPrettyPrinter(headerName, headerValue);
     } else {
+      ExecutorService  sExecutorService = AsyncPrettyPrinterExecutorHolder.getExecutorService();
```
longinoa: `executorService`

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterExecutorHolder.java
```diff
@@ -16,5 +18,21 @@
  * A holder class for the executor service used for pretty printing related tasks
  */
 class AsyncPrettyPrinterExecutorHolder {
-  public static final ExecutorService sExecutorService = Executors.newCachedThreadPool();
+  private static ExecutorService sExecutorService;
+
+  static void ensureInitialized() {
```
longinoa: this can be `public` your class is package protected so anything that is public is only visible to the package

stetho/src/main/java/com/facebook/stetho/inspector/network/AsyncPrettyPrinterExecutorHolder.java
```diff
@@ -16,5 +18,21 @@
  * A holder class for the executor service used for pretty printing related tasks
  */
 class AsyncPrettyPrinterExecutorHolder {
```
rickbrew: this class should also be `final` and have a private, empty, and parameterless constructor

PR #260, opened 2015-08-18
Add status code checking for schema download

mengdilin:
Right now, stetho attempts to download schema content with the url provided without checking status code. This PR adds the functionality of checking status code before passing the schema content to pretty printer.

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/network/DownloadingAsyncPrettyPrinterFactory.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/network/DownloadingAsyncPrettyPrinterFactory.java
@@ -3,6 +3,7 @@ package com.facebook.stetho.inspector.network;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.PrintWriter;
+import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.concurrent.Callable;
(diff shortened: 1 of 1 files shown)


stetho/src/main/java/com/facebook/stetho/inspector/network/DownloadingAsyncPrettyPrinterFactory.java
```diff
@@ -157,10 +158,18 @@ public Request(URL url) {
 
     @Override
     public String call() throws IOException {
-      InputStream urlStream = url.openStream();
-      String result = Util.readAsUTF8(urlStream);
-      urlStream.close();
-      return result;
+      HttpURLConnection.setFollowRedirects(false);
```
longinoa: We should follow redirects

longinoa: Actually I am going to ammend that... you shouldnt change this value :)

stetho/src/main/java/com/facebook/stetho/inspector/network/DownloadingAsyncPrettyPrinterFactory.java
```diff
@@ -157,10 +158,18 @@ public Request(URL url) {
...
+      HttpURLConnection connection = (HttpURLConnection)url.openConnection();
+      int statusCode = connection.getResponseCode();
+      if (statusCode != 200) {
+        throw new IOException("Got status code: " + statusCode + " while downloading schema");
```
longinoa: Where does this eventually get output? Also may want to include the url?

mengdilin: The IOException will be caught when pretty printer's printTo method is trying to call "response.get()" to retrieve schema. If an IOException is thrown from here, printTo will catch it and print out the error message following with the original response body in the Response tab of Network.

longinoa:
squash it il lmerge

PR #261, opened 2015-08-18
Added the ability to inspect an element's style

ichub:
This pull request adds the ability to view the style of an element that is selected in the elements tree using the same UI as what is shown in the styles tab when inspecting an element of a webpage.

Some gotchas:

* The names of style properties shown are not copied verbatim from the names of fields in the `View` class, but are translated to a format that Chrome understands. For example, `View` has a field called `paddingLeft`, but in order for that to be rendered correctly in Chrome, it needs to be transmitted as `padding-left`. 

* Default values are not transmitted, unless they are of an enum type, in which case they are *always* transmitted.

* Due to inconsistencies between annotations present on certain Android SDK versions, some fields that are enums are not translated into their string representation. This is unfortunate, but the only way to get around it is by special-casing those one or two cases, which is probably not viable long-term.

Some things that would be nice to add in the future:

* Inferring some way the element's style 'class', so that the styles aren't all aggregated into one blob. Same with style inheritance.

* In Chrome you can edit these CSS properties, but here you can't. Maybe a future project? Would probably relatively simple to add.

Style inspection of Stetho's 'Settings' button:

<p align="center">

</p>

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/AbstractChainedDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/AbstractChainedDescriptor.java
@@ -144,4 +144,14 @@ public abstract class AbstractChainedDescriptor<E> extends Descriptor implements
   protected void onSetAttributesAsText(E element, String text) {
     mSuper.setAttributesAsText(element, text);
   }
+
+  @Override
+  @SuppressWarnings("unchecked")
+  public final void getStyles(Object element, StyleAccumulator accumulator) {
+    mSuper.getStyles(element, accumulator);
+    onGetStyles((E) element, accumulator);
+  }
+
+  protected void onGetStyles(E element, StyleAccumulator accumulator) {
+  }
 }
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
@@ -140,6 +140,12 @@ public final class Document extends ThreadBoundProxy {
     mDocumentProvider.setAttributesAsText(element, text);
   }
 
+  public void getElementStyles(Object element, StyleAccumulator styleAccumulator) {
+    NodeDescriptor nodeDescriptor = getNodeDescriptor(element);
+
+    nodeDescriptor.getStyles(element, styleAccumulator);
+  }
+
   public DocumentView getDocumentView() {
     verifyThreadAccess();
     return mShadowDocument;
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/NodeDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/NodeDescriptor.java
@@ -33,4 +33,6 @@ public interface NodeDescriptor extends ThreadBound {
   void getAttributes(Object element, AttributeAccumulator attributes);
 
   void setAttributesAsText(Object element, String text);
+
+  void getStyles(Object element, StyleAccumulator accumulator);
 }
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/Origin.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/Origin.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.elements;
+
+import com.facebook.stetho.json.annotation.JsonValue;
+
+public enum Origin {
+  INJECTED("injected"),
+  USER_AGENT("user-agent"),
+  INSPECTOR("inspector"),
+  REGULAR("regular");
+
+  private final String mValue;
+
+  Origin(String value) {
+    mValue = value;
+  }
+
+  @JsonValue
+  public String getProtocolValue() {
+    return mValue;
+  }
+}
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/DialogFragmentDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/DialogFragmentDescriptor.java
@@ -23,6 +23,7 @@ import com.facebook.stetho.inspector.elements.ChainedDescriptor;
 import com.facebook.stetho.inspector.elements.Descriptor;
 import com.facebook.stetho.inspector.elements.DescriptorMap;
 import com.facebook.stetho.inspector.elements.NodeType;
+import com.facebook.stetho.inspector.elements.StyleAccumulator;
 
 import javax.annotation.Nullable;
 
@@ -124,5 +125,9 @@ final class DialogFragmentDescriptor
     }
 
     return null;
+  }
+
+  @Override
+  public void getStyles(Object element, StyleAccumulator styles) {
   }
 }
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
@@ -10,18 +10,94 @@
 package com.facebook.stetho.inspector.elements.android;
 
 import android.view.View;
+import android.view.ViewDebug;
+import com.facebook.stetho.common.LogUtil;
 import com.facebook.stetho.common.StringUtil;
 import com.facebook.stetho.common.android.ResourcesUtil;
-import com.facebook.stetho.inspector.elements.AttributeAccumulator;
 import com.facebook.stetho.inspector.elements.AbstractChainedDescriptor;
+import com.facebook.stetho.inspector.elements.AttributeAccumulator;
+import com.facebook.stetho.inspector.elements.StyleAccumulator;
+import com.facebook.stetho.inspector.helper.IntegerFormatter;
 
 import javax.annotation.Nullable;
+import javax.annotation.concurrent.GuardedBy;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 import java.util.Map;
+import java.util.regex.Pattern;
 
 final class ViewDescriptor extends AbstractChainedDescriptor<View> implements HighlightableDescriptor {
-  private static final String ID_ATTRIBUTE_NAME = "id";
+  private static final String ID_NAME = "id";
+  private static final String NONE_VALUE = "(none)";
+  private static final String NONE_MAPPING = "<no mapping>";
 
   private final MethodInvoker mMethodInvoker;
+
+  /**
+   * NOTE: Only access this via {@link #getWordBoundaryPattern}.
+   */
+  @Nullable
+  private Pattern mWordBoundaryPattern;
+
+  /**
+   * NOTE: Only access this via {@link #getViewProperties}.
+   */
+  @Nullable
+  @GuardedBy("this")
+  private volatile List<ViewCSSProperty> mViewProperties;
+
+  private Pattern getWordBoundaryPattern() {
+    if (mWordBoundaryPattern == null) {
+      mWordBoundaryPattern = Pattern.compile("(?<=\\p{Lower})(?=\\p{Upper})");
+    }
+
+    return mWordBoundaryPattern;
+  }
+
+  private List<ViewCSSProperty> getViewProperties() {
+    if (mViewProperties == null) {
+      synchronized (this) {
+        if (mViewProperties == null) {
+          List<ViewCSSProperty> props = new ArrayList<>();
+
+          for (final Method method : View.class.getDeclaredMethods()) {
+            ViewDebug.ExportedProperty annotation =
+                method.getAnnotation(
+                    ViewDebug.ExportedProperty.class);
+
+            if (annotation != null) {
+              props.add(new MethodBackedCSSProperty(
+                  method,
+                  convertViewPropertyNameToCSSName(method.getName()),
+                  annotation));
+            }
+          }
+
+          for (final Field field : View.class.getDeclaredFields()) {
+            ViewDebug.ExportedProperty annotation =
+                field.getAnnotation(
+                    ViewDebug.ExportedProperty.class);
+
+            if (annotation != null) {
+              props.add(new FieldBackedCSSProperty(
+                  field,
+                  convertViewPropertyNameToCSSName(field.getName()),
+                  annotation));
+            }
+          }
+
+          mViewProperties = Collections.unmodifiableList(props);
+        }
+      }
+    }
+
+    return mViewProperties;
+  }
 
   public ViewDescriptor() {
     this(new MethodInvoker());
@@ -44,7 +120,7 @@ final class ViewDescriptor extends AbstractChainedDescriptor<View> implements Hi
   protected void onGetAttributes(View element, AttributeAccumulator attributes) {
     String id = getIdAttribute(element);
     if (id != null) {
-      attributes.store(ID_ATTRIBUTE_NAME, id);
+      attributes.store(ID_NAME, id);
     }
   }
 
@@ -69,7 +145,190 @@ final class ViewDescriptor extends AbstractChainedDescriptor<View> implements Hi
 
   @Override
   public View getViewForHighlighting(Object element) {
-    return (View)element;
+    return (View) element;
+  }
+
+  @Override
+  protected void onGetStyles(View element, StyleAccumulator styles) {
+
+    List<ViewCSSProperty> properties = getViewProperties();
+    for (int i = 0, size = properties.size(); i < size; i++) {
+      ViewCSSProperty property = properties.get(i);
+      try {
+        getStyleFromValue(
+            element,
+            property.getCSSName(),
+            property.getValue(element),
+            property.getAnnotation(),
+            styles);
+      } catch (IllegalAccessException | InvocationTargetException e) {
+        LogUtil.e(e, "failed to get style property " + property.getCSSName() +
+            " of element= " + element.toString());
+      }
+    }
+  }
+
+  private static boolean canIntBeMappedToString(@Nullable ViewDebug.ExportedProperty annotation) {
+    return annotation != null
+        && annotation.mapping() != null
+        && annotation.mapping().length > 0;
+  }
+
+  private static String mapIntToStringUsingAnnotation(
+      int value,
+      @Nullable ViewDebug.ExportedProperty annotation) {
+    if (!canIntBeMappedToString(annotation)) {
+      throw new IllegalStateException("Cannot map using this annotation");
+    }
+
+    for (ViewDebug.IntToString map : annotation.mapping()) {
+      if (map.from() == value) {
+        return map.to();
+      }
+    }
+
+    // no mapping was found even though one was expected ):
+    return NONE_MAPPING;
+  }
+
+  private static boolean canFlagsBeMappedToString(@Nullable ViewDebug.ExportedProperty annotation) {
+    return annotation != null
+        && annotation.flagMapping() != null
+        && annotation.flagMapping().length > 0;
+  }
+
+  private static String mapFlagsToStringUsingAnnotation(
+      int value,
+      @Nullable ViewDebug.ExportedProperty annotation) {
+    if (!canFlagsBeMappedToString(annotation)) {
+      throw new IllegalStateException("Cannot map using this annotation");
+    }
+
+    StringBuilder stringBuilder = null;
+    boolean atLeastOneFlag = false;
+
+    for (ViewDebug.FlagToString flagToString : annotation.flagMapping()) {
+      if (flagToString.outputIf() == ((value & flagToString.mask()) == flagToString.equals())) {
+        if (stringBuilder == null) {
+          stringBuilder = new StringBuilder();
+        }
+
+        if (atLeastOneFlag) {
+          stringBuilder.append(" | ");
+        }
+
+        stringBuilder.append(flagToString.name());
+        atLeastOneFlag = true;
+      }
+    }
+
+    if (atLeastOneFlag) {
+      return stringBuilder.toString();
+    } else {
+      return NONE_MAPPING;
+    }
+  }
+
+  private static boolean isDefaultValue(
+      Float value) {
+    return value == 0.0f;
+  }
+
+  private static boolean isDefaultValue(
+      Integer value,
+      @Nullable ViewDebug.ExportedProperty annotation) {
+    // Mappable ints should always be shown, because enums don't necessarily have
+    // logical "default" values. Thus we mark all of them as not default, so that they
+    // show up in the inspector.
+    if (canFlagsBeMappedToString(annotation) || canIntBeMappedToString(annotation)) {
+      return false;
+    }
+
+    return value == 0;
+  }
+
+  private String convertViewPropertyNameToCSSName(String getterName) {
+    // Split string by uppercase characters. Thankfully since
+    // this is the android source we don't have to worry about
+    // internationalization funk.
+
+    String[] words = getWordBoundaryPattern().split(getterName);
+
+    StringBuilder result = new StringBuilder();
+
+    for (int i = 0; i < words.length; i++) {
+      if (words[i].equals("get") || words[i].equals("m")) {
+        continue;
+      }
+
+      result.append(words[i].toLowerCase());
+
+      if (i < words.length - 1) {
+        result.append('-');
+      }
+    }
+
+    return result.toString();
+  }
+
+  private void getStyleFromValue(
+      View element,
+      String name,
+      Object value,
+      @Nullable ViewDebug.ExportedProperty annotation,
+      StyleAccumulator styles) {
+
+    if (name.equals(ID_NAME)) {
+      getIdStyle(element, styles);
+    } else if (value instanceof Integer) {
+      getStyleFromInteger(name, (Integer) value, annotation, styles);
+    } else if (value instanceof Float) {
+      getStyleFromFloat(name, (Float) value, annotation, styles);
+    }
+  }
+
+  private void getIdStyle(
+      View element,
+      StyleAccumulator styles) {
+
+    @Nullable String id = getIdAttribute(element);
+
+    if (id == null) {
+      styles.store(ID_NAME, NONE_VALUE, false);
+    } else {
+      styles.store(ID_NAME, id, false);
+    }
+  }
+
+  private void getStyleFromInteger(
+      String name,
+      Integer value,
+      @Nullable ViewDebug.ExportedProperty annotation,
+      StyleAccumulator styles) {
+
+    String intValueStr = IntegerFormatter.getInstance().format(value, annotation);
+
+    if (canIntBeMappedToString(annotation)) {
+      styles.store(
+          name,
+          intValueStr + " (" + mapIntToStringUsingAnnotation(value, annotation) + ")",
+          false);
+    } else if (canFlagsBeMappedToString(annotation)) {
+      styles.store(
+          name,
+          intValueStr + " (" + mapFlagsToStringUsingAnnotation(value, annotation) + ")",
+          false);
+    } else {
+      styles.store(name, intValueStr, isDefaultValue(value, annotation));
+    }
+  }
+
+  private void getStyleFromFloat(
+      String name,
+      Float value,
+      @Nullable ViewDebug.ExportedProperty annotation,
+      StyleAccumulator styles) {
+    styles.store(name, String.valueOf(value), isDefaultValue(value));
   }
 
   private static String capitalize(String str) {
@@ -79,5 +338,62 @@ final class ViewDescriptor extends AbstractChainedDescriptor<View> implements Hi
     StringBuilder buffer = new StringBuilder(str);
     buffer.setCharAt(0, Character.toTitleCase(buffer.charAt(0)));
     return buffer.toString();
+  }
+
+  private final class FieldBackedCSSProperty extends ViewCSSProperty {
+    private final Field mField;
+
+    public FieldBackedCSSProperty(
+        Field field,
+        String cssName,
+        @Nullable ViewDebug.ExportedProperty annotation) {
+      super(cssName, annotation);
+      mField = field;
+      mField.setAccessible(true);
+    }
+
+    @Override
+    public Object getValue(View view) throws InvocationTargetException, IllegalAccessException {
+      return mField.get(view);
+    }
+  }
+
+  private final class MethodBackedCSSProperty extends ViewCSSProperty {
+    private final Method mMethod;
+
+    public MethodBackedCSSProperty(
+        Method method,
+        String cssName,
+        @Nullable ViewDebug.ExportedProperty annotation) {
+      super(cssName, annotation);
+      mMethod = method;
+      mMethod.setAccessible(true);
+    }
+
+    @Override
+    public Object getValue(View view) throws InvocationTargetException, IllegalAccessException {
+      return mMethod.invoke(view);
+    }
+  }
+
+  private abstract class ViewCSSProperty {
+    private final String mCSSName;
+    private final ViewDebug.ExportedProperty mAnnotation;
+
+    public ViewCSSProperty(String cssName, @Nullable ViewDebug.ExportedProperty annotation) {
+      mCSSName = cssName;
+      mAnnotation = annotation;
+    }
+
+    public final String getCSSName() {
+      return mCSSName;
+    }
+
+    public abstract Object getValue(View view)
+        throws InvocationTargetException, IllegalAccessException;
+
+    public final @Nullable ViewDebug.ExportedProperty getAnnotation() {
+      return mAnnotation;
+    }
   }
 }
--- a/stetho/src/main/java/com/facebook/stetho/inspector/helper/IntegerFormatter.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/helper/IntegerFormatter.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.helper;
+
+import android.annotation.TargetApi;
+import android.os.Build;
+import android.support.annotation.Nullable;
+import android.view.ViewDebug;
+
+public class IntegerFormatter {
+  private static IntegerFormatter cachedFormatter;
+
+  public static IntegerFormatter getInstance() {
+    if (cachedFormatter == null) {
+      synchronized (IntegerFormatter.class) {
+        if (cachedFormatter == null) {
+          if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+            cachedFormatter = new IntegerFormatterWithHex();
+          } else {
+            cachedFormatter = new IntegerFormatter();
+          }
+        }
+      }
+    }
+
+    return cachedFormatter;
+  }
+
+  private IntegerFormatter() {
+  }
+
+  public String format(Integer integer, @Nullable ViewDebug.ExportedProperty annotation) {
+    return String.valueOf(integer);
+  }
+
+  private static class IntegerFormatterWithHex extends IntegerFormatter {
+    @Override
+    @TargetApi(Build.VERSION_CODES.LOLLIPOP)
+    public String format(Integer integer, @Nullable ViewDebug.ExportedProperty annotation) {
+      if (annotation != null && annotation.formatToHexString()) {
+        return "0x" + Integer.toHexString(integer);
+      }
+
+      return super.format(integer, annotation);
+    }
+  }
+}
--- a/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
@@ -9,19 +9,36 @@
 
 package com.facebook.stetho.inspector.protocol.module;
 
-import java.util.Collections;
-import java.util.List;
-
+import com.facebook.stetho.common.ListUtil;
+import com.facebook.stetho.common.LogUtil;
+import com.facebook.stetho.common.Util;
+import com.facebook.stetho.inspector.elements.Document;
+import com.facebook.stetho.inspector.elements.Origin;
+import com.facebook.stetho.inspector.elements.StyleAccumulator;
+import com.facebook.stetho.inspector.helper.ChromePeerManager;
+import com.facebook.stetho.inspector.helper.PeersRegisteredListener;
 import com.facebook.stetho.inspector.jsonrpc.JsonRpcPeer;
 import com.facebook.stetho.inspector.jsonrpc.JsonRpcResult;
 import com.facebook.stetho.inspector.protocol.ChromeDevtoolsDomain;
 import com.facebook.stetho.inspector.protocol.ChromeDevtoolsMethod;
+import com.facebook.stetho.json.ObjectMapper;
 import com.facebook.stetho.json.annotation.JsonProperty;
-
 import org.json.JSONObject;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
 public class CSS implements ChromeDevtoolsDomain {
-  public CSS() {
+  private final ChromePeerManager mPeerManager;
+  private final Document mDocument;
+  private final ObjectMapper mObjectMapper;
+
+  public CSS(Document document) {
+    mDocument = Util.throwIfNull(document);
+    mObjectMapper = new ObjectMapper();
+    mPeerManager = new ChromePeerManager();
+    mPeerManager.setListener(new PeerManagerListener());
   }
 
   @ChromeDevtoolsMethod
@@ -33,18 +50,286 @@ public class CSS implements ChromeDevtoolsDomain {
   }
 
   @ChromeDevtoolsMethod
-  public JsonRpcResult getSupportedCSSProperties(JsonRpcPeer peer, JSONObject params) {
-    GetSupportedCSSPropertiesResponse response = new GetSupportedCSSPropertiesResponse();
-    response.cssProperties = Collections.emptyList();
-    return response;
+  public JsonRpcResult getComputedStyleForNode(JsonRpcPeer peer, JSONObject params) {
+    final GetComputedStyleForNodeRequest request = mObjectMapper.convertValue(
+        params,
+        GetComputedStyleForNodeRequest.class);
+
+    final GetComputedStyleForNodeResult result = new GetComputedStyleForNodeResult();
+    result.computedStyle = new ArrayList<>();
+
+    mDocument.postAndWait(new Runnable() {
+      @Override
+      public void run() {
+        Object element = mDocument.getElementForNodeId(request.nodeId);
+
+        if (element == null) {
+          LogUtil.e("Tried to get the style of an element that does not exist, using nodeid=" +
+              request.nodeId);
+
+          return;
+        }
+
+        mDocument.getElementStyles(
+            element,
+            new StyleAccumulator() {
+              @Override
+              public void store(String name, String value, boolean isDefault) {
+                if (!isDefault) {
+                  CSSComputedStyleProperty property = new CSSComputedStyleProperty();
+                  property.name = name;
+                  property.value = value;
+
+                  result.computedStyle.add(property);
+                }
+              }
+            });
+      }
+    });
+
+    return result;
   }
 
-  private static class GetSupportedCSSPropertiesResponse implements JsonRpcResult {
+  @ChromeDevtoolsMethod
+  public JsonRpcResult getMatchedStylesForNode(JsonRpcPeer peer, JSONObject params) {
+    final GetMatchedStylesForNodeRequest request = mObjectMapper.convertValue(
+        params,
+        GetMatchedStylesForNodeRequest.class);
+
+    final GetMatchedStylesForNodeResult result = new GetMatchedStylesForNodeResult();
+
+    final RuleMatch match = new RuleMatch();
+
+    result.matchedCSSRules = ListUtil.newImmutableList(match);
+
+    match.matchingSelectors = ListUtil.newImmutableList(0);
+
+    Selector selector = new Selector();
+    selector.value = "<this_element>";
+
+    CSSRule rule = new CSSRule();
+
+    rule.origin = Origin.REGULAR;
+    rule.selectorList = new SelectorList();
+
+    rule.selectorList.selectors = ListUtil.newImmutableList(selector);
+
+    rule.style = new CSSStyle();
+    rule.style.cssProperties = new ArrayList<>();
+
+    match.rule = rule;
+
+    rule.style.shorthandEntries = Collections.emptyList();
+
+    mDocument.postAndWait(new Runnable() {
+      @Override
+      public void run() {
+        Object elementForNodeId = mDocument.getElementForNodeId(request.nodeId);
+
+        if (elementForNodeId == null) {
+          LogUtil.w("Failed to get style of an element that does not exist, nodeid=" +
+              request.nodeId);
+          return;
+        }
+
+        mDocument.getElementStyles(
+            elementForNodeId,
+            new StyleAccumulator() {
+              @Override
+              public void store(String name, String value, boolean isDefault) {
+                if (!isDefault) {
+                  CSSProperty property = new CSSProperty();
+                  property.name = name;
+                  property.value = value;
+
+                  match.rule.style.cssProperties.add(property);
+                }
+              }
+            });
+      }
+    });
+
+    result.inherited = Collections.emptyList();
+    result.pseudoElements = Collections.emptyList();
+
+    return result;
+  }
+
+  private final class PeerManagerListener extends PeersRegisteredListener {
+    @Override
+    protected synchronized void onFirstPeerRegistered() {
+      mDocument.addRef();
+    }
+
+    @Override
+    protected synchronized void onLastPeerUnregistered() {
+      mDocument.release();
+    }
+  }
+
+  private static class CSSComputedStyleProperty {
     @JsonProperty(required = true)
-    public List<CSSPropertyInfo> cssProperties;
+    public String name;
+
+    @JsonProperty(required = true)
+    public String value;
   }
 
-  private static class CSSPropertyInfo {
-    // Incomplete
+  private static class RuleMatch {
+    @JsonProperty
+    public CSSRule rule;
+
+    @JsonProperty
+    public List<Integer> matchingSelectors;
+  }
+
+  private static class SelectorList {
+    @JsonProperty
+    public List<Selector> selectors;
+
+    @JsonProperty
+    public String text;
+  }
+
+  private static class SourceRange {
+    @JsonProperty(required = true)
+    public int startLine;
+
+    @JsonProperty(required = true)
+    public int startColumn;
+
+    @JsonProperty(required = true)
+    public int endLine;
+
+    @JsonProperty(required = true)
+    public int endColumn;
+  }
+
+  private static class Selector {
+    @JsonProperty(required = true)
+    public String value;
+
+    @JsonProperty
+    public SourceRange range;
+  }
+
+  private static class CSSRule {
+    @JsonProperty
+    public String styleSheetId;
+
+    @JsonProperty(required = true)
+    public SelectorList selectorList;
+
+    @JsonProperty
+    public Origin origin;
+
+    @JsonProperty
+    public CSSStyle style;
+  }
+
+  private static class CSSStyle {
+    @JsonProperty
+    public String styleSheetId;
+
+    @JsonProperty(required = true)
+    public List<CSSProperty> cssProperties;
+
+    @JsonProperty
+    public List<ShorthandEntry> shorthandEntries;
+
+    @JsonProperty
+    public String cssText;
+
+    @JsonProperty
+    public SourceRange range;
+  }
+
+  private static class ShorthandEntry {
+    @JsonProperty(required = true)
+    public String name;
+
+    @JsonProperty(required = true)
+    public String value;
+
+    @JsonProperty
+    public Boolean imporant;
+  }
+
+  private static class CSSProperty {
+    @JsonProperty(required = true)
+    public String name;
+
+    @JsonProperty(required = true)
+    public String value;
+
+    @JsonProperty
+    public Boolean important;
+
+    @JsonProperty
+    public Boolean implicit;
+
+    @JsonProperty
+    public String text;
+
+    @JsonProperty
+    public Boolean parsedOk;
+
+    @JsonProperty
+    public Boolean disabled;
+
+    @JsonProperty
+    public SourceRange range;
+  }
+
+  private static class PseudoIdMatches {
+    @JsonProperty(required = true)
+    public int pseudoId;
+
+    @JsonProperty(required = true)
+    public List<RuleMatch> matches;
+
+    public PseudoIdMatches() {
+      this.matches = new ArrayList<>();
+    }
+  }
+
+  private static class GetComputedStyleForNodeRequest {
+    @JsonProperty(required = true)
+    public int nodeId;
+  }
+
+  private static class InheritedStyleEntry {
+    @JsonProperty(required = true)
+    public CSSStyle inlineStyle;
+
+    @JsonProperty(required = true)
+    public List<RuleMatch> matchedCSSRules;
+  }
+
+  private static class GetComputedStyleForNodeResult implements JsonRpcResult {
+    @JsonProperty(required = true)
+    public List<CSSComputedStyleProperty> computedStyle;
+  }
+
+  private static class GetMatchedStylesForNodeRequest implements JsonRpcResult {
+    @JsonProperty(required = true)
+    public int nodeId;
+
+    @JsonProperty
+    public Boolean excludePseudo;
+
+    @JsonProperty
+    public Boolean excludeInherited;
+  }
+
+  private static class GetMatchedStylesForNodeResult implements JsonRpcResult {
+    @JsonProperty
+    public List<RuleMatch> matchedCSSRules;
+
+    @JsonProperty
+    public List<PseudoIdMatches> pseudoElements;
+
+    @JsonProperty
+    public List<InheritedStyleEntry> inherited;
   }
 }
--- a/stetho/src/main/java/com/facebook/stetho/Stetho.java
+++ b/stetho/src/main/java/com/facebook/stetho/Stetho.java
@@ -259,7 +259,6 @@ public class Stetho {
 
     public Iterable<ChromeDevtoolsDomain> finish() {
       provideIfDesired(new Console());
-      provideIfDesired(new CSS());
       provideIfDesired(new Debugger());
       if (Build.VERSION.SDK_INT >= AndroidDOMConstants.MIN_API_LEVEL) {
         Document document = new Document(
@@ -267,6 +266,7 @@ public class Stetho {
                 (Application) mContext.getApplicationContext()));
 
         provideIfDesired(new DOM(document));
+        provideIfDesired(new CSS(document));
       }
       provideIfDesired(new DOMStorage(mContext));
       provideIfDesired(new HeapProfiler());
--- a/stetho/src/main/java/com/facebook/stetho/common/ListUtil.java
+++ b/stetho/src/main/java/com/facebook/stetho/common/ListUtil.java
@@ -65,6 +65,10 @@ public final class ListUtil {
     }
   }
 
+  public static <T> List<T> newImmutableList(T item) {
+    return new OneItemImmutableList<>(item);
+  }
+
   private static interface ImmutableList<E> extends List<E>, RandomAccess {
   }
 
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/ObjectDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/ObjectDescriptor.java
@@ -51,4 +51,8 @@ public final class ObjectDescriptor extends Descriptor {
   @Override
   public void setAttributesAsText(Object element, String text) {
   }
+
+  @Override
+  public void getStyles(Object element, StyleAccumulator accumulator) {
+  }
 }
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/StyleAccumulator.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/StyleAccumulator.java
@@ -1,0 +1,14 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.elements;
+
+public interface StyleAccumulator {
+  void store(String name, String value, boolean isDefault);
+}


stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
```diff
@@ -140,6 +140,14 @@ public void setAttributesAsText(Object element, String text) {
     mDocumentProvider.setAttributesAsText(element, text);
   }
 
+  public void getElementStyle(int id, Accumulator<StyleInfo> styleInfoAccumulator) {
+    Object element = this.getElementForNodeId(id);
```
rickbrew: style: remove `this.` prefix

stetho/src/main/java/com/facebook/stetho/inspector/elements/AbstractChainedDescriptor.java
```diff
@@ -144,4 +144,14 @@ public final void setAttributesAsText(Object element, String text) {
...
   }
+
+  @Override
+  @SuppressWarnings("unchecked")
+  public final void getStyleProperties(Object element, Accumulator<StyleInfo> accumulator) {
+    onGetStyles((E) element, accumulator);
+  }
+
+  protected void onGetStyles(E element, Accumulator<StyleInfo> accumulator) {
+    mSuper.getStyleProperties(element, accumulator);
```
rickbrew: shouldn't this call to `mSuper.getStyleProperties(element, accumulator);` be done in `getStyleProperties` ? That way we get automatic aggregation of the super's style information. see also: `getChildren` and `onGetChildren` as well as `getAttributes` and `onGetAttributes`

stetho/src/main/java/com/facebook/stetho/inspector/elements/NodeDescriptor.java
```diff
@@ -33,4 +33,6 @@
   void getAttributes(Object element, AttributeAccumulator attributes);
 
   void setAttributesAsText(Object element, String text);
+
+  void getStyleProperties(Object element, Accumulator<StyleInfo> accumulator);
```
rickbrew: we have a naming problem here ... are we getting the element's style *properties* or *info* ? we should choose one word and stick with it everywhere (in other words, consistency please!)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -10,19 +10,71 @@
...
 import java.util.Map;
+import java.util.Set;
 
 final class ViewDescriptor extends AbstractChainedDescriptor<View> implements HighlightableDescriptor {
   private static final String ID_ATTRIBUTE_NAME = "id";
+  private static final Set<Method> viewGetters;
 
   private final MethodInvoker mMethodInvoker;
 
+  static {
```
rickbrew: I see 2 performance problems with this approach. Since this code is executing in the static initializer for ViewDescriptor, it will always run when the Elements tab is booted up even if the CSS stuff isn't being used. That violates pay-for-play, which is very important for a project like Stetho. The second problem is due to this being stored in static fields, which means it is not eligible for garbage collection once all Stetho clients have disconnected.

Because `ViewDescriptor` is essentially a singleton, it is perfectly valid for us to store this type of data in instance fields (btw, non-final/mutable static fields should be prefixed with 's'. final, or rather "effectively const" fields, should be ALL_UPPERCASE). And, this data should not be initialized until it's needed ... which looks like it'd be the first call to `onGetStyles()`. A good idea would be to hide this behind a method that handles the deferred initialization, e.g.

```
/**
  *  NOTE: Only access this via {@link #getViewGetters}.
  */
@GuardedBy("this")
private volatile Set<Method> mViewGetters;

private Set<Method> getViewGetters() {
  if (mViewGetters == null) {
    synchronized (this) {
      ... initialize mViewGetters ...
    }
  }
  return mViewGetters;
}

@Override
protected void onGetStyles(...) {
  for (Method method : getViewGetters()) { 
    ...
  }
}
```

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -72,6 +124,125 @@ public View getViewForHighlighting(Object element) {
...
+              getIdAttribute(element));
+        }
+      }
+    }
+  }
+
+  private static boolean canIntBeMappedToString(
+      ViewDebug.ExportedProperty annotation) {
+    return annotation != null
+           && annotation.mapping().length > 0;
```
rickbrew: style: if we're putting && on the next line, convention is to pull it under the return like so:

```
return annotation != null
    && annotation.mapping.length() > 0;
```

'return' is 6 characters, and so 4 space indent + 2 space operator results in things being lined up

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -72,6 +124,125 @@ public View getViewForHighlighting(Object element) {
...
+    }
+
+    return null;
+  }
+
+  private static String convertGetterNameToCssName(String getterName) {
+    // Split string by uppercase characters. Thankfully since
+    // this is the android source we don't have to worry about
+    // internationalization funk.
+    String[] words = getterName.split("(?=\\p{Upper})");
```
rickbrew: because `String.split` uses a regex, this means we're compiling a regex every single time this method is called! I'm not sure what a better approach is just yet, but we will probably need to optimize this. Even if we just create our own method that doesn't use regex and uses brute force loops and stuff. (and then the output could also be funneled into an `Accumulator` and thus allow us to save lots of temporary allocations :) )

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -9,19 +9,33 @@
...
 import org.json.JSONObject;
 
+import java.util.ArrayList;
+import java.util.List;
+
 public class CSS implements ChromeDevtoolsDomain {
-  public CSS() {
+  private final ChromePeerManager mPeerManager;
+  private final Document mDocument;
+  private ObjectMapper mObjectMapper;
```
rickbrew: can this be `final` ?

stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
```diff
@@ -140,6 +140,14 @@ public void setAttributesAsText(Object element, String text) {
     mDocumentProvider.setAttributesAsText(element, text);
   }
 
+  public void getElementStyle(int id, Accumulator<StyleInfo> styleInfoAccumulator) {
```
rickbrew: I think we need to figure out what type of API we are really wanting to provide with `Document`. This method is essentially a utility method that can serve to confuse clients ("but I can do that myself via NodeDescriptor ... why would I use Document.getElementStyle?").

I also find it weird that this method is providing a access via an integer NodeID, but many other methods are using an Object element. Right now this seems to map to the direct needs of `CSS` which receives its request as an integer NodeID, but I think we can do better and provide a less confusing API overall.

Nothing directly actionable from this comment -- fodder for chin-scratching and later discussion.

stetho/src/main/java/com/facebook/stetho/inspector/elements/StyleInfo.java
```diff
@@ -0,0 +1,34 @@
...
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.elements;
+
+public class StyleInfo {
```
rickbrew: `final`

stetho/src/main/java/com/facebook/stetho/inspector/elements/StyleInfo.java
```diff
@@ -0,0 +1,34 @@
...
+  private String name;
```
rickbrew: since we're going with an immutable contract, these fields should be `final`

rickbrew: since these fields are private we also need to add the `m` prefix

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -10,19 +10,71 @@
...
+import java.util.Set;
 
 final class ViewDescriptor extends AbstractChainedDescriptor<View> implements HighlightableDescriptor {
   private static final String ID_ATTRIBUTE_NAME = "id";
+  private static final Set<Method> viewGetters;
 
   private final MethodInvoker mMethodInvoker;
 
+  static {
+    Set<String> blacklistedGetters = new HashSet<>(
```
rickbrew: `disallowed` is probably a better term than `blacklisted`

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -10,19 +10,71 @@
...
+        "getNextFocusForwardId",
+        "getNextFocusLeftId",
+        "getNextFocusRightId",
+        "getNextFocusUpId",
+        "getOverScrollMode",
+        "getSystemUiVisibility"
+      ));
+
+    Method[] viewMethods = View.class.getDeclaredMethods();
+    List<Method> getters = new ArrayList<>();
```
rickbrew: why not just have this be a `HashSet` ? otherwise we're populating an `ArrayList` and then copying it into a `HashSet` and wrapping it into an unmodifiable `Set`. if this is going to be a `Set` at the end, and since we don't need `List` semantics (right?), we may as well just skip the middle man

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -10,19 +10,71 @@
...
 
 final class ViewDescriptor extends AbstractChainedDescriptor<View> implements HighlightableDescriptor {
   private static final String ID_ATTRIBUTE_NAME = "id";
+  private static final Set<Method> viewGetters;
 
   private final MethodInvoker mMethodInvoker;
 
+  static {
+    Set<String> blacklistedGetters = new HashSet<>(
+      Arrays.asList(
```
rickbrew: We should consider an alternate implementation for the disallowed list of method names. Instead of creating an array, then wrapping into a `List` so that we can copy it into a `HashSet` ... why don't we try this instead: Process the list of methods (`View.class.getDeclaredMethods`) without regard to whether they're allowed or disallowed. Then, before finalizing the `Set` of methods, just call `remove` for each of the disallowed methods, e.g.

```
Method[] viewMethods = ...
...
for (Method method : viewMethods) {
  if (isMethodsGetter(method) 
      && method.getReturnType().isPrimitive()) {
    getters.add(method);
  }
}

// These getters aren't allowed or something
getters.remove("getAccessibilityCursorPosition");
getters.remove("getAccessibilityWindowId");
...
getters.remove("getOverScrollMode");
getters.remove("getSystemUiVisibility");
```

This will help to eliminate a bunch of temporary allocations and CPU cycles. The one disadvantage is that the final `Set` of getters may end up with a slightly larger internal array allocation. We can still optimize that if necessary.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -72,6 +124,125 @@ public View getViewForHighlighting(Object element) {
     return (View)element;
   }
 
+  @Override
+  protected void onGetStyles(View element, Accumulator<StyleInfo> styles) {
+    for (Method method : viewGetters) {
+      boolean caught = false;
+      try {
+        ViewDebug.ExportedProperty annotation = getExportedAnnotation(method);
```
rickbrew: let's decorate this variable with `@Nullable` since it's pretty clear that it can be `null`

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -72,6 +124,125 @@ public View getViewForHighlighting(Object element) {
...
+        if (caught) {
+          LogUtil.w("failed to get the style of element id=" +
+              getIdAttribute(element));
+        }
+      }
+    }
+  }
+
+  private static boolean canIntBeMappedToString(
+      ViewDebug.ExportedProperty annotation) {
```
rickbrew: since `annotation` is allowed to be `null` let's add `@Nullable` to it

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -72,6 +124,125 @@ public View getViewForHighlighting(Object element) {
...
+      try {
+        ViewDebug.ExportedProperty annotation = getExportedAnnotation(method);
+        Object returnedValue = method.invoke(element);
+        String cssName = convertGetterNameToCssName(method.getName());
+        String cssValue = String.valueOf(returnedValue);
+        boolean isValueDefault = false;
+
+        // special case id to translate it into a human-readable string
+        if (method.getName().equals("getId")) {
+          cssValue = getIdAttribute(element);
```
rickbrew: We should strive to adhere to Single Assignment Form, which can be thought of as being part of Functional Programming Style. This method is writing to the value of `cssValue` multiple times, and this can make things very confusing when it comes time to read or debug it later.

Basically, stick a `final` on the front of each variable and find ways to transform your code to work within that constraint. In this case you can probably defer calling `String.valueOf(returnedValue)` (which is currently assigned to cssValue when it's declared) until much later, once you've ruled out the other alternatives.

This type of transformation isn't always straightforward if you're not accustomed to it, so I'm happy to help out with it. I promise it's worth it!

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -72,6 +124,125 @@ public View getViewForHighlighting(Object element) {
...
+              isValueDefault |= isNumberDefault(intValue, annotation);
+            }
+          } else if (method.getReturnType() == float.class) {
+            isValueDefault |= isNumberDefault((Number)returnedValue, annotation);
+          }
+        }
+
+        styles.store(new StyleInfo(cssName, cssValue, isValueDefault));
+      } catch (IllegalAccessException e) {
+        caught = true;
```
rickbrew: Hmm. This whole approach of trying to use all methods prefixed with `get` and then ignoring the ones that throw exceptions seems to be overreaching and dangerous. We should consider having a stock list of `get` methods that we support instead of vacuuming up *any* that begin with the word `get`.

Also, I'm worried that we'll be spamming logcat with all sorts of warnings that the developer can't do anything about and which are actually our fault.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -72,6 +124,125 @@ public View getViewForHighlighting(Object element) {
...
+    }
+
+    for (ViewDebug.IntToString map : annotation.mapping()) {
+      if (map.from() == value) {
+        return map.to();
+      }
+    }
+
+    // no mapping was found even though one was expected ):
+    return " <no mapping> ";
```
rickbrew: why is this surrounded by spaces?

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -72,6 +124,125 @@ public View getViewForHighlighting(Object element) {
...
+  }
+
+  private static boolean isNumberDefault(Number value, ViewDebug.ExportedProperty annotation) {
+    if (annotation != null) {
+      if (annotation.mapping().length > 0) {
+        return false;
+      }
+    }
+
+    return value.floatValue() == 0.0;
```
rickbrew: This seems fragile. Is `value` always going to be a float? Isn't `0.0` going to map to a `double`? (not 100% sure -- I'm used to C# where that is true). It's not a good idea to rely on floating point comparisons to an exact value. If this is an integer then we should compare to integer zero instead of converting to float and then comparing.

Also, this method might be better called `isDefaultValue`

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -72,6 +124,125 @@ public View getViewForHighlighting(Object element) {
...
+  }
+
+  private static boolean isMethodGetter(Method method) {
```
rickbrew: prefer to call this `isGetterMethod` ... otherwise it sounds like we're testing to see if this method can get other methods

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -72,6 +124,125 @@ public View getViewForHighlighting(Object element) {
...
+    return method.getName().toLowerCase().startsWith("get")
+           && method.getParameterTypes().length == 0;
```
rickbrew: style: same comment as before regarding indenting the `&&` underneath `return`

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -33,18 +47,278 @@ public void disable(JsonRpcPeer peer, JSONObject params) {
...
+  @ChromeDevtoolsMethod
+  public JsonRpcResult getMatchedStylesForNode(JsonRpcPeer peer, JSONObject params) {
+    final GetMatchedStylesForNodeRequest request = mObjectMapper.convertValue(
+      params,
+      GetMatchedStylesForNodeRequest.class);
+
+    final ArrayListAccumulator<StyleInfo> styles = new ArrayListAccumulator<StyleInfo>() {
+      @Override
+      public void store(StyleInfo object) {
+        if (!object.isDefault()) {
```
rickbrew: I'm betting we're going to be filtering out a lot of the `StyleInfo` objects? Many will be the default value? This leads me to think that a custom `Accumulator`, ala `AttributeAccumulator`, will be a better idea here. Then we wouldn't have to allocate unless we're going to transmit. Something like this might work:

```
interface StyleInfoAccumulator {
    void store(String name, String value, boolean isDefault);
}
```

And then right here our custom accumulator implementation would only have to do `new StyleInfo` for when `isDefault` is false.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -33,18 +47,278 @@ public void disable(JsonRpcPeer peer, JSONObject params) {
...
+    public List<RuleMatch> matchedCSSRules;
+
+    @JsonProperty
+    public List<PseudoIdMatches> pseudoElements;
+
+    @JsonProperty
+    public List<InheritedStyleEntry> inherited;
+
+    public GetMatchedStylesForNodeResult(List<StyleInfo> properties) {
+      this.matchedCSSRules = new ArrayList<>();
```
rickbrew: In general we haven't been putting any logic into these request/response classes. We should continue that pattern of leaving them as brainless POJOs. In other words, we should do all this preparation outside of this class, not in its constructor.

rickbrew:
Okay there's my first round of feedback. And please add a description and screenshot.

stetho/src/main/java/com/facebook/stetho/inspector/elements/AbstractChainedDescriptor.java
```diff
@@ -144,4 +144,14 @@ public final void setAttributesAsText(Object element, String text) {
...
   }
+
+  @Override
+  @SuppressWarnings("unchecked")
+  public final void getStyleProperties(Object element, StyleAccumulator accumulator) {
+    onGetStyles((E) element, accumulator);
+    mSuper.getStyleProperties(element, accumulator);
+  }
+
+  protected void onGetStyles(E element, StyleAccumulator accumulator) {
```
rickbrew: The `on___` methods should be named according to their dual (which is the one that calls it). This should be called `onGetStyleProperties`.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -10,19 +10,132 @@
...
+  /**
+   * NOTE: Only access this via {@link #getViewGetters}.
+   */
+  @GuardedBy("this")
+  private Set<Method> VIEW_GETTERS;
+
+  private final Set<Method> getViewGetters() {
+    if (VIEW_GETTERS == null) {
+      synchronized (this) {
+        Set<String> allowedGetters = new HashSet<>(
```
rickbrew: Don't forget to check `VIEW_GETTERS` for `null` again.

https://en.wikipedia.org/wiki/Double-checked_locking

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -10,19 +10,132 @@
...
   private static final String ID_ATTRIBUTE_NAME = "id";
-
+  private static final Pattern CAPITAL_PATTERN = Pattern.compile("(?=\\p{Upper})");
   private final MethodInvoker mMethodInvoker;
 
+  /**
+   * NOTE: Only access this via {@link #getViewGetters}.
+   */
+  @GuardedBy("this")
+  private Set<Method> VIEW_GETTERS;
```
jasta: Why is this named in all caps?  It should be `mViewGetters`, no?

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -10,19 +10,132 @@
...
+
+  private final Set<Method> getViewGetters() {
```
jasta: This method looks incredibly expensive.  Is there some reason this cannot be loaded more lazily?  Also, how is the performance in practice on a mid-range phone? Does it take several seconds to load this UI?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -33,18 +46,274 @@ public void disable(JsonRpcPeer peer, JSONObject params) {
...
+
+    final RuleMatch match = new RuleMatch();
+    match.matchingSelectors = new ArrayList<>();
+
+    Selector selector = new Selector();
+    selector.value = "<this_element>";
+
+    match.matchingSelectors.add(0);
+
+    match.rule = new CSSRule();
```
jasta: This style creates unnecessarily verbose code.  Prefer:
```java
CSSRule rule = new CSSRule();
rule.foo = ...
match.rule = rule;
```

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -10,19 +10,132 @@
...
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
+import java.util.regex.Pattern;
 
 final class ViewDescriptor extends AbstractChainedDescriptor<View> implements HighlightableDescriptor {
   private static final String ID_ATTRIBUTE_NAME = "id";
-
+  private static final Pattern CAPITAL_PATTERN = Pattern.compile("(?=\\p{Upper})");
```
rickbrew: We shouldn't have this as a `static` so that it can be GC'd once all peers disconnect from Stetho.

And, we should defer initialize it so that we don't pay for it until/unless the user clicks on the Styles stuff.

```
/**
  * NOTE: Only access this via {@link #getCapitalPattern}.
  */
private @Nullable Pattern mCapitalPattern;

private Pattern getCapitalPattern() {
  if (mCapitalPattern == null) {
    mCapitalPattern = Pattern.compile(...);
  }
  return mCapitalPattern;
}
```

(We don't really need to use double-checked-null and synchronized here because if 2 threads race to initialize this then we'll just end up with a wasted call to `Pattern.compile` and the object instance it returns. It won't affect correctness. It's unlikely, and not a big deal if it does happen, and we're probably better off saving the CPU cycles by not worrying about it.)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
rickbrew: And add `@Nullable` too. The `getViewGetters` method doesn't need `@Nullable`, but having it on the field at least helps ensure that accidental access gets visually flagged by the IDE.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -72,6 +185,127 @@ public View getViewForHighlighting(Object element) {
...
+        }
+      } catch (IllegalAccessException e) {
+        caught = true;
+      } catch (InvocationTargetException e) {
+        caught = true;
+      } finally {
+        if (caught) {
+          // something went terribly wrong... none of the called
+          // methods should throw
+          LogUtil.w("(THIS SHOULD NOT HAPPEN) failed to get the style of element id=" +
```
rickbrew: we probably shouldn't be spewing out things like `THIS SHOULD NOT HAPPEN` ... if that's really the case then just use `LogUtil.e` :)

stetho/src/main/java/com/facebook/stetho/inspector/elements/AbstractChainedDescriptor.java
```diff
@@ -144,4 +144,14 @@ public final void setAttributesAsText(Object element, String text) {
   protected void onSetAttributesAsText(E element, String text) {
     mSuper.setAttributesAsText(element, text);
   }
+
+  @Override
+  @SuppressWarnings("unchecked")
+  public final void getStyleProperties(Object element, StyleAccumulator accumulator) {
```
rickbrew: Now that I think about it, using "style" and "property" together is bad terminology. We already have something in the UI called "properties." I think we should just call this `getStyles`. We shouldn't throw in words like property, info, etc.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
ichub: @rickbrew commented earlier that this field is "effectively const" so it should be named according to the convention of const - https://github.com/facebook/stetho/pull/261#discussion_r37485734

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
ichub: This is loaded as late as possible: only when you **click** on an element whose descriptor is a `ViewDescriptor`. At that point, it's absolutely necessary to enumerate all of the methods of `View`, because that's how we get the 'css' values.

Not sure about performance on low end phones - don't have access to one atm.

stetho/src/main/java/com/facebook/stetho/inspector/elements/AbstractChainedDescriptor.java
```diff
@@ -144,4 +144,14 @@ public final void setAttributesAsText(Object element, String text) {
   protected void onSetAttributesAsText(E element, String text) {
     mSuper.setAttributesAsText(element, text);
   }
+
+  @Override
+  @SuppressWarnings("unchecked")
+  public final void getStyles(Object element, StyleAccumulator accumulator) {
+    onGetStyles((E) element, accumulator);
```
rickbrew: is there any reason why we're getting local styles and then super's styles? (instead of the other way around) ... if you look at `getAttributes`, it first grabs the super's attributes and then calls `onGetAttributes` ("local" attributes)

stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
```diff
@@ -140,6 +140,12 @@ public void setAttributesAsText(Object element, String text) {
     mDocumentProvider.setAttributesAsText(element, text);
   }
 
+  public void getElementStyle(Object element, StyleAccumulator styleInfoAccumulator) {
```
rickbrew: We still need to talk about the organization of this API, re: my older comment about this being, essesntially, a utility method

For now, this method's signature is slightly off ... it should be plural (`getElementStyles`) and the parameter should have the word "info" removed

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -72,6 +200,127 @@ public View getViewForHighlighting(Object element) {
...
   }
 
+  @Override
+  protected void onGetStyles(View element, StyleAccumulator styles) {
+    for (Method method : getViewGetters()) {
+      boolean caught = false;
+      try {
+        @Nullable ViewDebug.ExportedProperty annotation = getExportedAnnotation(method);
+        final Object returnedValue = method.invoke(element);
+        final String cssName = method.getName(); //convertGetterNameToCssName(method.getName());
```
rickbrew: remove commented-out code plz

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -72,6 +200,127 @@ public View getViewForHighlighting(Object element) {
...
+
+        if (method.getName().equals("getId")) {
+          @Nullable String id = getIdAttribute(element);
+
+          if (id == null) {
+            styles.store("id", "(none)", false);
```
rickbrew: we should put these string constants into actual string constants, e.g.

```
styles.store(ID_STYLE_NAME, ID_STYLE_VALUE_NONE, false);
```

etc.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -72,6 +200,127 @@ public View getViewForHighlighting(Object element) {
     return (View)element;
   }
 
+  @Override
+  protected void onGetStyles(View element, StyleAccumulator styles) {
+    for (Method method : getViewGetters()) {
```
rickbrew: I think @jasta  and I are both of the idea that we really shouldn't be using reflection for this. Reflection just isn't fast, and in addition to that we're doing lots of processing and allocating and etc. etc.

Which methods do we really really need/want for this feature? Do we need all 70 of those? Or can we just manually hardcode a list of 20 or so? (in other words, instead of using a `Method[]` and going at it generically with reflection, we could just have explicit code for the N that we directly want to support)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -72,6 +200,127 @@ public View getViewForHighlighting(Object element) {
...
+    for (Annotation annotation : annotations) {
+      if (annotation instanceof ViewDebug.ExportedProperty) {
+        return (ViewDebug.ExportedProperty) annotation;
+      }
+    }
+
+    return null;
+  }
+
+  private String convertGetterNameToCssName(String getterName) {
```
rickbrew: unused method = delete! :)

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -33,18 +46,276 @@ public void disable(JsonRpcPeer peer, JSONObject params) {
...
+
+    match.rule = rule;
+
+    rule.style.shorthandEntries = new ArrayList<>();
+
+    mDocument.postAndWait(new Runnable() {
+      @Override
+      public void run() {
+        mDocument.getElementStyle(
+            mDocument.getElementForNodeId(request.nodeId),
```
rickbrew: if the node ID doesn't exist, this will most likely crash hard on an NPE, right? this is probably another good reason we might want to just not have `Document.getElementStyle`

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -10,19 +10,147 @@
...
+    return mCapitalPattern;
+  }
+
+  private final Set<Method> getViewGetters() {
+    if (VIEW_GETTERS == null) {
+      synchronized (this) {
+        if (VIEW_GETTERS == null) {
+          Set<String> allowedGetters = new HashSet<>(
+              Arrays.asList(
+                  "getHorizontalScrollFactor",
```
rickbrew: we should alphabetize this list too, otherwise it's really difficult to navigate

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -10,19 +10,147 @@
...
+                  "getScrollBarSize",
+                  "getMeasuredWidth",
+                  "getWindowAttachCount",
+                  "getBaseline",
+                  "getY",
+                  "getX",
+                  "getRight"
+              ));
+
+          Method[] viewMethods = View.class.getDeclaredMethods();
```
rickbrew: Something I just realized we're not handling here: We only look at `View.class.getDeclaredMethods`. We are not picking up anything from derived classes, e.g. TextView. Either this doesn't matter, or we need to find a way to do it (or just decide not to :) )

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -9,20 +9,131 @@
...
   private static final String ID_ATTRIBUTE_NAME = "id";
+  private static final String NONE_VALUE = "(none)";
+  private static final String NONE_MAPPING = "<no mapping>";
 
   private final MethodInvoker mMethodInvoker;
 
+  /**
+   * NOTE: Only access this via {@link #getCapitalPattern}.
+   */
+  private
```
rickbrew: this formatting is super strange! prefer:

```
@Nullable
private Pattern mCapitalPattern;
```

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -37,7 +148,7 @@ protected String onGetNodeName(View element) {
 
     return
         StringUtil.removePrefix(className, "android.view.",
-        StringUtil.removePrefix(className, "android.widget."));
+            StringUtil.removePrefix(className, "android.widget."));
```
rickbrew: you accidentally a whitespace change

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -33,18 +46,276 @@ public void disable(JsonRpcPeer peer, JSONObject params) {
...
+  private static class CSSRule {
+    @JsonProperty
+    public String styleSheetId;
+
+    @JsonProperty(required = true)
+    public SelectorList selectorList;
+
+    // this is an enum, and can be one
+    // of the following values:
+    // ["injected", "user-agent", "inspector", "regular"]
```
rickbrew: enums are supported directly ... check out `NodeType` which is used by `DOM.Node`. there's a pattern where you implement a `getProtocolValue` method on your enum to supply the JSON serialized form. Hopefully it works with strings.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -33,18 +46,276 @@ public void disable(JsonRpcPeer peer, JSONObject params) {
...
+
+  @ChromeDevtoolsMethod
+  public JsonRpcResult getMatchedStylesForNode(JsonRpcPeer peer, JSONObject params) {
+    final GetMatchedStylesForNodeRequest request = mObjectMapper.convertValue(
+        params,
+        GetMatchedStylesForNodeRequest.class);
+
+    final GetMatchedStylesForNodeResult result = new GetMatchedStylesForNodeResult();
+
+    result.matchedCSSRules = new ArrayList<>();
```
rickbrew: if we know we're only going to have 1 item in this list, we can specialize this. First idea would be to make sure we do something like `new ArrayList<>(1);` (specify capacity). But, we can do better because that still requires 2 allocations (one for the ArrayList, one for its internal array). 

In this case, because we're not modifying the list, we can add a method to `ListUtil` for creating a special 1-item `List` implementation, similar to what we're doing with the "immutable" lists. In fact, really what we need is a `ListUtil.newImmutableList(T)` method,

```
public static <T> newImmutableList(T item) {
  return new OneItemImmutableList<>(item);
}
```

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -69,7 +180,159 @@ private static String getIdAttribute(View element) {
...
+        }
+      } catch (IllegalAccessException e) {
+        caught = true;
+      } catch (InvocationTargetException e) {
+        caught = true;
+      } finally {
+        if (caught) {
+          // something went terribly wrong... none of the called
+          // methods should throw
+          LogUtil.e("(THIS SHOULD NOT HAPPEN) failed to get the style of element id=" +
```
rickbrew: remove "THIS SHOULD NOT HAPPEN"

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -69,7 +180,159 @@ private static String getIdAttribute(View element) {
...
+                isDefaultValue(floatValue, annotation));
+          }
+        }
+      } catch (IllegalAccessException e) {
+        caught = true;
+      } catch (InvocationTargetException e) {
+        caught = true;
+      } finally {
+        if (caught) {
+          // something went terribly wrong... none of the called
```
rickbrew: i don't think we need this comment :)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -9,20 +9,131 @@
...
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
+import java.util.regex.Pattern;
 
 final class ViewDescriptor extends AbstractChainedDescriptor<View> implements HighlightableDescriptor {
   private static final String ID_ATTRIBUTE_NAME = "id";
```
rickbrew: let's also have `ID_STYLE_NAME` instead of reusing one named constant for both

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -9,20 +9,131 @@
...
+  private
+  @Nullable
+  Pattern mCapitalPattern;
+
+  /**
+   * NOTE: Only access this via {@link #getProperties}.
+   */
+  @Nullable
+  @GuardedBy("this")
+  private volatile Set<ViewCSSProperty> VIEW_PROPERTIES;
```
rickbrew: INSTANCE FIELDS SHOULD NOT BE CAPITALIZED :)

`mViewCSSProperties`

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -69,7 +180,159 @@ private static String getIdAttribute(View element) {
...
+        }
+      }
+    }
+  }
+
+  private static boolean canIntBeMappedToString(@Nullable ViewDebug.ExportedProperty annotation) {
+    return annotation != null && annotation.mapping() != null && annotation.mapping().length > 0;
+  }
+
+  private static String mapIntToStringUsingAnnotations(
```
rickbrew: annotation"s" ... should be singular?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -33,18 +46,276 @@ public void disable(JsonRpcPeer peer, JSONObject params) {
...
+        GetComputedStyleForNodeRequest.class);
+
+    final GetComputedStyleForNodeResult result = new GetComputedStyleForNodeResult();
+    result.computedStyle = new ArrayList<>();
+
+    mDocument.postAndWait(new Runnable() {
+      @Override
+      public void run() {
+        mDocument.getElementStyles(
+            mDocument.getElementForNodeId(request.nodeId),
```
rickbrew: we should call `getElementForNodeId` and check its return value before passing that along to `getElementStyles`

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/DialogFragmentDescriptor.java
```diff
@@ -11,7 +11,6 @@
 
 import android.app.Dialog;
 import android.view.View;
-
```
rickbrew: you accidentally a whitespace?

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -9,20 +9,94 @@
...
+  }
+
+  private Set<ViewCSSProperty> getProperties() {
+    if (VIEW_PROPERTIES == null) {
+      synchronized (this) {
+        if (VIEW_PROPERTIES == null) {
+          Set<ViewCSSProperty> props = new HashSet<>();
+
+          for (final Method method : View.class.getDeclaredMethods()) {
+            for (final Annotation annotation : method.getDeclaredAnnotations()) {
```
rickbrew: have you tried using `Method.getAnnotation(ViewDebug.ExportedProperty.class)` ?

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -9,20 +9,94 @@
...
+  private static final String NONE_MAPPING = "<no mapping>";
 
   private final MethodInvoker mMethodInvoker;
 
+  /**
+   * NOTE: Only access this via {@link #getWordBoundaryPattern}.
+   */
+  private
+  @Nullable
+  Pattern mWordBoundaryPattern;
```
rickbrew: formatting on this is still weird. see my comment on a prior commit

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/DialogFragmentDescriptor.java
ichub: Weird, this change isn't showing up in android studio...

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -69,7 +143,185 @@ private static String getIdAttribute(View element) {
...
   public View getViewForHighlighting(Object element) {
-    return (View)element;
+    return (View) element;
+  }
+
+  @Override
+  protected void onGetStyles(View element, StyleAccumulator styles) {
+
+    for (ViewCSSProperty property : getProperties()) {
+      Object value = null;
```
rickbrew: recommend not initializing this to `null` here. instead, set it to null in the error paths (within the `catch` clauses). That makes is easier to determine that every code path is assigning to this and that it didn't get forgotten.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -69,7 +143,185 @@ private static String getIdAttribute(View element) {
...
+
+      try {
+        value = property.getValue(element);
+      } catch (IllegalAccessException e) {
+        caught = true;
+      } catch (InvocationTargetException e) {
+        caught = true;
+      } finally {
+        if (caught) {
+          LogUtil.e("failed to get the style of element id=" + getIdAttribute(element));
```
rickbrew: Instead of using a flag to determine if we caught an exception and then logging it, try doing the logging from one catch clause, like so

```
try {
  value = property.getValue(element);
} catch (IllegalAccessException | InvocationTargetException ex) {
  value = null; // this is where'd you set it to null, re: my previous comment
  LogUtil.e(ex, ...);
}
```

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -80,4 +332,54 @@ private static String capitalize(String str) {
...
+    }
+
+    @Override
+    Object getValue(View view) throws InvocationTargetException, IllegalAccessException {
+      return mMethod.invoke(view);
+    }
+  }
+
+  private abstract class ViewCSSProperty {
+    private String mName;
```
rickbrew: add `final` to fields

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -80,4 +332,54 @@ private static String capitalize(String str) {
     buffer.setCharAt(0, Character.toTitleCase(buffer.charAt(0)));
     return buffer.toString();
   }
+
+  private class FieldBackedCSSProperty extends ViewCSSProperty {
+    private Field mField;
```
rickbrew: `final`

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -80,4 +332,54 @@ private static String capitalize(String str) {
...
+    }
+
+    @Override
+    Object getValue(View view) throws InvocationTargetException, IllegalAccessException {
+      return mField.get(view);
+    }
+  }
+
+  private class MethodBackedCSSProperty extends ViewCSSProperty {
+    private Method mMethod;
```
rickbrew: `final`

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -80,4 +332,54 @@ private static String capitalize(String str) {
...
+  private abstract class ViewCSSProperty {
+    private String mName;
+    private ViewDebug.ExportedProperty mAnnotation;
+
+    public ViewCSSProperty(String name, @Nullable ViewDebug.ExportedProperty annotation) {
+      mName = convertGetterNameToCssName(name);
+      mAnnotation = annotation;
+    }
+
+    String getName() {
```
rickbrew: these methods should be `public`

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -80,4 +332,54 @@ private static String capitalize(String str) {
...
+    Object getValue(View view) throws InvocationTargetException, IllegalAccessException {
+      return mMethod.invoke(view);
+    }
+  }
+
+  private abstract class ViewCSSProperty {
+    private String mName;
+    private ViewDebug.ExportedProperty mAnnotation;
+
+    public ViewCSSProperty(String name, @Nullable ViewDebug.ExportedProperty annotation) {
```
rickbrew: Why do we allow the annotation to be null? I thought we only process things which have this annotation?

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -80,4 +332,54 @@ private static String capitalize(String str) {
...
+      mName = convertGetterNameToCssName(name);
```
rickbrew: Doing this conversion in the constructor can make things confusing. Much prefer to have the caller do this.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -80,4 +332,54 @@ private static String capitalize(String str) {
...
+      mAnnotation = annotation;
+    }
+
+    String getName() {
+      return mName;
+    }
+
+    abstract Object getValue(View view) throws InvocationTargetException, IllegalAccessException;
```
rickbrew: do we still need the `IllegalAccessException` if we're always calling `setAccessible(true)` ?

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -69,7 +143,185 @@ private static String getIdAttribute(View element) {
...
+
+      if (i < words.length - 1) {
+        result.append('-');
+      }
+    }
+
+    return result.toString();
+  }
+
+  private void convertValueToCSSString(
```
rickbrew: Since this method is doing the `StyleAccumulator.store()` call, it's not really converting a value to a "CSS string". Also, the pattern for methods that take an `Accumulator` is to use "get" for the verb. So maybe we just call this method `getStyleFromPropertyValue`.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -69,7 +143,185 @@ private static String getIdAttribute(View element) {
...
+    if (name.equals(ID_NAME)) {
+      convertIdToCSSString(element, styles);
+    } else if (value instanceof Integer) {
+      convertIntegerToCSSString(name, (Integer) value, annotation, styles);
+    } else if (value instanceof Float) {
+      convertFloatToCSSString(name, (Float) value, annotation, styles);
+    }
+  }
+
+  private void convertIdToCSSString(
```
rickbrew: this method doesn't take an ID and convert it to a CSS String. A more appropriate name is probably just `getIdStyle`

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
ichub: For future whitelisting (allow-listing :stuck_out_tongue:) of methods/fields that do not have this annotation.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
ichub: Compiler doesn't know that we call that method, so yeah.

stetho/src/main/java/com/facebook/stetho/inspector/elements/Origin.java
```diff
@@ -0,0 +1,30 @@
...
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.elements;
+
+import com.facebook.stetho.json.annotation.JsonValue;
+
+public enum Origin {
+  Injected("injected"),
```
rickbrew: in java, convention is that enum values are ALL_CAPITALS

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -10,19 +10,97 @@
...
+
+  private Pattern getWordBoundaryPattern() {
+    if (mWordBoundaryPattern == null) {
+      mWordBoundaryPattern = Pattern.compile("(?<=\\p{Lower})(?=\\p{Upper})");
+    }
+
+    return mWordBoundaryPattern;
+  }
+
+  private Set<ViewCSSProperty> getProperties() {
```
rickbrew: I'm a big fan of symmetry/consistency in the field name and method name ... so I recommend renaming this method to `getViewProperties` (alternately you could rename the field to `mProperties` I suppose)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -69,7 +147,176 @@ private static String getIdAttribute(View element) {
...
   @Override
   public View getViewForHighlighting(Object element) {
-    return (View)element;
+    return (View) element;
+  }
+
+  @Override
+  protected void onGetStyles(View element, StyleAccumulator styles) {
+
+    for (ViewCSSProperty property : getProperties()) {
```
rickbrew: Have you considered making `mViewProperties` an array instead of a `Set`? I don't think we actually need the set mechanics, and it would allow us to avoid an allocation here in the for-each block. And since `HashSet<T>` is implemented as a wrapper around a `HashMap<T, null>', we end up with an additional allocation for every item we put into it (`Map.Entry`).

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -69,7 +147,176 @@ private static String getIdAttribute(View element) {
...
+      try {
+        getCSSStringFromValue(
+            element,
+            property.getCSSName(),
+            property.getValue(element),
+            property.getAnnotation(),
+            styles);
+      } catch (IllegalAccessException | InvocationTargetException e) {
+        LogUtil.e("failed to get the style of element id=" + getIdAttribute(element));
```
rickbrew: 1. Don't forget to pass the exception into `LogUtil.e`
2. You should include something about the property (the name should work :-) ). Otherwise if 10 properties on 1 element fail we'll get 10 identical things printed to logcat. Also, if only 1 of them fails then it'll be easier to debug if we know which property isn't working.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -80,4 +327,60 @@ private static String capitalize(String str) {
     buffer.setCharAt(0, Character.toTitleCase(buffer.charAt(0)));
     return buffer.toString();
   }
+
+  private class FieldBackedCSSProperty extends ViewCSSProperty {
```
rickbrew: these classes can be made final

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -80,4 +327,60 @@ private static String capitalize(String str) {
...
+  private abstract class ViewCSSProperty {
+    private final String mCSSName;
+    private final ViewDebug.ExportedProperty mAnnotation;
+
+    public ViewCSSProperty(String cssName, @Nullable ViewDebug.ExportedProperty annotation) {
+      mCSSName = cssName;
+      mAnnotation = annotation;
+    }
+
+    public String getCSSName() {
```
rickbrew: we should make this final, since our specific intention is to *not* allow derived classes to override this behavior

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -80,4 +327,60 @@ private static String capitalize(String str) {
...
+      return mCSSName;
+    }
+
+    abstract Object getValue(View view) throws InvocationTargetException, IllegalAccessException;
+
+    @Nullable ViewDebug.ExportedProperty getAnnotation() {
```
rickbrew: public and final

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -80,4 +327,60 @@ private static String capitalize(String str) {
...
+    public ViewCSSProperty(String cssName, @Nullable ViewDebug.ExportedProperty annotation) {
+      mCSSName = cssName;
+      mAnnotation = annotation;
+    }
+
+    public String getCSSName() {
+      return mCSSName;
+    }
+
+    abstract Object getValue(View view) throws InvocationTargetException, IllegalAccessException;
```
rickbrew: public

stetho/src/main/java/com/facebook/stetho/inspector/helper/IntegerFormatter.java
```diff
@@ -0,0 +1,54 @@
...
+    @TargetApi(Build.VERSION_CODES.LOLLIPOP)
+    public String format(Integer integer, @Nullable ViewDebug.ExportedProperty annotation) {
+      if (annotation != null && annotation.formatToHexString()) {
+        return "0x" + Integer.toHexString(integer);
+      }
+
+      return super.format(integer, annotation);
+    }
+  }
+}
```
rickbrew: we need a newline here. some tools freak out without it, apparently. http://stackoverflow.com/questions/729692/why-should-files-end-with-a-newline

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -33,18 +49,287 @@ public void disable(JsonRpcPeer peer, JSONObject params) {
...
+
+    Selector selector = new Selector();
+    selector.value = "<this_element>";
+
+    CSSRule rule = new CSSRule();
+
+    rule.origin = Origin.Regular;
+    rule.selectorList = new SelectorList();
+
+    rule.selectorList.selectors = new ArrayList<>();
```
rickbrew: is it possible to use `ListUtil.newImmutableList(selector)` here?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -33,18 +49,287 @@ public void disable(JsonRpcPeer peer, JSONObject params) {
...
+    rule.selectorList.selectors.add(selector);
+
+    rule.style = new CSSStyle();
+    rule.style.cssProperties = new ArrayList<>();
+
+    match.rule = rule;
+
+    rule.style.shorthandEntries = new ArrayList<>();
```
rickbrew: since we aren't putting anything in here we can use `Collections.emptyList()` (save 2 allocations)

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -33,18 +49,287 @@ public void disable(JsonRpcPeer peer, JSONObject params) {
...
+                  property.value = value;
+
+                  match.rule.style.cssProperties.add(property);
+                }
+              }
+            });
+      }
+    });
+
+    result.inherited = new ArrayList<>();
```
rickbrew: `Collections.emptyList`

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -33,18 +49,287 @@ public void disable(JsonRpcPeer peer, JSONObject params) {
...
+    result.pseudoElements = new ArrayList<>();
```
rickbrew: `Collections.emptyList`

rickbrew:
Okay there's one more round of feedback. Looking pretty good, getting much closer :)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
ichub: I don't think it's possible to get rid of allocations for every element since we don't know how many `@ViewDebug.ExportedProperty` annotated fields and methods there are at compile time.

jasta:
I'm officially going to wait on 1.2.0 for this feature.  It's too good to pass up and we're getting so close :)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -69,7 +145,177 @@ private static String getIdAttribute(View element) {
...
+    for (ViewCSSProperty property : getProperties()) {
+      try {
+        getCSSStringFromValue(
+            element,
+            property.getCSSName(),
+            property.getValue(element),
+            property.getAnnotation(),
+            styles);
+      } catch (IllegalAccessException | InvocationTargetException e) {
+        LogUtil.e(e, "failed to get style property" + property.getCSSName() +
```
rickbrew: missing a space at the end of "failed to get style property"

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -69,7 +145,177 @@ private static String getIdAttribute(View element) {
...
+            " of element id=" + getIdAttribute(element));
```
rickbrew: I believe elements can end up with the same ID value, or even just `View.NO_ID` (which is zero or -1, I forget), so this won't always give enough context to track down an error. I recommend using element.toString(), which will include the ID and other stuff which Android UI devs should be accustomed to parsing. e.g.

```
LogUtil.e(e, "failed to get style property " + property.getCSSName() +
    "of element " + element.toString());
```

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -69,7 +145,177 @@ private static String getIdAttribute(View element) {
...
+      }
+    }
+  }
+
+  private static boolean canIntBeMappedToString(@Nullable ViewDebug.ExportedProperty annotation) {
+    return annotation != null && annotation.mapping() != null && annotation.mapping().length > 0;
```
rickbrew: can we match the indentation style used in `canFlagsBeMappedToString` ?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -33,18 +50,286 @@ public void disable(JsonRpcPeer peer, JSONObject params) {
...
+
+    rule.style.shorthandEntries = Collections.emptyList();
+
+    mDocument.postAndWait(new Runnable() {
+      @Override
+      public void run() {
+        Object elementForNodeId = mDocument.getElementForNodeId(request.nodeId);
+
+        if (elementForNodeId == null) {
+          LogUtil.e("Failed to get style of an element that does not exist, nodeid=" +
```
rickbrew: since this is technically possible due to a race condition, I recommend using warning level instead of error level, e.g. `LogUtil.w`

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -69,7 +145,179 @@ private static String getIdAttribute(View element) {
...
   @Override
   public View getViewForHighlighting(Object element) {
-    return (View)element;
+    return (View) element;
+  }
+
+  @Override
+  protected void onGetStyles(View element, StyleAccumulator styles) {
+
+    for (ViewCSSProperty property : getViewProperties()) {
```
rickbrew: because this is a `List`, we can use the iterator-less pattern and save an allocation:
```
List<ViewCSSProperty> viewProperties = getViewProperties();
for (int i = 0, size = viewProperties.size(); i < size; ++i) {
  final ViewCSSProperty property = viewProperties.get(i);
  ...
}
```

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -69,7 +145,179 @@ private static String getIdAttribute(View element) {
...
+
+      if (i < words.length - 1) {
+        result.append('-');
+      }
+    }
+
+    return result.toString();
+  }
+
+  private void getCSSStringFromValue(
```
rickbrew: `getStyleFromValue` might be a better choice here since we're not getting a "css string". same goes for the other `getCSSStringFrom___` methods

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -80,4 +328,61 @@ private static String capitalize(String str) {
...
+      mCSSName = cssName;
+      mAnnotation = annotation;
+    }
+
+    public final String getCSSName() {
+      return mCSSName;
+    }
+
+    public abstract Object getValue(View view)
+        throws InvocationTargetException, IllegalAccessException;
```
rickbrew: since we always call `setAccessible(true)`, do we still need `IllegalAccessException` in the signature ?

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -69,7 +145,179 @@ private static String getIdAttribute(View element) {
...
+
+  private static String mapFlagsToStringUsingAnnotation(
+      int value,
+      @Nullable ViewDebug.ExportedProperty annotation) {
+
+    if (!canFlagsBeMappedToString(annotation)) {
+      throw new IllegalStateException("Cannot map using this annotation");
+    }
+
+    StringBuilder stringBuilder = new StringBuilder();
```
rickbrew: We can optimize away the allocation of a `StringBuilder` in the case where `atLeastOneFlag` ends up being false. We can do this by initializing `stringBuilder` to null here, and then lazily instantiating it in the loop below if it's null.

rickbrew:
Looks good to me! Squash 'n merge :+1:

felipecsl:
regarding this item: 
`In Chrome you can edit these CSS properties, but here you can't. Maybe a future project? Would probably relatively simple to add.`

I'm really interested on this. Is this something you guys or someone is already tackling? Otherwise, do know what's the correct chrome devtools method used for this (and how to tell it that a specific property is editable)?

ichub:
This file contains the entire chrome dev tools protocol: https://code.google.com/p/chromium/codesearch#chromium/src/third_party/WebKit/Source/devtools/protocol.json&q=protocol.json&sq=package:chromium&type=cs

It can also be found in the directory https://github.com/facebook/stetho/tree/master/build-tools (along with a script I wrote that generates stubs for methods defined in protocol.json).

The methods I used to get styles include `getMatchedStylesForNode` (line 3130) and `getComputedStylesForNode` (line 3115). There seem to be methods for setting style text too, for example `addRule` (line 3240).

The way I'd do this is implement the relevant method stubs in `CSS.java`, and see if they actually try do what is expected. If not, search around more in the protocol for methods that seem relevant, rinse and repeat.

good luck 👍

felipecsl:
@ichub yeah I played with it last night and apparently the method is `CSS.setPropertyText`. However, that is not present in the protocol specification json you mentioned above (I was using it already BTW), which is weird. The closest match I found is `setStyleText` or `setStyleSheetText` and the request payload seems to be the same. I'll keep digging, but I was able to make it work by setting a `stylesheetId` into the returned `CSSStyle` object from `getMatchedStylesForNode`.

ichub:
Have you tried `setEffectivePropertyValueForNode` (line 3268)?

```
{
    "name": "setEffectivePropertyValueForNode",
    "parameters": [
        { "name": "nodeId", "$ref": "DOM.NodeId", "description": "The element id for which to set property." },
        { "name": "propertyName", "type": "string"}
        { "name": "value", "type": "string"}
    ],
    "description": "Find a rule with the given active property for the given node and set the new value for this property",
    "hidden": true
}
```

felipecsl:
Not really. All I did is go to the CSS panel in the Elements tab and try to change one of the properties (eg.: alpha). What happens is that DevTools calls `CSS.setPropertyText` immediately and that was my starting point

ichub:
What kind of information is passed along to you in `setStyleText`, ie.`styleSheetId`, `range`, and `text`? Is it garbage or useful? I know that we don't use stylesheets to send over styles to chrome since there android elements aren't styled using stylesheets. How does Chrome react to that?

felipecsl:
`styleSheetId` is the same id that you returned from `getMatchedStylesForNode`. `range` doesn't seem to be always sent, I can't remember exactly, and `text` is in the format `property: value;`, for example: `height: 145px;' so devtools seem to be working correctly, i haven't seen any garbage in there.
The only problem I've seen is taht it doesnt always call `CSS.setPropertyText`, sometimes you change the property and nothing happens. I **think** that may be happening because I'm not returning anything from that method and DevTools might be waiting for a response from a previous request and then timing out.

Issue #258, opened 2015-08-15
Ability to add Stetho protocol/domain library extensions


sromku:
Chrome protocol calls are designed to be coupled to `ChromeDevtoolsDomain` implementations and it is impossible to add more functionality to existing domains/modules. 

For example there is a `Database` class that has methods: `getDatabaseTableNames` and `executeSQL`.
Because of `MethodDispatcher` implementation and [dispatch table map](https://github.com/facebook/stetho/blob/master/stetho%2Fsrc%2Fmain%2Fjava%2Fcom%2Ffacebook%2Fstetho%2Finspector%2FMethodDispatcher.java#L39), the calls will look for specific class name and for method as [defined in protocol](https://code.google.com/p/chromium/codesearch#chromium/src/third_party/WebKit/Source/devtools/protocol.json&l=1668). This is very coupled which is right for low level protocol wrapping. 

But, this limits the ability to extend and add new functionality to already covered domains. Check this screenshot. I had to make some not healthy implementations to make it work. The **Web SQL** tab shows all Android and custom made content providers. Since Android content providers like *Calendar, Contacts, ...* aren't accessible through raw sqlite database, I had to create new `Database` class (same name) with more extended functionality and override Stetho Database class in dispatch table map. 

![Screenshot](https://github.com/EverythingMe/easy-content-providers/wiki/images/stetho-providers.png)
This is the repo with new stetho extension: https://github.com/EverythingMe/easy-content-providers

Here, I would like to suggest ability to add external extensions to Stetho library and discuss it :) 

I can suggest some PR for better discussion. For example adding some `ChromeDevtoolsDomainListener` and then be able to add functionality to existing domains. Or suggest another level of abstraction in addition to `ChromeDevtoolsDomain` that will support extensions. 

What do you think?

longinoa:
@sromku What do you men to extend the protocol? Are you also modifying the implementation of the Chrome Dev tools in chrome?

I think adding content providers would be pretty awesome but Im just not sure I follow the protocol extension.

sromku:
@longinoa What I mean is not by modifying the protocol itself, but adding an option to the Stetho lib. When Chrome invokes protocol method like `getDatabaseTableNames`, `Database` class is the only one to get this call. I suggest, to add an option of registering listeners and be notified about Chrome calls.

I will create PR with some suggested solution/direction and provide some new API snippets. This will help me to explain it better :)

auchenberg:
[...]

PR #264, opened 2015-08-26
Added support of Database Domain extensions

sromku:
Based on ticket #258


stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Database.java
```diff
@@ -244,4 +272,34 @@ public ExecuteSQLResponse handleUpdateDelete(int count) throws SQLiteException {
...
+
+  public static abstract class DatabasePeer {
+
+    protected Context mContext;
+
+    public DatabasePeer(Context context) {
+      mContext = context;
+    }
+
+    protected abstract void onRegistered(JsonRpcPeer peer);
```
jasta: I'm not sure it's required to make this abstraction work at the peer manager level.  Probably good enough to just have an onFirstPeerConnected and onLastPeerDisconnected method, no?

stetho-sample/src/debug/java/com/facebook/stetho/sample/CalendarProviderDatabasePeer.java
```diff
@@ -0,0 +1,106 @@
...
+
+  private final static String sDatabaseName = "provider-calendars";
+
+  public CalendarProviderDatabasePeer(Context context) {
+    super(context);
+  }
+
+  @Override
+  protected void onRegistered(JsonRpcPeer peer) {
+    Database.DatabaseObject databaseParams = new Database.DatabaseObject();
```
jasta: This junk should not be leaked into the implementor.  Strongly prefer a simpler mechanism of listing the databases.  The timing and behaviour of this should be managed more internally to the Database module and/or the actual peer manager.

stetho-sample/src/debug/java/com/facebook/stetho/sample/CalendarProviderDatabasePeer.java
```diff
@@ -0,0 +1,106 @@
...
+import com.facebook.stetho.inspector.jsonrpc.JsonRpcPeer;
+import com.facebook.stetho.inspector.protocol.module.Database;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * This is example of extending database domain by adding calendar provider
+ */
+public class CalendarProviderDatabasePeer extends Database.DatabasePeer {
```
jasta: `DatabasePeer` is a strange name for this responsibility.  Prefer `DatabaseDriver` or something.  It really has little to do with the Stetho/Chrome peer that has attached and more to do with adapting the Chrome protocol to something implementable inside the context of an Android app.

stetho-sample/src/debug/java/com/facebook/stetho/sample/CalendarProviderDatabasePeer.java
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  @TargetApi(Build.VERSION_CODES.ICE_CREAM_SANDWICH)
+  @Override
+  public <T> T executeSQL(String databaseName, String query, Database.DatabasePeer.ExecuteResultHandler<T> handler) throws SQLiteException {
+
+    // Resolve query and make the right Cursor. Some better query parser should exist. But for now it's fine.
+    Uri uri = null;
+    String[] projection = null;
+    if (query.contains("calendars")) {
```
jasta: Would be really great if we provided a helper that lets you declarative the schema and then just let a generic driver operate on that schema.

For instance, something like:
```java
new ContentProviderSchema.Builder(...)
   .table(new Table.Builder()
       .uri(...)
       .projection(...) // optional
       .build())
   .build();
```

Then you could have an abstract `ContentProviderDatabaseDriver` that takes in the schema.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Database.java
sromku: Please check the latest commit. As you suggested, implementor shouldn't implement this class anymore.

stetho-sample/src/debug/java/com/facebook/stetho/sample/CalendarProviderDatabasePeer.java
sromku: You're right. This code was shouldn't be used by library consumer. The latest commit based on your suggestion doesn't have this option anymore.

stetho-sample/src/debug/java/com/facebook/stetho/sample/CalendarProviderDatabasePeer.java
sromku: I assumed that you would come with better name :) Changed.

stetho-sample/src/debug/java/com/facebook/stetho/sample/CalendarProviderDatabasePeer.java
sromku: In the last update, I changed to what you suggested.

sromku:
@jasta Thanks for your review! :). So for the *round 2* of reviews these are the changes that I made based on your comments. I hope, it is the way you intended, if not, please correct me.
1. Now, there is only two drivers: `SqliteDatabaseDriver` and `ContentProviderDatabaseDriver`
2. User will create as many as he/she wants of `ContentProviderSchema`s and register them to `ContentProviderDatabaseDriver`. It means, that user shouldn't implement some low level logic anymore.

In the sample project, the `CalendarContentProvide..` class was deleted and `SampleDebugApplication` was updated.

stetho-sample/src/debug/java/com/facebook/stetho/sample/SampleDebugApplication.java
```diff
@@ -42,7 +53,60 @@ private void initializeStetho(final Context context) {
...
+                    CalendarContract.Events.DESCRIPTION,
+                    CalendarContract.Events.ACCOUNT_NAME,
+                    CalendarContract.Events.DTSTART,
+                    CalendarContract.Events.DTEND,
+                    CalendarContract.Events.CALENDAR_ID,
+                })
+                .build())
+            .build();
+
+        Database database = new Database();
```
jasta: Sorry for the late review, I was working on getting stetho-1.2.0 out with a much improved API to support this kind of customization.  I think we can now approach this with a new, much stronger API:

    new DefaultInspectorModulesProvider()
      .provideDatabaseDriver(new ContentProviderDatabaseDriver(...))
      .finish();

This implies that the default provider is install installed. If the default is to be configured, they can use the existing (as of 1.2.0) method .databaseFiles(...) and just provide a no-op files lister.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Database.java
```diff
@@ -46,36 +45,46 @@
...
-   * @param context the context
+   * Constructs the object.
    */
-  @Deprecated
-  public Database(Context context) {
-    mDatabasePeerManager = new DatabasePeerManager(context);
+  public Database() {
+    mDatabaseDrivers = new ArrayList<>();
+    mChromePeerManager = new ChromePeerManager();
+    mChromePeerManager.setListener(new PeerRegistrationListener() {
```
jasta: nit, prefer this in a named subclass or a member variable for improved fluency:

    mChromePeerManager.setListener(mPeerListener);

stetho/src/main/java/com/facebook/stetho/inspector/database/SqliteDatabaseDriver.java
```diff
@@ -82,9 +80,20 @@ private void bootstrapNewPeer(JsonRpcPeer peer) {
       eventParams.database = databaseParams;
 
       peer.invokeMethod("Database.addDatabase", eventParams, null /* callback */);
```
jasta: I'd really prefer not to repeat this method.  Can we refine the driver implementation to something more database-focused and use a generalized harness to interact with Chrome DevTools:

```java
interface DatabaseDriver {
  List<String> getDatabaseNames();
  List<String> getTableNames(String databaseName);
  ExecuteSQLResponse executeSQL(String database, String query, ExecuteResultHandler<ExecuteSQLResponse>);
}
```

We may even wish to generalize `ExecuteResultHandler` a bit so that it's not so directly cursor-focused but rather works more closely as an abstraction of how Chrome DevTools interprets the response.

jasta:
This is looking really great.  I think with a couple of refinements we'd be ready to go.  Sorry again for the delay :(

sromku:
Thanks for your comment @jasta. Sorry for long response to your response :) Just couldn't leave this beach https://www.facebook.com/pages/Paralia-Stefanou-Seitan-limania-Chania/163349380464470 on my vacation. Now, back to work :) I will take 1.2.0 and make the necessary updates based on latest lib changes and your comments. Will update you soon. Thanks.

stetho/src/main/java/com/facebook/stetho/inspector/database/SqliteDatabaseDriver.java
sromku: @jasta the `DatabaseDriver` is an abstract class now and all registration databases logic to DevTools is now part of this class and not in implementators. Which means that `peer.invokeMethod("Database.addDatabase", eventParams, null /* callback */);` is part of this class only and the subclasses don't know what is `JsonRpcPeer` is. Which is good I think. The method names was changes as you suggested and they are abstract. 

Regarding the `ExecuteResultHandler`, I'm not sure I understood you, but as it looks from code the generics can be removed at all in this interface. It could be strictly types like this: 
``` java
public interface ExecuteResultHandler {
  ExecuteSQLResponse handleRawQuery() throws SQLiteException;
  ExecuteSQLResponse handleSelect(Cursor result) throws SQLiteException;
  ExecuteSQLResponse handleInsert(long insertedId) throws SQLiteException;
  ExecuteSQLResponse handleUpdateDelete(int count) throws SQLiteException;
}
```
For this interface, I didn't make this change. Let me know if you want to remove the `<T>` at all. wdyt?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Database.java
sromku: Done

stetho-sample/src/debug/java/com/facebook/stetho/sample/SampleDebugApplication.java
sromku: Done

sromku:
@jasta Please check latest changes. The changes are based on your last comments.
 - `DefaultInspectorModulesBuilder ` class has new method *`provideDatabaseDriver(DatabaseDriver)`* to support more cleaner API.
 - `DatabaseDriver` is responsible for invoking DevTools peer event for database registration instead of duplicated code in each of subclasses.
 - `mChromePeerManager.setListener(mPeerListener);` is used as you suggested. The anonymous listener class was changed to inner static one.

sromku:
@jasta do you want me to make any changes?

stetho-sample/src/debug/java/com/facebook/stetho/sample/SampleDebugApplication.java
```diff
@@ -42,7 +49,56 @@ private void initializeStetho(final Context context) {
                 .finish();
           }
         })
-        .enableWebKitInspector(Stetho.defaultInspectorModulesProvider(context))
+        .enableWebKitInspector(new ExtInspectorModulesProvider(context))
```
jasta: Not sure why we need a separate class here.  It should be possible to write:

```java
.enableWebKitInspector(
  new DefaultInspectorModulesBuilder()
    .databaseDriver(createBlaBla())
    .finish());
```

This was the original intention of the new builder API at least.

stetho-sample/src/debug/java/com/facebook/stetho/sample/SampleDebugApplication.java
```diff
@@ -42,7 +49,56 @@ private void initializeStetho(final Context context) {
...
+    private Context mContext;
+
+    ExtInspectorModulesProvider(Context context) {
+      mContext = context;
+    }
+
+    @Override
+    public Iterable<ChromeDevtoolsDomain> get() {
+      return new Stetho.DefaultInspectorModulesBuilder(mContext)
+          .provideDatabaseDriver(createContentProviderDatabaseDriver(mContext))
```
jasta: nit: prefer less verbose method names, such as `databaseDriver`.  Note the others: `documentProvider`, `databaseFiles`, etc.

stetho-sample/src/debug/java/com/facebook/stetho/sample/SampleDebugApplication.java
```diff
@@ -42,7 +49,56 @@ private void initializeStetho(final Context context) {
...
+                  CalendarContract.Calendars.IS_PRIMARY,
+              })
+              .build())
+          .build();
+
+      // sample events content provider we want to support
+      ContentProviderSchema eventsSchema = new ContentProviderSchema.Builder()
+          .table(new Table.Builder()
+              .uri(CalendarContract.Events.CONTENT_URI)
+              .projection(new String[]{
```
jasta: Why not use the default "null" projection?  It's almost assuredly what folks will want for their own apps.

stetho/src/main/java/com/facebook/stetho/Stetho.java
```diff
@@ -290,7 +294,21 @@ public DefaultInspectorModulesBuilder runtimeRepl(RuntimeReplFactory factory) {
...
     public DefaultInspectorModulesBuilder databaseFiles(DatabaseFilesProvider provider) {
-      mDatabaseFiles = provider;
+      mDatabaseFilesProvider = provider;
+      return this;
+    }
+
+    /**
+     * Extend and provide additional database drivers. Currently two database drivers are supported
+     * in this lib: <br>
+     *   1. <code>SqliteDatabaseDriver</code> - Presents sqlite databases and all tables of the app.<br>
```
jasta: nit: instead of `<br>` we should use:

```html
<ol>
<li>...</li>
</ol>
```

This will automatically create a new paragraph block.

stetho/src/main/java/com/facebook/stetho/Stetho.java
```diff
@@ -290,7 +294,21 @@ public DefaultInspectorModulesBuilder runtimeRepl(RuntimeReplFactory factory) {
...
+     *   2. <code>ContentProviderDatabaseDriver</code> - Configure and present content providers data.
+     */
+    public DefaultInspectorModulesBuilder provideDatabaseDriver(Database.DatabaseDriver databaseDriver) {
+      if (mDatabaseDrivers == null) {
+        mDatabaseDrivers = new ArrayList<>();
```
jasta: Oh, that's why you used the word `provide` here, touché :)

Prefer instead `addDatabaseDriver` then?

stetho/src/main/java/com/facebook/stetho/Stetho.java
```diff
@@ -347,12 +365,17 @@ public DefaultInspectorModulesBuilder remove(String moduleName) {
...
-                mDatabaseFiles != null ?
-                    mDatabaseFiles :
-                    new DefaultDatabaseFilesProvider(mContext)));
+        Database database = new Database();
+        database.add(new SqliteDatabaseDriver(mContext,
+            mDatabaseFilesProvider != null ?
+                mDatabaseFilesProvider :
+                new DefaultDatabaseFilesProvider(mContext)));
+        if (mDatabaseDrivers != null) {
+          for (Database.DatabaseDriver databaseDriver : mDatabaseDrivers) {
```
jasta: nit: this syntax evaluates to a slow expression on Dalvik devices.  I usually get in the habit of writing:

```java
for (int i = 0, N = mDatabaseDrivers.size(); i < N; i++) {
  database.add(mDatabaseDrivers.get(i));
}
```

It's almost the same amount of typing but doesn't allocate the iterator and the pattern also works to iterate things like `SimpleArrayMap` and `SparseArray`.

stetho/src/main/java/com/facebook/stetho/inspector/database/ContentProviderDatabaseDriver.java
```diff
@@ -0,0 +1,94 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+@ThreadSafe
+public class ContentProviderDatabaseDriver extends Database.DatabaseDriver {
+
+  private final static String sDatabaseName = "content-providers";
+
+  private final ContentProviderSchema[] mContentProviderSchemas;
```
jasta: It seems we allow this to be `@Nullable` but I'm not sure why that would be desirable.  It seems like it should fail in the constructor if it's null or empty?

stetho/src/main/java/com/facebook/stetho/inspector/database/ContentProviderDatabaseDriver.java
```diff
@@ -0,0 +1,94 @@
...
+    // find the right ContentProviderSchema
+    int index = mTableNames.indexOf(tableName);
+    ContentProviderSchema contentProviderSchema = mContentProviderSchemas[index];
+
+    // execute the query
+    ContentResolver contentResolver = mContext.getContentResolver();
+    Cursor cursor = contentResolver.query(
+        contentProviderSchema.getUri(),
+        contentProviderSchema.getProjection(),
+        null,
```
jasta: style nit: prefer `null /* argumentName */`.  The same generally for true/false and other context-less literals.  It's just a little easier for the reviewer...

stetho/src/main/java/com/facebook/stetho/inspector/database/ContentProviderDatabaseDriver.java
```diff
@@ -0,0 +1,94 @@
...
+  /**
+   * Fetch the table name from query
+   */
+  private String fetchTableName(String query) {
+    for (String tableName : mTableNames) {
+      if (query.contains(tableName)) {
+        return tableName;
+      }
+    }
+    return "";
```
jasta: Prefer null and to mark this method `@Nullable`.  Using magic string constants like "" can end up very hard to track down when bugs occur.

jasta:
This is very well done, thank you.  I again apologize for such a long delay in review.

After getting through the full diff series I really only have nit-pick comments and I'm happy to just take those as notes to follow-up myself to do some tidy work or if you want you can follow-up after the fact.  I otherwise am comfortable merging as-is.

stetho-sample/src/debug/java/com/facebook/stetho/sample/SampleDebugApplication.java
sromku: Yes, it's right. The issue, is with a lot of selected columns. For example in android content provider of `event ` there is like more than 20 columns. The problem is that the view of table in chrome makes everything shrinked due to luck of space and it's impossible to hide or resize columns in such case. And also, not all columns have interesting info. Beside this, I have no other better explanation of selecting only few columns in this sample, so using  `.defaultProjection()` is also fine :)

stetho/src/main/java/com/facebook/stetho/Stetho.java
sromku: Yep. Some time ago when I used Eclipse, java docs were really useful and presented nicely and formatted. Then I used correct tags. In Android Studio, it's just not the same. But, ye, for java docs generation it's important I think. Will use the right tags in the future pull requests ;) Thanks.

stetho/src/main/java/com/facebook/stetho/Stetho.java
sromku: :+1: fine by me

stetho/src/main/java/com/facebook/stetho/Stetho.java
sromku: Good advice! Thanks! Will use this approach.

stetho/src/main/java/com/facebook/stetho/inspector/database/ContentProviderDatabaseDriver.java
sromku: There is null checks later in code. But if we want to prevent this from being a null, then throwing runtime exception from constructor can work I think.

stetho/src/main/java/com/facebook/stetho/inspector/database/ContentProviderDatabaseDriver.java
sromku: Right, good stuff! Will adopt it in out internal company code reviews :)

stetho/src/main/java/com/facebook/stetho/inspector/database/ContentProviderDatabaseDriver.java
sromku: Right, agree.

sromku:
@jasta Thanks for your comments and review. Glad that this PR made it's way to the master :) We are using stetho in debug build types and this is one of the helpful and time saving tools!

jasta:
@sromku of course, totally my fault making you wait so long.  We're looking at a 1.3.0 release this month so it will be nice to finally get this turned around for you in an official release :)

sromku:
Much thanks! :)

PR #265, opened 2015-09-01
Further refine DefaultInspectorModulesBuilder

jasta:
As the DevTools implementation has matured it has become clear that the
modules are not truly modular or interchangable at all and have heavy
dependencies between the two that make it impossible to provide or
configure them in isolation.  Because of this, we're moving to a more
explicit customization API which is responsible for managing the
dependencies of the implementation directly while still preserving the
level of abstractions already present in the system (such as providing a
custom document model, runtime REPL, etc).

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/Stetho.java
+++ b/stetho/src/main/java/com/facebook/stetho/Stetho.java
@@ -12,6 +12,8 @@ package com.facebook.stetho;
 import com.facebook.stetho.dumpapp.plugins.CrashDumperPlugin;
 import com.facebook.stetho.dumpapp.plugins.FilesDumperPlugin;
 import com.facebook.stetho.dumpapp.plugins.HprofDumperPlugin;
+import com.facebook.stetho.inspector.console.RuntimeReplFactory;
+import com.facebook.stetho.inspector.database.DatabaseFilesProvider;
 import com.facebook.stetho.inspector.database.DefaultDatabaseFilesProvider;
 import javax.annotation.Nullable;
 
@@ -36,6 +38,7 @@ import com.facebook.stetho.dumpapp.plugins.SharedPreferencesDumperPlugin;
 import com.facebook.stetho.inspector.ChromeDevtoolsServer;
 import com.facebook.stetho.inspector.ChromeDiscoveryHandler;
 import com.facebook.stetho.inspector.elements.Document;
+import com.facebook.stetho.inspector.elements.DocumentProviderFactory;
 import com.facebook.stetho.inspector.elements.android.ActivityTracker;
 import com.facebook.stetho.inspector.elements.android.AndroidDOMConstants;
 import com.facebook.stetho.inspector.elements.android.AndroidDocumentProviderFactory;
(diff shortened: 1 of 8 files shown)


stetho/src/main/java/com/facebook/stetho/Stetho.java
```diff
@@ -234,14 +243,65 @@ public DefaultDumperPluginsBuilder remove(String pluginName) {
...
+    }
+
+    /**
+     * Provide a custom document model factory which can operate on the logical DOM exposed to
+     * Chrome in the Elements tab.  An Android View hierarchy instance is provided by
+     * default if this method is not called.
+     * <p />
+     * <i>Experimental.</i>  This API may be changed or removed in the future.
+     */
+    public DefaultInspectorModulesBuilder documentModel(DocumentProvider.Factory factory) {
```
rickbrew: `documentProvider()`

stetho/src/main/java/com/facebook/stetho/Stetho.java
```diff
@@ -274,13 +340,33 @@ public DefaultInspectorModulesBuilder remove(String moduleName) {
...
+                mDatabaseFiles != null ?
+                    mDatabaseFiles :
+                    new DefaultDatabaseFilesProvider(mContext)));
       }
       return mDelegate.finish();
     }
+
+    @Nullable
+    private DocumentProvider.Factory resolveDocumentModel() {
+      if (mDocumentModel != null) {
```
rickbrew: model vs. provider

stetho/src/main/java/com/facebook/stetho/Stetho.java
rickbrew: and let's rename `DocumentProvider.Factory` to `DocumentProviderFactory`

stetho/src/main/java/com/facebook/stetho/Stetho.java
```diff
@@ -234,14 +243,65 @@ public DefaultDumperPluginsBuilder remove(String pluginName) {
...
 
+  /**
+   * Configuration mechanism to customize the behaviour of the standard set of inspector
+   * modules satisfying the Chrome DevTools protocol.  Note that while it is still technically
+   * possible to manually control these modules, this API is strongly discouraged and will not
+   * necessarily be supported in future releases.
+   */
   public static final class DefaultInspectorModulesBuilder {
-    private final Context mContext;
+    private final Application mContext;
```
rickbrew: if this is now an `Application`, why is the field called `mContext` ?

stetho/src/main/java/com/facebook/stetho/Stetho.java
```diff
@@ -234,14 +243,65 @@ public DefaultDumperPluginsBuilder remove(String pluginName) {
...
+    @Nullable private RuntimeReplFactory mRuntimeRepl;
+    @Nullable private DatabaseFilesProvider mDatabaseFiles;
+
     public DefaultInspectorModulesBuilder(Context context) {
-      mContext = context;
+      mContext = (Application)context.getApplicationContext();
+    }
+
+    /**
+     * Provide a custom document model factory which can operate on the logical DOM exposed to
```
rickbrew: "Provides a custom document model ... " <-- should be "document provider", yes?

rickbrew:
Other than a few nits, this looks good to me :+1: Squash 'n merge

stetho/src/main/java/com/facebook/stetho/Stetho.java
jasta: I sometimes do it this way to imply that most users don't actually care that it is an application specifically but at least one user does.  It makes it easier to refactor later when that user is refined or isolated away.

stetho/src/main/java/com/facebook/stetho/Stetho.java
jasta: Ack'd, edited

PR #268, opened 2015-09-03
Fix the Rhino autodetection feature

jasta:
This was accidentally broken in #245 by not modifying the default
inspector modules provider to use the new Runtime(Context) constructor,
effectively disabling the entire feature.


jasta:
@potyl this fixes the most obvious oops in my autodetection feature :)

potyl:
Thanks! That's exactly as one of the commits in my branch.

PR #274, opened 2015-09-04
Fixing NPE crash in ElementInfo.<init>() from Document.createDOMUpdate()

### LINKED ISSUE #272: NPE crash in ElementInfo.<init>
From inspecting the code, a `Descriptor` is emitting a null child which is why the `ElementInfo` constructor is unhappy. My bet is on `DialogFragmentDescriptor` because it doesn't check if `getDialog()` returns null before handing it off to `children.store()`.

```
com.facebook.stetho.common.Util.throwIfNull(Util.java:26)
com.facebook.stetho.inspector.elements.ElementInfo.<init>(ElementInfo.java:29)
com.facebook.stetho.inspector.elements.ShadowDocument$UpdateBuilder.setElementParent(ShadowDocument.java:231)
com.facebook.stetho.inspector.elements.ShadowDocument$UpdateBuilder.setElementChildren(ShadowDocument.java:176)
com.facebook.stetho.inspector.elements.Document.createShadowDOMUpdate(Document.java:283)
com.facebook.stetho.inspector.elements.Document.updateTree(Document.java:298)
com.facebook.stetho.inspector.elements.Document.access$1200(Document.java:30)
com.facebook.stetho.inspector.elements.Document$ProviderListener.onPossiblyChanged(Document.java:622)
com.facebook.stetho.inspector.elements.android.AndroidDocumentProvider$1.run(AndroidDocumentProvider.java:68)
android.os.Handler.handleCallback(Handler.java:733)
android.os.Handler.dispatchMessage(Handler.java:95)
android.os.Looper.loop(Looper.java:136)
android.app.ActivityThread.main(ActivityThread.java:5017)
java.lang.reflect.Method.invokeNative(Native Method:0)
java.lang.reflect.Method.invoke(Method.java:515)
com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:779)
com.android.internal.os.ZygoteInit.main(ZygoteInit.java:595)
dalvik.system.NativeStart.main(Native Method:0)
```

rickbrew:
I'm fixing this by turning it into a soft error. `Descriptor`s shouldn't emit null from their `getChildren` implementation, but it's easy enough to recover from this and at least not crash the app. This is the only place where the children are funneled out of `Descriptor` and into `Document`.

Closes #272

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
@@ -280,10 +280,28 @@ public final class Document extends ThreadBoundProxy {
       NodeDescriptor descriptor = mDocumentProvider.getNodeDescriptor(element);
       mObjectIdMapper.putObject(element);
       descriptor.getChildren(element, childrenAccumulator);
-      updateBuilder.setElementChildren(element, childrenAccumulator);
-      for (int i = 0, N = childrenAccumulator.size(); i < N; ++i) {
-        mCachedUpdateQueue.add(childrenAccumulator.get(i));
+
+      for (int i = 0, size = childrenAccumulator.size(); i < size; ++i) {
+        Object child = childrenAccumulator.get(i);
+        if (child != null) {
+          mCachedUpdateQueue.add(child);
+        } else {
+          // This could be indicative of a bug in Stetho code, but could also be caused by a
+          // custom element of some kind, e.g. ViewGroup. Let's not allow it to kill the hosting
+          // app.
+          LogUtil.e(
+              "%s.getChildren() emitted a null child at position %s for element %s",
+              descriptor.getClass().getName(),
+              Integer.toString(i),
+              element);
+
+          childrenAccumulator.remove(i);
+          --i;
+          --size;
+        }
       }
+
+      updateBuilder.setElementChildren(element, childrenAccumulator);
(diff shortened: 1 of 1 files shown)


stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
```diff
@@ -280,10 +280,28 @@ private void releaseChildrenAccumulator(ArrayListAccumulator<Object> accumulator
...
+        if (child != null) {
+          mCachedUpdateQueue.add(child);
+        } else {
+          // This could be indicative of a bug in Stetho code, but could also be caused by a
+          // custom element of some kind, e.g. ViewGroup. Let's not allow it to kill the hosting
+          // app.
+          LogUtil.e(
+              "%s.getChildren() emitted a null child at position %s for element %s",
+              descriptor.getClass().getName(),
+              Integer.toString(i),
```
jasta: Why not use `%d`? :)

jasta:
lgtm, merge at will

stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
rickbrew: I tend to use %s and then `Integer.toString` because ... less boxing?

PR #275, opened 2015-09-04
Finish renaming DOM to Document where appropriate

### LINKED ISSUE #273: Finish renaming several instances "DOM" to "Document"
We missed a few, e.g. `Document.applyDOMUpdate`, `DOMHiddenView`, `AndroidDOMConstants`, etc.

rickbrew:
We renamed `DOMProvider` to `DocumentProvider` while doing the refactor to extract `Document` out of `DOM`. There were a few places we missed. We also need to use some different terminology within `Document`: instead of saying things like "transmit to the DOM" we need to say "transmit to our listener." In some places it's fine to refer to DOM because it's a reference to the realities of the concrete protocol we're using.

Closes #273


rickbrew:
CC @ichub

ichub:
Thanks (:

jasta:
lgtm, merge at will

Issue #278, opened 2015-09-11
Styles support should show margins


jasta:
The new styles sub-tab feature only shows dimensions and padding, not margins.  This can be accessed from the view's layout parameters (`View#getLayoutParams()`) though some type magic is necessary.

PR #282, opened 2015-09-14
Added margins to css styles tab.

ichub:
I briefly explored doing this in a more generic way using the `ViewDebug.ExportedProperty#deepExport()` option, as it was specifically designed for the case where properties that we care about are nested in fields/method return values. However I found that really the only thing we *would* care about is the margin, so wrangling `deepExport`ed props wouldn't be worth the effort.

EDIT: Actually now I just use `deepExport`ed properties and special case margins so that they are displayed in the box-model.

Here's a screenshot with the margin visible in text form (highlighted by me) and in the box-model:
screen shot 2015-09-15 at 12 10 58 pm

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
@@ -284,6 +284,8 @@ final class ViewDescriptor extends AbstractChainedDescriptor<View> implements Hi
       getStyleFromInteger(name, (Integer) value, annotation, styles);
     } else if (value instanceof Float) {
       getStyleFromFloat(name, (Float) value, annotation, styles);
+    } else {
+      getStylesFromObject(element, name, value, annotation, styles);
     }
   }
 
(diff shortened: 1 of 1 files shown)


stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -331,6 +333,56 @@ private void getStyleFromFloat(
...
+              field.getAnnotation(ViewDebug.ExportedProperty.class);
+
+          getStyleFromValue(
+              view,
+              convertViewPropertyNameToCSSName(propertyName),
+              field.get(value),
+              subAnnotation,
+              styles);
+        }
+      } catch (IllegalAccessException e) {
```
rickbrew: let's wrap only the call(s) that might throw this exception, instead of wrapping almost the entire body of this method with it

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -331,6 +333,56 @@ private void getStyleFromFloat(
...
+              break;
+            case "leftMargin":
+              propertyName = "margin-left";
+              break;
+            case "rightMargin":
+              propertyName = "margin-right";
+              break;
+            default:
+              propertyName =
+                  (annotation.prefix() == null ? "" : annotation.prefix()) + propertyName;
```
rickbrew: prefer to avoid the string concatenation if prefix is null :

```
string annotationPrefix = annotation.prefix();
propertyName = (annotationPrefix == null) ? propertyName : (annotationPrefix + propertyName);
```

rickbrew:
Just two little nits and then it's good :)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -331,6 +333,66 @@ private void getStyleFromFloat(
...
+      Object propertyValue;
+      for (Field field : fields) {
+        try {
+            field.setAccessible(true);
+            propertyValue = field.get(value);
+        } catch (IllegalAccessException e) {
+          LogUtil.e(
+              e,
+              "failed to get property of name: \"" + name + "\" of object: " + String.valueOf(value));
+          propertyValue = null;
```
rickbrew: We can just 'return' here instead of setting `propertyValue` to null. And then we can remove the `if (propertyValue != null)` down below

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -331,6 +333,66 @@ private void getStyleFromFloat(
     styles.store(name, String.valueOf(value), isDefaultValue(value));
   }
 
+  private void getStyleFromObject(
+      View view,
+      String name,
+      Object value,
+      @Nullable ViewDebug.ExportedProperty annotation,
+      StyleAccumulator styles) {
+    if (annotation != null && annotation.deepExport() && value != null) {
```
rickbrew: consider inverting this conditional so that the body of the method isn't indented, like so ...

```
if (annotation == null || !annotation.deepExport() || value == null) {
  return;
}
```

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -331,6 +333,66 @@ private void getStyleFromFloat(
...
+      Field[] fields = value.getClass().getFields();
+
+      Object propertyValue;
```
rickbrew: this variable should definitely be defined inside the loop

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -331,6 +333,66 @@ private void getStyleFromFloat(
     styles.store(name, String.valueOf(value), isDefaultValue(value));
   }
 
+  private void getStyleFromObject(
```
rickbrew: Since this is grabbing multiple styles, perhaps we should name it `getStylesFromObject`

rickbrew:
Looks good to me :+1:  squash and then I'll merge!

ichub:
Closes https://github.com/facebook/stetho/issues/278

PR #286, opened 2015-09-24
Js tweaks

potyl:
A few tweaks to stetho-js-rhino.

 - fix stacktrace in android m
 - fix console.log("%d", 2); // in javascript numbers are always floats

### DIFF
--- a/stetho-js-rhino/src/test/java/com/facebook/stetho/rhino/JsFormatTest.java
+++ b/stetho-js-rhino/src/test/java/com/facebook/stetho/rhino/JsFormatTest.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.rhino;
+
+import android.os.Build;
+import org.junit.Test;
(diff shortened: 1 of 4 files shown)


stetho-js-rhino/src/test/java/com/facebook/stetho/rhino/JsFormatTest.java
```diff
@@ -0,0 +1,54 @@
...
+import android.os.Build;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.robolectric.RobolectricTestRunner;
+import org.robolectric.annotation.Config;
+
+import static org.junit.Assert.*;
+
+@Config(emulateSdk = Build.VERSION_CODES.JELLY_BEAN)
+@RunWith(RobolectricTestRunner.class)
```
jasta: Do you really need this test to use Robolectric?  It seems quite like this should be able to survive with just the standard JRE.

jasta:
lgtm, feel free to cleanup/squash as you see fit and I'll merge.

stetho-js-rhino/src/test/java/com/facebook/stetho/rhino/JsFormatTest.java
potyl: Fixing it. I just copied the existing unit tests steth-okhttp and built on top.

potyl:
Unit tests running with junit only and squashed.

jasta:
Thanks as always :)

potyl:
I just noticed that in master roboelectric is still used!

PR #288, opened 2015-09-28
Remove dependency on Apache HttpClient

### LINKED ISSUE #251: Support target SDK version of android-MNC for android m release
What's the timeline for an Android M compatible release?

[Apache HTTP client is going away](https://developer.android.com/preview/behavior-changes.html#behavior-apache-http-client) and I see various classes (DumpappHandler.java, ChromeDiscoveryHandler.java, etc.) using it.

jasta:
Android 6.0 removes support for HttpClient and forces applications to
bundle the jars directly if targeting 6.0 in an effort to phase out its
usage more aggressively.  Unfortunately, Stetho depended on HttpClient
for a crude HTTP server implementation to facilitate dumpapp and the
Chrome DevTools protocol.

This diff adds an HTTP server for general purpose usage within Stetho
but also heavily modifies dumpapp to no longer use HTTP at all.  This
became possible late in dumpapp's development when a framing protocol
was developed to support stdout/stderr.  This framing protocol made it
easy to use pure sockets and to further mimic the command-line paradigm.

Unfortunately this diff now starts two Stetho servers, one for dumpapp
and one for devtools, which noticably slowed down Stetho.initialize
performance, though that can be addressed in the future by using a
simple "knocking" mechanism whereby the first n bytes of the connection
are sniffed looking for whether its speaking HTTP or the dumpapp
protocol.  While hacky, this does have some other nice benefits beyond
performance like improved forward compatibility (if newer dumpapp
targets an older Stetho or vice versa).

Closes #251


scripts/dumpapp
```diff
@@ -24,28 +22,17 @@ def main():
...
-      http_method = 'POST'
-      body = sys.stdin.detach().read()
-    else:
-      http_method = 'GET'
-      body = None
-
-    conn.request(http_method, url, body, fake_host)
-
-    reply = conn.getresponse()
+    enter_frame = b'!' + struct.pack('!L', len(args))
```
rightparen: Ideas to consider here:
1) Make the magic more than one character.
2) Include a version number to make reporting version mismatches simpler in the future.
3) As discussed, this might need to look sufficiently like http that if this code connects to an old Stetho server we can deterministically detect and report the version mismatch error.

scripts/dumpapp
```diff
@@ -24,28 +22,17 @@ def main():
...
+    for arg in args:
+      enter_frame += struct.pack(
+          '!H' + str(len(arg)) + 's',
+          len(arg),
+          arg.encode('utf-8'))
```
rightparen: We should have a non-trivial utf-8 test if we don't already.  It seems surprising here that we would send the length of the pre-encoded string rather than the number of utf-8 bytes.

scripts/dumpapp
```diff
@@ -53,19 +40,26 @@ def main():
...
     if code == b'1':
       if n > 0:
-        sys.stdout.buffer.write(read_input(conn, n, 'stdout blob'))
+        sys.stdout.buffer.write(read_input(sock, n, 'stdout blob'))
     elif code == b'2':
       if n > 0:
-        sys.stderr.buffer.write(read_input(conn, n, 'stderr blob'))
+        sys.stderr.buffer.write(read_input(sock, n, 'stderr blob'))
         sys.stderr.buffer.flush()
+    elif code == b'_':
```
rightparen: Consider forwarding stdin unconditionally.  What you have will be sufficient for most plugins, but now that we have a full-duplex transport, it seems unnecessarily restrictive.

scripts/dumpapp
```diff
@@ -53,19 +40,26 @@ def main():
...
     # All frames have a single character code followed by a big-endian int
-    code = read_input(conn, 1, 'code')
-    n = struct.unpack('!L', read_input(conn, 4, 'int4'))[0]
+    code = read_input(sock, 1, 'code')
+    n = struct.unpack('!L', read_input(sock, 4, 'int4'))[0]
 
     if code == b'1':
       if n > 0:
-        sys.stdout.buffer.write(read_input(conn, n, 'stdout blob'))
+        sys.stdout.buffer.write(read_input(sock, n, 'stdout blob'))
```
rightparen: At one point we discussed flushing on every frame so that flush in the plugin implicitly causes flush on the host.  Is that still the plan and is that a separate diff?

stetho/src/main/java/com/facebook/stetho/server/LocalSocketServer.java
```diff
@@ -0,0 +1,175 @@
...
+            WORKER_THREAD_NAME_PREFIX +
+            "-" + mFriendlyName +
+            "-" + mThreadId.incrementAndGet());
+        t.setDaemon(true);
+        t.start();
+      } catch (SocketException se) {
+        // ignore exception if interrupting the thread
+        if (!Thread.interrupted()) {
+          LogUtil.w(se, "I/O error");
+        }
```
rightparen: Thread.interrupted resets its state (http://docs.oracle.com/javase/7/docs/api/java/lang/Thread.html#interrupted%28%29), so this should include 'else { break; }'.

stetho/src/main/java/com/facebook/stetho/server/http/HandlerRegistry.java
```diff
@@ -0,0 +1,39 @@
...
+public class HandlerRegistry {
+  private final ArrayList<PathMatcher> mPathMatchers = new ArrayList<>();
+  private final ArrayList<HttpHandler> mHttpHandlers = new ArrayList<>();
+
+  public synchronized void register(PathMatcher path, HttpHandler handler) {
+    mPathMatchers.add(path);
+    mHttpHandlers.add(handler);
+  }
+
+  public synchronized void unregister(PathMatcher path, HttpHandler handler) {
```
rightparen: Prefer taking just a PathMatcher, finding it by index, and removing the entries at that index from both ArrayLists.

As it is, if you had something like:
path0: handler0
path1: handler1
path2: handler0

and you called:
unregister(path2, handler0)

you would end up with:
path0: handler1
path1: handler0

stetho/src/main/java/com/facebook/stetho/server/http/LeakySocket.java
```diff
@@ -0,0 +1,43 @@
...
+  LeakySocket(LocalSocket socket, LeakyBufferedInputStream leakyInput) {
+    mSocket = socket;
+    mLeakyInput = leakyInput;
+  }
+
+  public InputStream getInput() throws IOException {
+    return new CompositeInputStream(new InputStream[] {
+        new ByteArrayInputStream(mLeakyInput.clearBuffer()),
+        mSocket.getInputStream()
+    });
```
rightparen: Two concerns here:
1) Do we need the clearBuffer() machinery?  It looks like the contents remaining in the stream we return via this machinery are exactly the same as if we just returned mLeakyInput here.
2) The remainder of the stream is unbuffered.  Is that intentional?

scripts/dumpapp
jasta: Done 1 and 2.  Not yet sure how to handle 3.

scripts/dumpapp
jasta: Oops, yeah this is a mistake.  Fixed by encoding ahead of time and treating the length as the byte count, not char count.  I'm not used to working in Python :)

scripts/dumpapp
jasta: We can't directly "connect" stdin from the terminal to the app because it means we will perform a read that wasn't requested.  This manifests itself in strange ways for non-trivial scripts such as:

```
$ echo -en "123" | (head -c 1; tail -c 1); echo
13
```

Both the head and tail invocation are given the same fd (/dev/stdin is fixed at this point) and both open the fd, read what they are told to read, then exit.  In this way, I can instruct head to read just 1 byte, leaving the remaining two on the stream for tail to skip one and print the last.  If head was coded as you suggest dumpapp to be coded, then the fd will be exhausted immediately and forwarded to head.  Tail would then print nothing.

This framing protocol design allows us instead to "connect" the fds perfectly, transferring each individual read call from the dumper plugin to a read call on the dumpapp script's stdin as it would work in a pipe situation normally.

scripts/dumpapp
jasta: I'd prefer to tackle that in a separate diff, but yes I did still intend to have it work that way.

stetho/src/main/java/com/facebook/stetho/server/LocalSocketServer.java
jasta: Good catch, this bug existed in the old code too.

stetho/src/main/java/com/facebook/stetho/server/http/HandlerRegistry.java
jasta: I actually wanted the behaviour you described.  They register as a pair and should unregister as a pair.  This makes register/unregister balanced.

stetho/src/main/java/com/facebook/stetho/server/http/LeakySocket.java
jasta: 1) We don't explicitly need clearBuffer (it could just return the buffer and not clear it) but I designed it this way to be explicit that you are taking over ownership of the original, unbuffered stream (see [2])
2) Yes, I wanted to make it possible to transition from a buffered stream to an unbuffered one as this is used to transfer protocol to a text-based one (HTTP/1.x) to a binary one (WebSocket).  The WebSocket approach wants to do its own buffering, if it buffers at all.  It should therefore be given the original, unbuffered socket.

I did rework this a bit as I re-used the code for the protocol sniffing socket handler in the latest round of changes.  Specifically, this now encapsulates the `getInput()` functionality in `LeakyBufferedInputStream` with simply: `leakAndClearInput()` which returns a composite of the buffer and the unbuffered stream.

scripts/dumpapp
jasta: With this new architecture it was actually fairly trivial to support the old dumpapp script unmodified.  I added an extra crude HTTP handler that exclusively buffers responses and prints a warning:
$ ./scripts/dumpapp 

```
*** WARNING: Using legacy HTTP protocol; update dumpapp script! ***
Usage: dumpapp [options] <plugin> [plugin-options]
 -h,--help            Print this help
 -l,--list            List available plugins
 -p,--process <arg>   Specify target process
```

The warning is on stderr, so piping the above output to null causes:

```
$ ./scripts/dumpapp -h > /dev/null
*** WARNING: Using legacy HTTP protocol; update dumpapp script! ***
```

jasta:
Rebased and squashed.  @rightparen I'm assuming based on our out-of-band conversation you're good with this so landing as soon as Travis says it's fine.

PR #292, opened 2015-09-30
Json serialization fixes

potyl:
This branch fixes serialization issues that cause stetho to die when we try to send back NaN, Infinity or -Infinity. These values are not part of the JSON specification. What this patch does is to serialize them as String since JavaScript will accept them back as Doubles.

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/json/ObjectMapper.java
+++ b/stetho/src/main/java/com/facebook/stetho/json/ObjectMapper.java
@@ -101,7 +101,7 @@ public class ObjectMapper {
       Object value = jsonObject.opt(field.getName());
       Object setValue = getValueForField(field, value);
       try {
-        field.set(instance, getValueForField(field, value));
+        field.set(instance, setValue);
       } catch (IllegalArgumentException e) {
         throw new IllegalArgumentException(
             "Class: " + type.getSimpleName() + " " +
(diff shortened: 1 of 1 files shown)


stetho/src/main/java/com/facebook/stetho/json/ObjectMapper.java
```diff
@@ -273,6 +273,18 @@ private Object getJsonValue(Object value, Class<?> clazz, Field field)
       return convertValue(value, JSONObject.class);
     }
     // hmm we should be able to directly serialize here...
```
jasta: this command should probably be moved to after this new `if` statement

stetho/src/main/java/com/facebook/stetho/json/ObjectMapper.java
```diff
@@ -101,7 +101,7 @@
       Object value = jsonObject.opt(field.getName());
       Object setValue = getValueForField(field, value);
       try {
-        field.set(instance, getValueForField(field, value));
+        field.set(instance, setValue);
```
jasta: good catch :)

jasta:
only a minor comment, looks good otherwise

stetho/src/main/java/com/facebook/stetho/json/ObjectMapper.java
potyl: I don't understand this comment. Do you mean "this comment" ???

jasta: yes sorry, I meant this comment.  it should be just before the last line's return statement.

potyl: Yes. I'll change it.

PR #294, opened 2015-10-01
Add support for STETHO_PROCESS environment variable to dumpapp

### LINKED ISSUE #291: Add ability to specify process in an environment variable for dumpapp.
Currently we support `ANDROID_SERIAL` we should also support `STETHO_PROCESS` in dumpapp.

longinoa:
We want to be able to support passing the process via environment
variable. This gets around having to create arbitrary wrapper scripts
that just pass `-p myprocess $@` to dumpapp.

Closes #291 

https://gist.github.com/anonymous/b20a4e34cf774b4abdb6

### DIFF
--- a/scripts/dumpapp
+++ b/scripts/dumpapp
@@ -11,7 +11,11 @@ from stetho_open import *
 def main():
   # Manually parse out -p <process>, all other option handling occurs inside
   # the hosting process.
-  process = None
+
+  # Connect to the process passed in via -p. If that is not supplied fallback
+  # the process defined in STETHO_PROCESS. If neither are defined throw.
+  process = os.environ.get('STETHO_PROCESS')
+
   args = sys.argv[1:]
   if len(args) > 0 and (args[0] == '-p' or args[0] == '--process'):
     if len(args) < 2:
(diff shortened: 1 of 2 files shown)


jasta:
You don't need to say `Closes #294` when this is 294, that's actually pretty confusing :)

jasta:
Also I'm confused about your pasted transcript: Stetho was intended to auto-select the only Stetho-enabled process if there is just 1 but in your transcript it shows that no longer working.  Not sure if your diff broke it or something else did but that should be fixed as a precursor to make sure your diff hasn't affected the intended flow.

scripts/dumpapp
```diff
@@ -11,7 +11,11 @@ from stetho_open import *
 def main():
   # Manually parse out -p <process>, all other option handling occurs inside
   # the hosting process.
-  process = None
+
+  # Connect to the process passed in via -p.If that is not supplied fallback
```
jasta: nit: space between `.` and first character

longinoa:
Whoops ... auto complete hurt me there it should close #291 not #294 :)

jasta:
lgtm, merge at will

PR #296, opened 2015-10-06
Support for Dialogs without a DialogFragment

### LINKED ISSUE #295: Adding Dialogs to descriptors (not DialogFragment)
As far as I can tell, Stetho supports displaying `DialogFragments` in the Elements tab, but it doesn't support plain old dialogs, which I use.

I need to be able to add more Windows (the dialogs ones) to what Stetho is collecting. I thought you'd have implemented this for ChatHead already, so I might have missed something, let me know if that's the case.

In the meantime... Here's my dirty hack:

```
public class MyActivity extends Activity {

  @Override public Window getWindow() {
    if (!BuildConfig.DEBUG || !switchWindowToDialogForStetho) {
      return super.getWindow();
    }
    Window dialogWindow = mainPathContainer.getDialogWindowForDebug();

    if (dialogWindow == null) {
      return super.getWindow();
    }

    StackTraceElement[] stackTrace = new RuntimeException().getStackTrace();

    if (stackTrace[1].getClassName()
        .equals("com.facebook.stetho.inspector.elements.android.ActivityDescriptor")) {
      return dialogWindow;
    }

    return super.getWindow();
  }
}
```

rickbrew:
This is preliminary support for `Dialogs` that are not hosted by a `DialogFragment`.

I haven't had really had time to test this yet, but it should work.

Closes #295

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ActivityDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ActivityDescriptor.java
@@ -10,10 +10,13 @@
 package com.facebook.stetho.inspector.elements.android;
 
 import android.app.Activity;
+import android.app.Dialog;
+import android.util.SparseArray;
 import android.view.View;
 import android.view.Window;
 
 import com.facebook.stetho.common.Accumulator;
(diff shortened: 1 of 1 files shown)


rickbrew:
Superseded by #325

PR #297, opened 2015-10-07
Sort the View CSS properties by cssName

felipecsl:
Makes it much easier to read

Before:
screen shot 2015-10-08 at 9 52 11 pm

After:
screen shot 2015-10-08 at 9 51 13 pm

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
@@ -30,8 +30,6 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
-import java.util.TreeSet;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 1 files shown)


stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -63,7 +66,12 @@ private Pattern getWordBoundaryPattern() {
     if (mViewProperties == null) {
       synchronized (this) {
         if (mViewProperties == null) {
-          List<ViewCSSProperty> props = new ArrayList<>();
+          Set<ViewCSSProperty> props = new TreeSet<>(new Comparator<ViewCSSProperty>() {
```
rickbrew: What about keeping it as a `List` and then sorting afterward? We're intentionally avoiding things like `Map` and `Set` because they add an extra allocation per entry. This code also adds an extra copy operation (`new ArrayList<>(props)') that we'd prefer to avoid.

rickbrew:
Screenshot of before and after?

felipecsl:
PR updated, screenshots attached

jasta:
lgtm, thanks!

PR #298, opened 2015-10-07
Support for CSS.setPropertyText

felipecsl:
This is an initial stab at adding support for editing View properties via the CSS tab in the Elements tab.
There are a few problems/drawbacks here:
- I'm using Chrome 45 which uses `CSS.setPropertyText`. However, that doesn't seem to be in the [protocol specification](https://code.google.com/p/chromium/codesearch#chromium/src/third_party/WebKit/Source/devtools/protocol.json&q=protocol.json&sq=package:chromium&type=cs) anymore. In fact, `setPropertyText` has been migrated to `setStyleText` [in June](https://chromium.googlesource.com/chromium/src/+/43034046ffeeb2316ba26d00fde97b4202cf404e) but apparently hasn't shipped to Chrome Stable yet.
- Right now this is only able to set simple properties (egg.: x, y, scaleX, scaleY) that have a corresponding setter method (setX, setY, setScaleX, setScaleY, etc) using `MethodInvoker`. We probably need to be able to support setting other stuff like padding and margins, however that may require a more customized approach. I see that some similar logic has been implemented for `getStyles()` so @ichub might have ideas on how we could achieve this.

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/common/Ascii.java
+++ b/stetho/src/main/java/com/facebook/stetho/common/Ascii.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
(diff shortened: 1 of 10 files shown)


jasta:
Too ancient after a number of large refactors in this area.  Please re-submit if this is a priority!

PR #299, opened 2015-10-14
Fixed a couple of memory leaks.

jferlisi:
1. The map in ViewGroupDescriptor now stores WeakReference<Object> as a value instead of Object directly. Since views were stored there directly that could cause activities to be leaked through the context of the view.
2. The highlight view field is now nulled out when the highlighting is cleared.

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewGroupDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewGroupDescriptor.java
@@ -16,11 +16,13 @@ import com.facebook.stetho.common.Accumulator;
 import com.facebook.stetho.common.android.FragmentCompatUtil;
 import com.facebook.stetho.inspector.elements.AbstractChainedDescriptor;
 
+import java.lang.ref.WeakReference;
 import java.util.Collections;
 import java.util.Map;
 import java.util.WeakHashMap;
 
 final class ViewGroupDescriptor extends AbstractChainedDescriptor<ViewGroup> {
+
   /**
    * This is a cache that maps from a View to the Fragment that contains it. If the View isn't
    * contained by a Fragment, then this maps the View to itself. For Views contained by Fragments,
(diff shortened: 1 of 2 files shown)


stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewGroupDescriptor.java
```diff
@@ -27,8 +28,8 @@
    * we emit the Fragment instead, and then let the Fragment's descriptor emit the View as its sole
    * child. This allows us to see Fragments in the inspector as part of the UI tree.
    */
-  private final Map<View, Object> mViewToElementMap =
-      Collections.synchronizedMap(new WeakHashMap<View, Object>());
+  private final Map<View, WeakReference<Object>> mViewToElementMap =
+      Collections.synchronizedMap(new WeakHashMap<View, WeakReference<Object>>());
```
rickbrew: Using a `WeakReference` for the value will allocate an extra object for every single `View` that's present in the view hierarchy. Instead of always using a WR for the value, why not use a sentinel object to indicate "the value is the same as the key" (which is always the case when the value is a `View`). And then only use a WR for fragments (of which there are very few). This will save the allocations from always using a WR, while also avoiding the memory leak you discovered.

```
private final Object mViewSentinelValue = new Object();
...
Object elementReference = mViewToElementMap.get(childView);
if (elementReference == mViewSentinelValue) {
  return childView;
} else if (elementReference instanceof WeakReference) { 
  ...
}
...
mViewToElementMap.put(childView, mViewSentinalValue);
```

jferlisi: Good call. I made changes to do this.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewGroupDescriptor.java
```diff
@@ -16,11 +16,19 @@
...
 import java.util.Map;
 import java.util.WeakHashMap;
 
 final class ViewGroupDescriptor extends AbstractChainedDescriptor<ViewGroup> {
+
+  /**
+   * This is a marker object for values in the {@link #mViewToElementMap}. When this is present as a
+   * value the key should be used as the value.
+   */
+  private static final Object SENTINEL = new Object();
```
rickbrew: let's make this non-static. that way it can be GC'd once Chrome disconnects. (descriptors are effectively singletons anyway)

```
private final Object mSentinal = new Object();
```

Heck we could even save the allocation and initialize mSentinel to equal `this` in the constructor :) ...

jferlisi: I'll just remove the sentinel object and use `this`.

rickbrew:
Looks good to me! :) Squash the commits and then I'll merge it. Thanks for noticing and fixing this!

jferlisi:
The commits are squashed. Sorry it took a couple of days to get to it.

rickbrew:
Thanks :)

PR #302, opened 2015-10-23
Display logs with ASSERT level as ERROR messages in Chrome console.

colriot:
`ASSERT` log level often means **never_should_happen** type of errors, so let them look like errors in Chrome console.


jasta:
I had never even heard of `Log.ASSERT`, neat.

PR #305, opened 2015-10-30
Bundle default proguard rules

frett:
v0.5.7 of the android gradle build plugin added support for bundling proguard rules in an aar using `consumerProguardFiles`

This adds base proguard rules to the stetho aar and stetho-js-rhino aar. This will allow usage of these modules in a proguarded project without the developer needing to hunt down and configure proguard for stetho.

### DIFF
--- a/stetho-js-rhino/build.gradle
+++ b/stetho-js-rhino/build.gradle
@@ -9,6 +9,7 @@ android {
         targetSdkVersion 23
         versionCode 1
         versionName "1.0"
+        consumerProguardFiles 'proguard-consumer.pro'
     }
 
     lintOptions {
--- a/stetho-js-rhino/proguard-consumer.pro
+++ b/stetho-js-rhino/proguard-consumer.pro
@@ -1,0 +1,22 @@
+-keep class com.facebook.stetho.rhino.** { *; }
+
+# rhino (javascript)
+-dontwarn org.mozilla.javascript.**
+-dontwarn org.mozilla.classfile.**
+-keep class org.mozilla.classfile.** { *; }
+-keep class org.mozilla.javascript.* { *; }
+-keep class org.mozilla.javascript.annotations.** { *; }
+-keep class org.mozilla.javascript.ast.** { *; }
+-keep class org.mozilla.javascript.commonjs.module.** { *; }
+-keep class org.mozilla.javascript.commonjs.module.provider.** { *; }
+-keep class org.mozilla.javascript.debug.** { *; }
+-keep class org.mozilla.javascript.jdk13.** { *; }
+-keep class org.mozilla.javascript.jdk15.** { *; }
+-keep class org.mozilla.javascript.json.** { *; }
+-keep class org.mozilla.javascript.optimizer.** { *; }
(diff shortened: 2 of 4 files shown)


felipecsl:
This is nice, but ideally stetho should provide a no-op module (like LeakCanary does, for example), that could be added as a release dependency, so you dont need to worry about including it in your production builds:

```
debugCompile 'com.facebook.stetho:stetho:1.1.0'
releaseCompile 'com.facebook.stetho:stetho-no-op:1.1.0'
```

If the maintainers agree with this, I can prepare a PR quickly with this change

Anyways +1 for this

frett:
I'm actually already only including stetho on debug builds, but for reasons related to method counts, multidex, and testing of debug builds on a Gingerbread device I had to enable proguard on my debug build.

This led me to having to hunt down the proguard rules so I could get stetho to work in a proguarded debug build :)

felipecsl:
yeah we have very similar setups then :)

jasta:
lgtm, thanks for adding this!

jasta:
@felipecsl I do agree with the no-op module but unfortunately it's not that simple for Stetho since the public surface area is quite large.  The no-op trick relies on being able to ship a simple dummy stub that is disabled internally but that provides the full public API surface area of the library in question (actually not unlike Android's own android.jar that ships in the SDK).

To maintain this manually for Stetho would be quite a lot of work considering we have been changing the public initialization API pretty steadily in almost every release, and are expecting to continue as things mature.  We're strictly backwards compatible, of course, but it still means that with a no-op module you'd have to add all the bells and whistles with each release to stub the following example init:

```java
Stetho.initialize(
  Stetho.newInitializerBuilder(context)
    .enableDumpapp(
      new DefaultDumperPluginsBuilder(context)
        .provide(new MyPlugin(...))
        .finish())
    .build()));
```

That's quite a lot of classes and such that would be needed to be stubbed.  

Note that we currently actually do have this trick enabled for the stetho-urlconnection module.  It is designed to be included and directly invoked regardless of whether the core stetho module is present at runtime.  The difficulty of this trick without tools increases significantly for the stetho core module for the API surface area reason above though.

One possible strategy for this would be to divide stetho into stetho-common so that stetho-noop could share some of that "real" code, though unfortunately that's a lame compromise because in reality all of the methods could just be empty (for instance, new DefaultDumperPluginsBuilder(...).provide(...) really doesn't need to maintain an ArrayList data structure at all).

ashughes:
It's worth noting here that you must include `@aar` at the end of your gradle compile statement, otherwise you just get the jar and no Proguard rules are included. For example:

    debugCompile 'com.facebook.stetho:stetho:1.3.1@aar'

However, I tried building with the `1.3.1` AAR and it crashed (see #419).

ashughes:
As was mentioned on #419, in order to target the AAR you must set `transitive = true`:

    debugCompile('com.facebook.stetho:stetho:1.3.1@aar') {
        transitive = true
    }

Ideally, the default artifact should be the AAR, then we'd automatically get the Proguard rules and not worry about transitive dependencies.

PR #317, opened 2015-12-14
Update Gradle wrapper and Android plugin

friederbluemle:
- Gradle wrapper `2.9`
- Android Gradle plug-in `1.5.0`


jasta:
Travis checks are failing.  Can you take a look?

friederbluemle:
@jasta I did some investigation and it turned out the build failure was caused by a Lint error on Travis: `The SDK platform-tools version ((21)) is too old  to check APIs compiled with API 23; please update`.
It seems the Travis build server is using an old version, and this check was added with the updated Android plug-in. Simply adding `platform-tools` to the components list in .travis.yml fixed the error. The build should pass now.

jasta:
Thanks!

Issue #210, opened 2015-06-26
AndroidDOMProvider is overzealous with updates


rickbrew:
`DOMProvider` is obligated to tell its `Listener` that the tree changed by calling the `DOMProvider.Listener.onPossiblyChanged()` method. `DOM` then picks this up and re-reads the entire tree and does some smart stuff and pushes the delta over to Chrome.

However, `AndroidDOMProvider` calls `onPossiblyChanged()` once per second. Every second. Even if nothing could possibly have changed! If we can hook into something like `ViewTreeObserver` correctly then we should be able to have a better idea of when we should call actually call this method.

The goal: if the app is idle, then `AndroidDOMProvider` should not call `onPossiblyChanged()`.

jasta:
I'm taking a look at this :)

PR #323, opened 2016-01-02
Add ability to toggle view inspection

bgogetap:
This is a quick fix to add the option to disable view inspection. It's a great addition, but can cause significant slowdown for applications with deep hierarchies.

### DIFF
--- a/stetho-sample/src/debug/java/com/facebook/stetho/sample/SampleDebugApplication.java
+++ b/stetho-sample/src/debug/java/com/facebook/stetho/sample/SampleDebugApplication.java
@@ -50,6 +50,7 @@ public class SampleDebugApplication extends SampleApplication {
           }
         })
         .enableWebKitInspector(new ExtInspectorModulesProvider(context))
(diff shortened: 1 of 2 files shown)


jasta:
While I agree this is sorely needed, the slowdown caused by the view inspection code is actually more of a bug than its intrinsic performance cost.  The solution we use recursively scans the view hierarchy _every second_ regardless of whether the user is actively looking at the Elements tab or even changing the UI.  I've been nagging @rickbrew a bit to address this and I think we have good options we just haven't gotten to them yet.

Specifically, I believe it should be possible to utilize http://developer.android.com/reference/android/view/View.html#getViewTreeObserver() to detect changes such as layout, measure, draw events and then update accordingly.  Maybe I can take a crack at this tonight :)

jasta:
This is being tracked in #210

jasta:
I believe we have fixed the major performance issues in the latest master (expecting a 1.3.1 release soon).  Please test and re-open the issue if I'm wrong :)

PR #324, opened 2016-01-03
Add OkHttp 3 integration, Deprecate OkHttp 2 integration.

swankjesse:


jasta:
Why is this labeled as an RC if it's ready for this level of attention in Stetho?  Also, this diff would make it so that 2.x would no longer function.  We are forced to do a similar artifact change to offer stetho-okhttp and stetho-okhttp3 (which is fine, it's just not reflected in this diff).

FWIW, I've looked over https://github.com/square/okhttp/blob/master/CHANGELOG.md and I like what I see in general.

swankjesse:
I’m planning on releasing 3.0 final this week or early next week. Would you like me to update this PR to include separate okhttp and okhttp3 artifacts?

jasta:
@swankjesse ack'd re the 3.0 final, but regarding separate artifacts I worry that maintaining these may be difficult for us moving forward if we simply copy/paste them but at the same time it would break version continuity if we, say, wanted to fix a bug in an older version without requiring users upgrade through to okhttp 3.x.

Given that this code has not really been modified in quite some time, and is generally a trivial integration component, I think we can probably tolerate just forking stetho-okhttp as stetho-okhttp3 in our own tree and then deprecating stetho-okhttp.  If this was under active development we would likely need a better solution somehow.

swankjesse:
Yeah, OkHttp 3 is the only OkHttp going forward. I’m hoping to migrate the entire world to OkHttp 3 very quickly, hence this PR. What can I do to unblock that work?

b95505017:
Time to switch to `compile 'com.squareup.okhttp3:okhttp:3.0.1'` and publish a new release?

jakewilson801:
Please I'm about to build my own version of this library ;)

swankjesse:
I’ve updated the PR description and uploaded a new change. The new change leaves the existing 2.x integration in place and adds a new `okhttp3` submodule that contains the new integration. This should make it simple to offer OkHttp 3 to everyone who wants it, without breaking anyone who hasn’t upgraded yet.

swankjesse:
(The OkHttp 3 submodule is an almost perfect copy-paste of the OkHttp 2 submodule.)

b95505017:
:+1:  Please merge & publish this ASAP.

panzerdev:
:+1: Please make it happen so I can update to OkHttp 3.0.1 right away..

stetho-okhttp/src/main/java/com/facebook/stetho/okhttp/StethoInterceptor.java
```diff
@@ -13,7 +13,13 @@
 import com.facebook.stetho.inspector.network.NetworkEventReporter;
 import com.facebook.stetho.inspector.network.NetworkEventReporterImpl;
 import com.facebook.stetho.inspector.network.RequestBodyHelper;
-import com.squareup.okhttp.*;
+import com.squareup.okhttp.Connection;
```
jasta: Why did this change?

jasta:
Thanks for doing this.  lgtm.  I'll be cutting a new Stetho release tomorrow with this included.

jasta:
In the mean time, if you are aching to get this integrated you can actually copy the StethoInterceptor.java found in this diff into your project and use it directly.  It's discrete and should stop the bleeding while you wait for us to cut a release.

stetho-okhttp/src/main/java/com/facebook/stetho/okhttp/StethoInterceptor.java
swankjesse: I just replaced wildcard imports with full imports. Behavior is the same!

swankjesse:
Thanks @jasta!

lourenci:
How can I use it? Do I need a new gradle release? 🤔

b95505017:
@lourenci yes

stetho-okhttp/src/main/java/com/facebook/stetho/okhttp/StethoInterceptor.java
jasta: Ack'd, obviously we should edit our IntelliJ settings too ;)

androideveloper:
When will this be released?

jasta:
It's being released right now.  We had an issue last night with Travis CI that prevented the release.

jasta:
Moving through the tubes, should be available as soon as replication through Maven Central and jcenter is complete.

b95505017:
Will https://github.com/ParsePlatform/ParseInterceptors-Android need upgrade too?

panzerdev:
@jasta Thanks! Great to see things move so fast. OkHttp and Stetho are just basic for every project :+1:

jasta:
@b95505017 no it shouldn't, it doesn't depend on okhttp.  This is an okhttp issue only.

PR #325, opened 2016-01-06
Detect all Android windows

### LINKED ISSUE #295: Adding Dialogs to descriptors (not DialogFragment)

pyricau:
- Adds support for dialogs (child windows of an activity) as well as any window manually added to the window manager.
- Added a dialog and a fake chathead to the sample

Fixes #295

**DO NOT MERGE**, this is a WIP: works alongside the existing mechanism, so reporting is duplicated.
## More context

The way Stetho currently works is incorrect: it assumes that there is only ever one window per activity, and that activities are the only source of windows. It recently added support for dialog fragments, but dialog fragments are just (bad) wrappers around real dialogs, and Stetho currently does not support normal dialogs. At the end of the day, a dialog is just a fancy window.

An application can spawn any number of windows, regardless of activities running.

I'm opening this PR to discuss the way forward, see what we can do to reunite the two approaches.

screenshot-2016-01-06_13 47 08 388_framed

screen shot 2016-01-06 at 1 48 06 pm


rickbrew:
Oh nice! I'll start looking at this ...

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/RootsOracle.java
```diff
@@ -0,0 +1,89 @@
...
+  private static final String GET_GLOBAL_INSTANCE = "getInstance";
+
+  private boolean initialized;
+  private Object windowManager;
+  private Field viewsField;
+
+  /**
+   * Looks for all Root views
+   */
+  public List<View> findAllRootViews() {
```
rickbrew: Please use the Allocator pattern instead of returning a List, which will help avoid allocations

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDocumentRoot.java
```diff
@@ -38,5 +40,8 @@ protected String onGetNodeName(AndroidDocumentRoot element) {
   @Override
   protected void onGetChildren(AndroidDocumentRoot element, Accumulator<Object> children) {
     children.store(mApplication);
+    for (View view : rootsOracle.findAllRootViews()) {
```
rickbrew: please use the `Allocator` pattern here so we don't have an iterator allocation

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/RootsOracle.java
```diff
@@ -0,0 +1,89 @@
...
+    if (currentThread() != Looper.getMainLooper().getThread()) {
```
rickbrew: Regarding thread affinity, look into implementing the `ThreadBound` interface.

A good way to do this would be to have this class extend from `ThreadBoundProxy`. this now means your constructor needs to take a `ThreadBound` to pass down to the super class. Which means `AndroidDocumentRoot` should take a `ThreadBound` from `AndroidDocumentProvider`, which conveniently has one available from its constructor.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/RootsOracle.java
```diff
@@ -0,0 +1,89 @@
...
+ * core/src/main/java/android/support/test/espresso/base/RootsOracle.java
+ */
+final class RootsOracle {
+  private static final String WINDOW_MANAGER_IMPL_CLASS = "android.view.WindowManagerImpl";
+  private static final String WINDOW_MANAGER_GLOBAL_CLASS = "android.view.WindowManagerGlobal";
+  private static final String VIEWS_FIELD = "mViews";
+  private static final String GET_DEFAULT_IMPL = "getDefault";
+  private static final String GET_GLOBAL_INSTANCE = "getInstance";
+
+  private boolean initialized;
```
rickbrew: please be sure to use the `mFieldName` pattern

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/RootsOracle.java
```diff
@@ -0,0 +1,89 @@
...
+  private Field viewsField;
+
+  /**
+   * Looks for all Root views
+   */
+  public List<View> findAllRootViews() {
+    if (currentThread() != Looper.getMainLooper().getThread()) {
+      throw new UnsupportedOperationException("Expecting main thread, not " + currentThread());
+    }
+    if (!initialized) {
```
rickbrew: I'd prefer to have this check done in the `initialize` method. I tend to call the method something like `ensureInitialized` to help document that.

stetho-sample/src/main/java/com/facebook/stetho/sample/ChatHeadService.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  public static void stop(Context context) {
+    context.stopService(new Intent(context, ChatHeadService.class));
+  }
+
+  private WindowManager windowManager;
+  private ImageView chatHead;
+
+  @Override public IBinder onBind(Intent intent) {
```
rickbrew: newline after `@Override`

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/RootsOracle.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    try {
+      Class<?> clazz = Class.forName(accessClass);
+      Method getMethod = clazz.getMethod(instanceMethod);
+      windowManager = getMethod.invoke(null);
+      viewsField = clazz.getDeclaredField(VIEWS_FIELD);
+      viewsField.setAccessible(true);
+    } catch (Exception ignored) {
+    }
```
rickbrew: if we hit this exception then we should log a warning via `com.facebook.stetho.common.LogUtil` to help developers debug why they don't see dialogs

rickbrew:
Okay there's my first round of code review comments on the stetho side. I haven't looked at much of the stetho-sample additions yet.

pyricau:
Thanks for the feedback, I'll try to fix the styling issues tomorrow.

jasta:
Where are we at with this diff?

pyricau:
Haven't found time to work on this.

jasta:
Closing to clean up our queue

SR1s:
so... still not support dialog?

PR #345, opened 2016-01-22
Fix for duplicates in Element tree, and fix JSON unit test

### LINKED ISSUE #339: Duplication in the Elements tree
Looks like this actually broke back in #237 .

When we transmit DOM changes to Chrome, we need to keep track of all the elements for each transmission so we don't send over duplicates. When we send over `DOM.childNodeInserted(E)`, which includes E and its sub-tree "by value", we must not send over `DOM.childNodeInserted()` for any children of E, otherwise they'll be duplicated.

### LINKED ISSUE #347: ObjectMapperTest unit test fails on Java 8
Apparnetly the order of serialization changed in Java8, so this test is now failing.

rickbrew:
Looks like this actually broke back in #237. We missed 1 tiny detail in the middle of a very large refactoring job.

When we transmit DOM changes to Chrome, we need to keep track of all the elements for each transmission so we don't send over duplicates. When we send over `DOM.childNodeInserted(E)`, which includes E and its whole sub-tree "by value", we must not send a subsequent `DOM.childNodeInserted(*)` for any elements within E's sub-tree, otherwise they'll show up as duplicates in the DevTools UI.

This was managed with a `HashSet` indirectly populated by `DOM.createNodeForElement()` and checked by `ShadowDocument.applyDocumentUpdate()`. The code is still there in the latter, but it got lost in the former. It was only storing the root node of any transmitted sub-tree into the `HashSet` (indirectly via the `Accumulator` interface).

Also, the ObjectMapperTest has been failing due to Java 8 changing the order of things. That commit is also part of thie PR.

Closes #339 
Closes #347

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
@@ -32,6 +32,8 @@ import com.facebook.stetho.json.ObjectMapper;
 import com.facebook.stetho.json.annotation.JsonProperty;
 import org.json.JSONObject;
 
+import javax.annotation.Nullable;
+
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -79,7 +81,7 @@ public class DOM implements ChromeDevtoolsDomain {
       @Override
       public Node call() {
         Object element = mDocument.getRootElement();
-        return createNodeForElement(element, mDocument.getDocumentView());
+        return createNodeForElement(element, mDocument.getDocumentView(), null);
       }
     });
 
@@ -253,7 +255,14 @@ public class DOM implements ChromeDevtoolsDomain {
     }
   }
 
-  private Node createNodeForElement(Object element, DocumentView view) {
+  private Node createNodeForElement(
+      Object element,
+      DocumentView view,
+      @Nullable Accumulator<Object> processedElements) {
+    if (processedElements != null) {
+      processedElements.store(element);
+    }
+
     NodeDescriptor descriptor = mDocument.getNodeDescriptor(element);
 
     Node node = new DOM.Node();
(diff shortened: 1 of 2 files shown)


stetho/src/test/java/com/facebook/stetho/json/ObjectMapperTest.java
```diff
@@ -69,14 +70,23 @@ public void testNestedProperty() throws JSONException {
     njp.child1.testString = "testString";
     njp.child2.i = 4;
 
-    String expected = "{\"child1\":{\"testString\":\"testString\"},\"child2\":{\"i\":4}}";
-    NestedJsonProperty parsed = mObjectMapper.convertValue(
-        new JSONObject(expected),
+    // The ordering of serialization changes depending on Java 7 vs Java 8.
+    String expected7 = "{\"child1\":{\"testString\":\"testString\"},\"child2\":{\"i\":4}}";
+    String expected8 = "{\"child2\":{\"i\":4},\"child1\":{\"testString\":\"testString\"}}";
```
jasta: Why wouldn't we just do an order-independent comparison?  The order of fields is not important in JSON.  It would seem to me like linking in something like https://github.com/skyscreamer/JSONassert would be the right way to go.  This can be done easily by modifying stetho/build.gradle and adding testCompile 'org.skyscreamer:jsonassert:1.3.0' and then just using the code.  This also looks nice: https://github.com/hertzsprung/hamcrest-json.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/DOM.java
```diff
@@ -79,7 +81,7 @@ public JsonRpcResult getDocument(JsonRpcPeer peer, JSONObject params) {
       @Override
       public Node call() {
         Object element = mDocument.getRootElement();
-        return createNodeForElement(element, mDocument.getDocumentView());
+        return createNodeForElement(element, mDocument.getDocumentView(), null);
```
jasta: style nit: null /* argName */

jasta:
lgtm, though I would like to see the JSON unit test improved a bit.

PR #348, opened 2016-02-02
Cache the result from ObjectMapper.getJsonValueMethod()

rickbrew:
Caching this results in significant performance wins for the Elements tab, and probably for any other JSON serialization we're doing that uses enums with `@JsonValue `.

On a Moto E, I saw `Document.updateTree()` times drop from ~600ms to ~150ms in a scrolling ListView container.

I also experimented with a general reflection cache, e.g. Class<?>.getFields(), getEnumConstants(), etc. While that also seemed to improve things, it wasn't enough to inspire confidence just yet. I'll want to employ a micro-benchmark and/or sampling profiler before heading in that direction further.


jasta:
lgtm, thanks so much for doing this investigation

PR #349, opened 2016-02-04
A couple of script fixes

mottis:
- Reading from socket in stetho_open was not correctly maintaining the number of bytes to read
- Reading from socket in dumper script must be done in a loop until the expected number of bytes is received
- Writing to stdout should be followed by a flush, so data is displayed as soon as it's received

### DIFF
--- a/scripts/dumpapp
+++ b/scripts/dumpapp
@@ -58,6 +58,7 @@ def read_frames(sock):
     if code == b'1':
       if n > 0:
         sys.stdout.buffer.write(read_input(sock, n, 'stdout blob'))
+        sys.stdout.buffer.flush()
     elif code == b'2':
       if n > 0:
         sys.stderr.buffer.write(read_input(sock, n, 'stderr blob'))
@@ -75,12 +76,6 @@ def read_frames(sock):
       if raise_on_eof:
         raise IOError('Unexpected header: %s' % code)
       break
-
-def read_input(sock, n, tag):
-  data = sock.recv(n)
-  if not data or len(data) != n:
-    raise IOError('Unexpected end of stream while reading %s.' % tag)
-  return data
 
 if __name__ == '__main__':
   main()
--- a/scripts/stetho_open.py
+++ b/scripts/stetho_open.py
@@ -35,6 +35,17 @@ def stetho_open(device=None, process=None):
 
   return adb.sock
 
+def read_input(sock, n, tag):  
+  data = b'';
+  while len(data) < n:
+    incoming_data = sock.recv(n - len(data))
+    if len(incoming_data) == 0:
+      break
+    data += incoming_data
+  if len(data) != n:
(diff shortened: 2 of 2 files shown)


scripts/dumpapp
```diff
@@ -58,6 +58,7 @@ def read_frames(sock):
     if code == b'1':
       if n > 0:
         sys.stdout.buffer.write(read_input(sock, n, 'stdout blob'))
+        sys.stdout.buffer.flush()
```
jasta: Oops, @rightparen and I even discussed doing this and I wrong asserted it was not necessary.  I forgot about the streaming use cases :)

scripts/dumpapp
```diff
@@ -76,9 +77,12 @@ def read_frames(sock):
         raise IOError('Unexpected header: %s' % code)
       break
 
-def read_input(sock, n, tag):
-  data = sock.recv(n)
-  if not data or len(data) != n:
+def read_input(sock, n, tag):  
+  data = b'';
+  while len(data) < n:
+    incoming_data = sock.recv(n - len(data))
```
jasta: sock.recv should return an empty byte object (len(incoming_data) == 0) on EOF/error.  You have to check for this otherwise the client script will hang when the server terminates (for instance with `./scripts/dumpapp crash exit`)

scripts/stetho_open.py
```diff
@@ -121,7 +121,7 @@ def select_service(self, service):
   def _read_exactly(self, num_bytes):
     buf = b''
     while len(buf) < num_bytes:
-      new_buf = self.sock.recv(num_bytes)
+      new_buf = self.sock.recv(num_bytes - len(buf))
```
jasta: Testing fail.  Also, prefer that you expose this function for `dumpapp` to share in common and fix the new_buf being empty bug here as well.  The correct response is something like `raise IOError('EOF')`

jasta:
Great work, thanks for helping out with this!

scripts/stetho_open.py
```diff
@@ -107,24 +118,17 @@ def connect(self, port=5037):
...
-    status = self._read_exactly(4)
+    status = read_input(self.sock, 4, "status")
     if status == b'OKAY':
       # All good...
       pass
     elif status == b'FAIL':
-      reason_len = int(self._read_exactly(4), 16)
-      reason = self._read_exactly(reason_len).decode('ascii')
+      reason_len = int(read_input(self.sock, 4, "fail reason"), 16)
+      reason = read_input(self.sock, reason_len, "fail reason lean").decode('ascii')
```
jasta: typo: len

jasta:
lgtm minus the nit, will merge after Travis finishes.

scripts/stetho_open.py
mottis: Ugh!

jasta:
Just kicked the build.  If it still fails, go ahead and rebase your changes onto the latest master and try again.

PR #353, opened 2016-02-11
Added a utility method to convert args iterator to array of args.

mottis:
This can come handy, especially when working with libraries that require an array of string to parse commands, such as Commons CLI's CommandLineParser.

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/dumpapp/ArgsHelper.java
+++ b/stetho/src/main/java/com/facebook/stetho/dumpapp/ArgsHelper.java
@@ -9,7 +9,9 @@
 
 package com.facebook.stetho.dumpapp;
 
+import java.util.ArrayList;
 import java.util.Iterator;
+import java.util.List;
 
 public class ArgsHelper {
   public static String nextOptionalArg(Iterator<String> argsIter, String defaultValue) {
@@ -22,5 +24,13 @@ public class ArgsHelper {
       throw new DumpUsageException(errorIfMissing);
     }
     return argsIter.next();
+  }
+
+  public static String[] drainToArray(Iterator<String> argsIter) {
+    List<String> args = new ArrayList<>();
+    while (argsIter.hasNext()) {
+      args.add(argsIter.next());
+    }
+    return args.toArray(new String[args.size()]);
   }
 }


rickbrew:
It might be handy, but where's it being used? Why wouldn't you just iterate through the iterator? We try to keep Stetho small for the sake of APK size, and I don't see how this fits.

stetho/src/main/java/com/facebook/stetho/dumpapp/ArgsHelper.java
```diff
@@ -23,4 +25,12 @@ public static String nextArg(Iterator<String> argsIter, String errorIfMissing)
     }
     return argsIter.next();
   }
+
+  public static String[] argsIterToArray(Iterator<String> argsIter) {
```
jasta: Prefer a name consistent with the other methods in this class such as `toArray`.  All `ArgsHelper` methods assume you have an `Iterator<String>`.

stetho/src/main/java/com/facebook/stetho/dumpapp/ArgsHelper.java
```diff
@@ -23,4 +25,12 @@ public static String nextArg(Iterator<String> argsIter, String errorIfMissing)
...
+    List<String> args = new ArrayList<>();
```
jasta: Specify the capacity explicitly.  It's much faster :)

jasta: Oops, nevermind, you cannot :)

rickbrew: `Iterator` doesn't provide a `size()` method though.

jasta:
@rickbrew this is actually pretty common to need for any caller that uses sub-option parsing, for instance to handle:

    dumpapp mycommand something --verbose

You would take "something" off the `Iterator<String>` then pass it to some other function which would use `CommandLineParser` to parse the remaining args ("--verbose") but those args have to be passed in as a `String[]` not an `Iterator<String>`.

rickbrew:
okay :)

stetho/src/main/java/com/facebook/stetho/dumpapp/ArgsHelper.java
jasta: I just realized this method has a tragic side effect of exhausting the iterator as well which should also be captured in the name.  Prefer instead then `drainToArray` or something that communicates it has this side effect.

jasta:
Man what is going on with Travis :(

stetho/src/main/java/com/facebook/stetho/dumpapp/ArgsHelper.java
mottis: Renamed method to `drainToArray`.

jasta:
Force merging without Travis...

Issue #84, opened 2015-03-10
Database journal mode changed errors


Tolriq:
When using Stetho to inspect database it seems it tries to change the journal mode of the database.

My app force writeAhead and keep a connection to the db the whole time so this generate logcat errors :

```
03-10 20:00:38.057    6477-7353/xxx E/SQLiteLog﹕ (5) statement aborts at 2: [PRAGMA journal_mode=PERSIST]
03-10 20:00:38.057    6477-7353/xxx W/SQLiteConnection﹕ Could not change the database journal mode of '/data/data/xxx/databases/xxx.db' from 'wal' to 'PERSIST' because the database is locked.  This usually means that there are other open connections to the database which prevents the database from enabling or disabling write-ahead logging mode.  Proceeding without changing the journal mode.
03-10 20:00:39.088    6477-7353/xxx E/SQLiteLog﹕ (5) statement aborts at 2: [PRAGMA journal_mode=PERSIST]
03-10 20:00:39.088    6477-7353/xxx W/SQLiteConnection﹕ Could not change the database journal mode of '/data/data/xxx/databases/xxx.db' from 'wal' to 'PERSIST' because the database is locked.  This usually means that there are other open connections to the database which prevents the database from enabling or disabling write-ahead logging mode.  Proceeding without changing the journal mode.
```

Is there a reason for that ? What are the impacts of the changes failing ?
And if you are just trying to activate writeAhead as the error suggest maybe just check before if it's not the case ?

jasta:
The issue, I believe, is that we open the database handle ourselves rather than trying to re-use an existing connection which means that the current process may have already opened the database using different flags.  Attempting to re-open it may reconfigure details like write-ahead logging, which cannot be done while the database is locked.

For this to work, I think we'd need some way of detecting the correct configuration and applying it, but I'm not extremely familiar with this problem space and would need to investigate much deeper.  Maybe it works to just check for a `-wal` file? :)

Tolriq:
Well this sounds more complex indeed and in the end is not error prone due to concurrency issues :(

I suppose locks would prevents database corruption and just throws errors in Stetho.

Maybe for applications that needs to keep database opens add a callback in the instance builder to ask the application for the SQLiteConnection or whatever Stetho needs on target database ?

facebook-github-bot-7:
Thank you for reporting this issue and appreciate your patience. We've  notified the core team for an update on this issue. We're looking for a response within the next 30 days or the issue may be closed.

[...]

PR #366, opened 2016-03-23
Database connection provider

frett:
Allows users to work around #84 by providing their own Connection Provider that returns their active Database handle.

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/database/DefaultDatabaseConnectionProvider.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/database/DefaultDatabaseConnectionProvider.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.database;
+
+import android.content.Context;
+import android.database.sqlite.SQLiteDatabase;
+import android.database.sqlite.SQLiteException;
+import com.facebook.stetho.common.Util;
+
+import java.io.File;
+
+/**
+ * Opens the requested database using {@link SQLiteDatabase#openDatabase(String, SQLiteDatabase.CursorFactory, int)} directly.
+ */
+public class DefaultDatabaseConnectionProvider implements DatabaseConnectionProvider {
+  private final Context mContext;
+
+  public DefaultDatabaseConnectionProvider(Context context) {
+    mContext = context;
+  }
+
+  @Override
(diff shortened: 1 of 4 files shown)


mikeyxkcd:
This is a much more flexible alternative to the #367 pull request.

Could one of the senior members like @jasta or @rickbrew look into merging this?

I don't want to have to fork and create my own local maven repo just for this single bug fix.

Thanks in advance as this has been plaguing me all week!

stetho/src/main/java/com/facebook/stetho/inspector/database/DefaultDatabaseConnectionProvider.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.io.File;
+
+/**
+ * Opens the requested database using {@link SQLiteDatabase#openDatabase(String, SQLiteDatabase.CursorFactory, int)} directly.
+ */
+public class DefaultDatabaseConnectionProvider implements DatabaseConnectionProvider {
+  private final Context mContext;
+
+  public DefaultDatabaseConnectionProvider(Context context) {
```
jasta: I'd like to see this specific provider become a bit more flexible so that you could easily express something like:
```java
  new DefualtInspectorModulesBuilder(...)
    .databaseConnections(DefaultDatabaseConnectionProvider.createForWAL(context))
    ...
```

jasta:
Thinking about this at a high level I think it makes most sense to include the API along with SqliteDatabaseDriver instead of as a new top-level feature (for instance, what if I provide both `.databaseConnections(...) and `.provideDatabaseDriver(...)`?  The two generally appear in conflict with one another.  To solve this, I propose the following high-level configuration:

```java
new DefaultInspectorModulesBuilder(...)
    .provideDatabaseDriver(
        new SqliteDatabaseDriver(
            context,
            new DefaultDatabaseFilesProvider(context),
            DefaultSqliteConnectionProvider.createForWAL(context)))
   ...
```  

Internally in the builder we would observe that there is already a `SqliteDatabaseDriver` provided and avoid configuring the default one.  This advice pairs nicely with #367 which attempts to work around a pretty serious gotcha whereby the open provider and the files provider need to have symmetric APIs (the files provider returns files, but the connection provider receives string file names creating a serious problem in practice).

I think the two diff can proceed more or less independently but I will unify them with a mapping of databaseId (name that appears in the UI) to File and add a bit of disambiguation logic in the core of Stetho.

jasta:
Actually, looking at this again, I think we can accept this diff as-is, but just adjust the API as part of addressing this and #367 in tandem before we issue a public release (as that would be a public API change).

mikeyxkcd:
Thanks a lot for looking into this @jasta, and for @frett for implementing the pull-request. I really appreciate it.

Is there a release plan for when the new bug fixes will be deployed to maven central?

Thanks

jasta:
Yes, a release is expected within the next few weeks incorporating elements tab bug fixes and these database driver improvements.

PR #367, opened 2016-03-29
Fixed SqliteDatabaseDriver with custom DatabaseFilesProvider

jferlisi:
When opening a table the DatabaseFilesProvider is first checked for the table then if not found there we fall back to the context.getDatabasePath() call.

If a custom DatabaseFilesProvider is used then the context.getDatabasePath() won't find those databases.

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/database/SqliteDatabaseDriver.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/database/SqliteDatabaseDriver.java
@@ -201,12 +201,22 @@ public class SqliteDatabaseDriver extends Database.DatabaseDriver {
 
   private SQLiteDatabase openDatabase(String databaseName) throws SQLiteException {
     Util.throwIfNull(databaseName);
-    File databaseFile = mContext.getDatabasePath(databaseName);
+    File databaseFile = findDatabaseFile(databaseName);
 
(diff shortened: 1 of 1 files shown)


misakuo:
### Bugs with 'DefaultDatabaseFilesProvider'
On `DefaultDatabaseFilesProvider`, the database file was added by 
```
databaseFiles.add(new File(filename));
```
It created a invalid file on directory `/` instead of database dir , so, when you traversal `mDatabaseFilesProvider` on `findDatabaseFile`, it will return the invalid file.

mikeyxkcd:
I just ran into this bug while trying to read my local WebView's cookies database located in `/data/data/com.example.app/app_webview/Cookies` and `/data/data/com.example.app/app_webview/Web Data` which were not being picked up by the `Context#databaseList()` call.

Could one of the senior members like @jasta look into merging this?
Thanks.

jasta:
Hmm, there appear to be deeper issues here than I initially realized.  This solution isn't incorrect by itself, but highlights a greater issue whereby our databaseId is the file name itself failing to recognize that it could create a conflict with multiple database names provided across the system.  We appear to need a disambiguation strategy and a proper mapping between databaseId and location.

jasta:
This fix looks locally correct and for that reason we can merge it in but I need to do some work to correct the larger problem here...

PR #387, opened 2016-05-31
Accessibility

blavalla:
Added the basic start for accessibility inspection, and a wrapper class to hold all of the methods for various accessibility properties I plan to add to the inspector.  For now these will appear as a second CSS Selector alongside other view properties.  The only one currently implemented is "ignored", which signifies whether or not a node will be ignored by accessibility services.

I plan to follow the properties of the AXNode type in the Chrome Debugging Protocol (http://chromedevtools.github.io/debugger-protocol-viewer/tot/Accessibility/#type-AXNode).  This will make it easier to transition to the Accessibility Panel if Stetho is upgraded past the 1.1 version of the protocol.

I'm not sure if this is a problem, but by adding the "getStyleFromBoolean" and getStyleFromCharString" methods into ViewDescriptor.java, I have exposed a lot more of the methods/fields of the standard View class.  If that's not desired, I can change the approach here.

Screenshot:
screen shot 2016-05-31 at 2 33 01 pm

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/AbstractChainedDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/AbstractChainedDescriptor.java
@@ -155,4 +155,14 @@ public abstract class AbstractChainedDescriptor<E> extends Descriptor implements
 
   protected void onGetStyles(E element, StyleAccumulator accumulator) {
   }
+
+  @Override
+  @SuppressWarnings("unchecked")
+  public final void getAccessibilityStyles(Object element, StyleAccumulator accumulator) {
+    mSuper.getAccessibilityStyles(element, accumulator);
+    onGetAccessibilityStyles((E) element, accumulator);
+  }
+
+  protected void onGetAccessibilityStyles(E element, StyleAccumulator accumulator) {
+  }
 }
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AccessibilityNodeInfoWrapper.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AccessibilityNodeInfoWrapper.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.inspector.elements.android;
+
+import android.support.v4.view.ViewCompat;
+import android.support.v4.view.accessibility.AccessibilityNodeInfoCompat;
+import android.view.View;
+import android.view.ViewParent;
+
+public final class AccessibilityNodeInfoWrapper {
+
+  public AccessibilityNodeInfoWrapper() {
+  }
+
+  public static boolean getIgnored(View view, AccessibilityNodeInfoCompat nodeInfo) {
+    int important = ViewCompat.getImportantForAccessibility(view);
+
+    if (important == ViewCompat.IMPORTANT_FOR_ACCESSIBILITY_NO ||
+        important == ViewCompat.IMPORTANT_FOR_ACCESSIBILITY_NO_HIDE_DESCENDANTS) {
+      return true;
+    }
+
+    // Go all the way up the tree to make sure no parent has hidden its descendants
+    ViewParent parent = view.getParent();
+    while (parent instanceof View) {
+      if (ViewCompat.getImportantForAccessibility((View) parent)
+          == ViewCompat.IMPORTANT_FOR_ACCESSIBILITY_NO_HIDE_DESCENDANTS) {
+        return true;
+      }
+      parent = parent.getParent();
+    }
+
+    // AccessibilityNodeInfo is actionable
+    if (nodeInfo.isClickable() || nodeInfo.isLongClickable() || nodeInfo.isFocusable()) {
+      return false;
+    }
+
+    // AccessibilityNodeInfo has text set (commonly set from TextViews and Switches).
+    CharSequence nodeText = nodeInfo.getText();
+    if (nodeText != null && nodeText.length() > 0) {
+      return false;
+    }
+
+    // AccessibilityNodeInfo has a content description
+    CharSequence contentDescription = nodeInfo.getContentDescription();
+    if (contentDescription != null && contentDescription.length() > 0) {
+      return false;
+    }
+
+    // View has an AccessibilityNodeProvider
+    if (ViewCompat.getAccessibilityNodeProvider(view) != null) {
+      return false;
+    }
+
+    // View has an AccessibilityDelegate
+    if (ViewCompat.hasAccessibilityDelegate(view)){
+      return false;
+    }
+
+    return true;
+  }
+}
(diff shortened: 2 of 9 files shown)


blavalla:
Alright, this should be rebased with master so that the changes from #399 included.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AccessibilityNodeInfoWrapper.java
```diff
@@ -0,0 +1,53 @@
...
+  private AccessibilityNodeInfoCompat mNodeInfo;
+  private View mView;
+  private ViewParent mParent;
+
+  public AccessibilityNodeInfoWrapper(AccessibilityNodeInfoCompat nodeInfo, View view) {
+    mNodeInfo = nodeInfo;
+    mView = view;
+    mParent = view.getParent();
+  }
+  @ViewDebug.ExportedProperty(category = "accessibility")
```
rickbrew: nit: line of whitespace between methods

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AccessibilityNodeInfoWrapper.java
```diff
@@ -0,0 +1,53 @@
...
+import android.support.v4.view.accessibility.AccessibilityNodeInfoCompat;
+import android.view.View;
+import android.view.ViewDebug;
+import android.view.ViewParent;
+import android.widget.TextView;
+
+public class AccessibilityNodeInfoWrapper {
+  private AccessibilityNodeInfoCompat mNodeInfo;
+  private View mView;
+  private ViewParent mParent;
```
rickbrew: can these be made `final` ? if so, please do so

blavalla:
- Added some short-circuits to AccessibilityNodeInfoWrapper.
- Removed a bunch of unnecessary reflection from ViewDescriptor in favor of simply calling the methods from AccessibilityNodeInfoWrapper.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AccessibilityNodeInfoWrapper.java
```diff
@@ -0,0 +1,65 @@
...
+      parent = parent.getParent();
+    }
+
+    // AccessibilityNodeInfo is actionable
+    if (mNodeInfo.isClickable() || mNodeInfo.isLongClickable() || mNodeInfo.isFocusable()) {
+      return false;
+    }
+
+    // AccessibilityNodeInfo has text set (commonly set from TextViews and Switches).
+    if (mNodeInfo.getText() != null && mNodeInfo.getText().length() > 0) {
```
rickbrew: does `getText()` allocate? if it does, or if we don't know, then we should call it once

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AccessibilityNodeInfoWrapper.java
```diff
@@ -0,0 +1,65 @@
...
+      return false;
+    }
+
+    // AccessibilityNodeInfo has a content description
+    if (mNodeInfo.getContentDescription() != null &&
```
rickbrew: does `getContentDescription()` allocate? if it does, or if we don't know, then we should only call it once

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AccessibilityNodeInfoWrapper.java
```diff
@@ -0,0 +1,65 @@
+package com.facebook.stetho.inspector.elements.android;
+
+import android.support.v4.view.ViewCompat;
+import android.support.v4.view.accessibility.AccessibilityNodeInfoCompat;
+import android.view.View;
+import android.view.ViewParent;
+
+public class AccessibilityNodeInfoWrapper {
```
rickbrew: Can we make this into a class with static utility methods? I'm not sure we need an allocation for the sake of a `getIgnored` method with 3 parameters (or even for when/if we add more methods)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AccessibilityNodeInfoWrapper.java
```diff
@@ -0,0 +1,65 @@
+package com.facebook.stetho.inspector.elements.android;
```
rickbrew: new files need copyright added a top. (copy+paste from any other .java file.)

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -99,8 +99,13 @@ public JsonRpcResult getMatchedStylesForNode(JsonRpcPeer peer, JSONObject params
...
 
     final RuleMatch match = new RuleMatch();
+    final RuleMatch accessibilityMatch = getAccessibilityRuleMatch();
 
-    result.matchedCSSRules = ListUtil.newImmutableList(match);
+    List<RuleMatch> matchedCSSRules = new ArrayList<>();
+    matchedCSSRules.add(match);
+    matchedCSSRules.add(accessibilityMatch);
+
+    result.matchedCSSRules = ListUtil.copyToImmutableList(matchedCSSRules);
```
rickbrew: in this case, since you're always populating the `List` with 2 elements, just use `ListUtil.newImmutableList` directly. You'll need to add a 2 parameter version that instantiates `TwoItemImmutableList`.

stetho/src/main/java/com/facebook/stetho/inspector/elements/AbstractChainedDescriptor.java
```diff
@@ -155,4 +155,14 @@ public final void getStyles(Object element, StyleAccumulator accumulator) {
 
   protected void onGetStyles(E element, StyleAccumulator accumulator) {
   }
+
+  @Override
+  @SuppressWarnings("unchecked")
+  public final void getAccessibilityNodeInfo(Object element, StyleAccumulator accumulator) {
```
rickbrew: maybe call this `getAccessibilityStyles` so it matches up with `getStyles`, and because we're still using `StyleAccumulator`. (and then follow up with the cascading renames elsewhere of course)

rickbrew:
Okay I added another round of notes. Looking pretty good, getting pretty close :)

blavalla:
Latest changes should all be addressed.  Thanks!

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AccessibilityNodeInfoWrapper.java
```diff
@@ -0,0 +1,69 @@
...
+
+import android.support.v4.view.ViewCompat;
+import android.support.v4.view.accessibility.AccessibilityNodeInfoCompat;
+import android.view.View;
+import android.view.ViewParent;
+
+public final class AccessibilityNodeInfoWrapper {
+
+  public AccessibilityNodeInfoWrapper() {
```
rickbrew: prefer a private constructor, or none at all

private constructor helps to mimic "static classes" in C#/VB. not having an explicitly defined constructor is equivalent to having this public parameterless constructor, but is a bit cleaner for us here

rickbrew:
Awesome ... looks good to me 👍  Let me know when yo'ud like me to merge

PR #391, opened 2016-06-02
Improve support for custom SQLite open flags

jasta:
This diff adds a WAL-guessing heuristic to support most database users
for the common case as well as expands upon
DefaultDatabaseConnectionProvider to fully support subclassing to
conveniently select custom options on a per database file basis.


jasta:
This diff addresses some of my concerns with #366, generally making it more useful by default.

PR #392, opened 2016-06-02
Fix database id mapping to remove guesswork hacks

jasta:
The SqliteDatabaseDriver had to resort to counter intuitive hacks
to manually locate a database by its file name and didn't
properly handle deduplicating multiple databases with the same name.  As
we expanded flexibility we failed to address the possibility (nay,
likelihood) that this will occur and it would create very confusing
results for users.

This diff doesn't address the confusion in the UI (two database entries
of the same name can still occur) but it internally makes it possible to
treat them as separate databases and track their filenames separately.
It also makes it relatively straightforward to fix the uniqueness of the
user visible naming by introducing a "context" parameter to
DatabaseDescriptor which would be used in the case of name ambiguity.

This diff addresses concerns raised in #366 and #367.

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/database/ContentProviderDatabaseDriver.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/database/ContentProviderDatabaseDriver.java
@@ -16,7 +16,6 @@ import android.database.sqlite.SQLiteException;
 
 import com.facebook.stetho.inspector.protocol.module.Database;
 import com.facebook.stetho.inspector.protocol.module.DatabaseDescriptor;
-import com.facebook.stetho.inspector.protocol.module.DatabaseDriver;
 import com.facebook.stetho.inspector.protocol.module.DatabaseDriver2;
 
 import java.util.ArrayList;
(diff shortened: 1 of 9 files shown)


stetho/src/main/java/com/facebook/stetho/inspector/database/ContentProviderDatabaseDriver.java
```diff
@@ -13,38 +13,44 @@
...
-  public ContentProviderDatabaseDriver(Context context, ContentProviderSchema... contentProviderSchemas) {
+  public ContentProviderDatabaseDriver(
+      Context context,
+      ContentProviderSchema... contentProviderSchemas) {
     super(context);
     mContentProviderSchemas = contentProviderSchemas;
   }
 
   @Override
-  public List<String> getDatabaseNames() {
```
jasta: Oops, this is not backwards compatible (either source or binary).  I'll fix that.

stetho/src/main/java/com/facebook/stetho/inspector/database/SqliteDatabaseDriver.java
```diff
@@ -206,18 +212,28 @@ private static String getFirstWord(String s) {
...
+    public final File file;
+
+    public SqliteDatabaseDescriptor(File file) {
+      this.file = file;
+    }
 
-     return mContext.getDatabasePath(databaseName);
-   }
+    @Override
+    public String context() {
```
jasta: Discussed with @longinoa, this approach is misleading and kludgey.  Will follow up with using a namespace / tag field on the driver itself and disambiguating when multiple drivers are in use.

jasta:
Wow this is a big, old diff apparently by me but it does look important.  I think my patch is overall viable but I think it got stalled due to a binary compat issue I'm looking into now

jasta:
@longinoa ping ping :)

jasta:
YOLO

PR #396, opened 2016-06-04
update timber version

mmirhoseini:
This old Timber version causes me many troubles,
so please merge this new version fix

### DIFF
--- a/stetho-timber/build.gradle
+++ b/stetho-timber/build.gradle
@@ -14,7 +14,7 @@ android {
 
 dependencies {
     compile project(':stetho')
-    compile 'com.jakewharton.timber:timber:3.0.1'
+    compile 'com.jakewharton.timber:timber:4.1.2'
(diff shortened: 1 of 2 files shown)


jasta:
Build fails in travis CI?  Looks related, can you investigate?

stetho/src/main/AndroidManifest.xml
```diff
@@ -2,6 +2,9 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.facebook.stetho">
```
kleisauke: You should add this: `xmlns:tools="http://schemas.android.com/tools"` here. Otherwise Travis CI will fail because the XML namespace could not be recognized by the parser.

mmirhoseini:
@kleisauke Thanks for the hint, Build passed! 👍 🙂

PR #399, opened 2016-06-14
Added methods to get styles from other data types

blavalla:
Adds methods to get styles from Boolean, Short, Long, Double, Byte, and CharSequence.  Along with the existing methods for Float, Int, and Object, this should cover all of the View methods and fields with ViewDebug annotations so that all properties are shown in the Styles tab of the Elements panel.

Screenshot showing new properties:
screen shot 2016-06-14 at 1 51 53 pm

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
@@ -243,24 +243,6 @@ final class ViewDescriptor extends AbstractChainedDescriptor<View> implements Hi
     }
   }
 
-  private static boolean isDefaultValue(
-      Float value) {
-    return value == 0.0f;
-  }
-
-  private static boolean isDefaultValue(
-      Integer value,
-      @Nullable ViewDebug.ExportedProperty annotation) {
-    // Mappable ints should always be shown, because enums don't necessarily have
-    // logical "default" values. Thus we mark all of them as not default, so that they
-    // show up in the inspector.
-    if (canFlagsBeMappedToString(annotation) || canIntBeMappedToString(annotation)) {
-      return false;
-    }
-
-    return value == 0;
-  }
-
   private String convertViewPropertyNameToCSSName(String getterName) {
     // Split string by uppercase characters. Thankfully since
     // this is the android source we don't have to worry about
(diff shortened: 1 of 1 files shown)


stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -298,6 +317,18 @@ private void getStyleFromValue(
       getStyleFromInteger(name, (Integer) value, annotation, styles);
     } else if (value instanceof Float) {
       getStyleFromFloat(name, (Float) value, annotation, styles);
+    } else if (value instanceof Boolean) {
```
jasta: missing Character

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -243,8 +243,7 @@ private static String mapFlagsToStringUsingAnnotation(
     }
   }
 
-  private static boolean isDefaultValue(
-      Float value) {
+  private static boolean isDefaultValue(Float value) {
```
jasta: Strongly prefer undoing this pattern.  It increases the number of methods for very weak utility.

jasta:
lgtm, will merge after you respond

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -344,7 +349,63 @@ private void getStyleFromFloat(
...
-    styles.store(name, String.valueOf(value), isDefaultValue(value));
+    styles.store(name, String.valueOf(value), value == 0.0f);
+  }
+
+  private void getStyleFromBoolean(
+      String name,
+      Boolean value,
+      @Nullable ViewDebug.ExportedProperty annotation,
+      StyleAccumulator styles) {
+    styles.store(name, String.valueOf(value), false);
```
jasta: As I look one more time, these methods themselves also do not make sense.  The if / else if blocks above would read so much more succinctly as:

    if (value instanceof Boolean) {
      styles.store(name, String.valueOf(value), false);
    } else { 
      ...

jasta:
One last bit of feedback but it is optional; squash and rebase and we can merge.

blavalla:
I think this should be good to go now.

PR #406, opened 2016-07-05
Fix importClass and importPackage not finding external classes

### LINKED ISSUE #405: Incorrect Rhino initialization
In [`JsRuntimeReplFactoryBuilder.JsRuntimeReplFactoryBuilder`](https://github.com/facebook/stetho/blob/master/stetho-js-rhino/src/main/java/com/facebook/stetho/rhino/JsRuntimeReplFactoryBuilder.java#L87), the default package that is added is the app's packageName, but that may not be the actual Java package name, it's  really just the applicationId. This may cause an exception at launch.

In any case, the [`importPackages`](https://github.com/facebook/stetho/blob/master/stetho-js-rhino/src/main/java/com/facebook/stetho/rhino/JsRuntimeReplFactoryBuilder.java#L204) method doesn't work for the application's packages, since it doesn't prefix the package name with `Packages`, [which is required](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/Rhino/Scripting_Java#External_Packages_and_Classes) to import non-default-java packages.

paour:
For external packages, the package name needs to be prefixed with `Packages.` per https://developer.mozilla.org/en-US/docs/Mozilla/Projects/Rhino/Scripting_Java#External_Packages_and_Classes

Mostly fixes #405

### DIFF
--- a/stetho-js-rhino/src/main/java/com/facebook/stetho/rhino/JsRuntimeReplFactoryBuilder.java
+++ b/stetho-js-rhino/src/main/java/com/facebook/stetho/rhino/JsRuntimeReplFactoryBuilder.java
@@ -189,7 +189,7 @@ public class JsRuntimeReplFactoryBuilder {
     for (Class<?> aClass : mClasses) {
       String className = aClass.getName();
       try {
-        String expression = String.format("importClass(%s)", className);
+        String expression = String.format("importClass(Packages.%s)", className);
         jsContext.evaluateString(scope, expression, SOURCE_NAME, 1, null);
       } catch (Exception e) {
         throw new StethoJsException(e, "Failed to import class: %s", className);
@@ -201,7 +201,7 @@ public class JsRuntimeReplFactoryBuilder {
     // Import the packages that the caller requested
     for (String packageName : mPackages) {
       try {
-        String expression = String.format("importPackage(%s)", packageName);
+        String expression = String.format("importPackage(Packages.%s)", packageName);
         jsContext.evaluateString(scope, expression, SOURCE_NAME, 1, null);
       } catch (Exception e) {
         throw new StethoJsException(e, "Failed to import package: %s", packageName);
--- a/stetho-js-rhino/src/main/java/com/facebook/stetho/rhino/JsRuntimeReplFactoryBuilder.java
+++ b/stetho-js-rhino/src/main/java/com/facebook/stetho/rhino/JsRuntimeReplFactoryBuilder.java
@@ -189,10 +189,17 @@ public class JsRuntimeReplFactoryBuilder {
     for (Class<?> aClass : mClasses) {
       String className = aClass.getName();
       try {
-        String expression = String.format("importClass(Packages.%s)", className);
+        // import from default classes
+        String expression = String.format("importClass(%s)", className);
         jsContext.evaluateString(scope, expression, SOURCE_NAME, 1, null);
       } catch (Exception e) {
-        throw new StethoJsException(e, "Failed to import class: %s", className);
+        try {
+          // import from application classes
+          String expression = String.format("importClass(Packages.%s)", className);
+          jsContext.evaluateString(scope, expression, SOURCE_NAME, 1, null);
+        } catch (Exception e1) {
+          throw new StethoJsException(e1, "Failed to import class: %s", className);
+        }
       }
     }
   }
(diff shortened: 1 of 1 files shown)


facebook-github-bot-5:
Thank you for your pull request and welcome to our community. We require contributors to sign our Contributor License Agreement, and we don't seem to have you on file. In order for us to review and merge your code, please sign up at https://code.facebook.com/cla - and if you have received this in error or have any questions, please drop us a line at cla@fb.com. Thanks!

facebook-github-bot-0:
Thank you for signing our Contributor License Agreement. We can now accept your code for this (and any) Facebook open source project. Thanks!

longinoa:
What does this do with internal packages?

longinoa:
cc @potyl

paour:
Perhaps I can rework the PR to try both internal and application packages, so that the same method would work for both. Or to avoid (presumably rare) collisions, separate methods could be added for internal and external classes/packages.

potyl:
Will the patch still work for internal packages? Another option would be to introduce 2 keywords for external packages, but this could get confusing.

potyl:
I tried the following in my app:
```
importPackage(android.text)
TextUtils.htmlEncode("hello & world");
```
And I got:
> "hello &amp; world"

Can you show an example how this doesn't work?

I also tried with a package from our app (not the default package) and I was able to call `name()` on an enum.

paour:
In my app, without this PR, if I try `importPackage(fr.lelivrescolaire)`, I get an EcmaError `"fr" is not defined`. Importing `android.text` does work, as you described.

I initialize Stetho with

        Stetho.initialize(Stetho.newInitializerBuilder(context)
                .enableWebKitInspector(new InspectorModulesProvider() {
                    @Override
                    public Iterable<ChromeDevtoolsDomain> get() {
                        return new Stetho.DefaultInspectorModulesBuilder(context).runtimeRepl(
                                new JsRuntimeReplFactoryBuilder(context)
                                        .importPackage("fr.lelivrescolaire.db")
                                        .importClass(fr.lelivrescolaire.db.DbManager.class)
                                        .build()
                        ).finish();
                    }
                })
                .build());

potyl:
Can you test that with your patch we can still load `android.text.TextUitls` (sample class) and that all works fine?

paour:
Injecting `Package.` would work for `importPackage` (passed by String), but not for `importClass` (passed by Class object).

Loading without `Packages.` doesn't work for application classes, in my experience (and based on the docs linked in my first comment).

I'll test with built-in and app classes and packages and report shortly.

paour:
With the latest patch, and with the following initialization:

        Stetho.initialize(Stetho.newInitializerBuilder(context)
                .enableWebKitInspector(new InspectorModulesProvider() {
                    @Override
                    public Iterable<ChromeDevtoolsDomain> get() {
                        return new Stetho.DefaultInspectorModulesBuilder(context).runtimeRepl(
                                new JsRuntimeReplFactoryBuilder(context)
                                        .importPackage("fr.lelivrescolaire.db")
                                        .importPackage("android.app")
                                        .importClass(fr.lelivrescolaire.db.DbManager.class)
                                        .importClass(android.text.TextUtils.class)
                                        .build()
                        ).finish();
                    }
                })
                .build());


The following console commands work fine, and no errors are logged to the Android or JS consoles:

                > TextUtils.htmlEncode("hello & world");
                < "hello &amp; world"

                > DbManager.isReplicating()
                < false

paour:
@potyl any news on this?

potyl:
When i tried i wasn't able to reproduce the bug. I even tried adding a fake `fr.lelivrescolaire.db.DbManager` class to a sample app just in case "fr" was causing a problem.

I'll try once more.

potyl:
Ok, I have to reproduce the error and i can. The branch https://github.com/potyl/stetho/tree/bug-406 shows the problem (if we remove the last commit). The last commit is what @paour proposed as a patch.

Note that only his last patch fixes the issues.

longinoa:
This fell off my radar (sorry) @potyl The patch that @paour proposes seems fine to me, especially given your repro.

jasta:
@potyl deferring to you on whether this should be merged.  lgtm but I have a lot less context here.

potyl:
Yes. Merge it (I thought that it was already merged)

PR #410, opened 2016-07-25
Added ignoredReasons to accessibility inspection

blavalla:
Added ignored-reasons as a property to the Accessibility Properties panel, which explains in plain language why a node is being ignored by accessibility.

I've also updated the ignored property to be more accurate to what Talkback and other Accessibility Services actually look for to determine whether or not to ignore a view.

screen shot 2016-08-08 at 12 44 55 pm

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/common/android/AccessibilityUtil.java
+++ b/stetho/src/main/java/com/facebook/stetho/common/android/AccessibilityUtil.java
@@ -1,0 +1,262 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho.common.android;
+
+import android.support.annotation.Nullable;
+import android.support.v4.view.ViewCompat;
+import android.support.v4.view.accessibility.AccessibilityNodeInfoCompat;
+import android.text.TextUtils;
+import android.view.View;
+import android.view.ViewGroup;
+import android.view.ViewParent;
+import android.widget.AdapterView;
+import android.widget.HorizontalScrollView;
+import android.widget.ScrollView;
+import android.widget.Spinner;
+
+import java.util.List;
+
+/**
+ * This class provides utility methods for determining certain accessibility properties of
+ * {@link View}s and {@link AccessibilityNodeInfoCompat}s. It is porting some of the checks from
+ * {@link com.googlecode.eyesfree.utils.AccessibilityNodeInfoUtils}, but has stripped many features
+ * which are unnecessary here.
+ */
+public final class AccessibilityUtil {
+  private AccessibilityUtil() {
+  }
+
+  /**
+   * Returns whether the specified node has text or a content description.
+   *
+   * @param node The node to check.
+   * @return {@code true} if the node has text.
+   */
+  public static boolean hasText(@Nullable AccessibilityNodeInfoCompat node) {
+    if (node == null) {
+      return false;
+    }
+
+    return !TextUtils.isEmpty(node.getText()) || !TextUtils.isEmpty(node.getContentDescription());
+  }
+
+  /**
+   * Returns whether the supplied {@link View} and {@link AccessibilityNodeInfoCompat} would
+   * produce spoken feedback if it were accessibility focused.  NOTE: not all speaking nodes are
+   * focusable.
+   *
+   * @param view The {@link View} to evaluate
+   * @param node The {@link AccessibilityNodeInfoCompat} to evaluate
+   * @return {@code true} if it meets the criterion for producing spoken feedback
+   */
+  public static boolean isSpeakingNode(
+      @Nullable AccessibilityNodeInfoCompat node,
+      @Nullable View view) {
+    if (node == null || view == null) {
+      return false;
+    }
+
+    if (!node.isVisibleToUser()) {
+      return false;
+    }
+
+    int important = ViewCompat.getImportantForAccessibility(view);
+    if (important == ViewCompat.IMPORTANT_FOR_ACCESSIBILITY_NO_HIDE_DESCENDANTS ||
+        (important == ViewCompat.IMPORTANT_FOR_ACCESSIBILITY_NO &&
+            node.getChildCount() <= 0)) {
+      return false;
+    }
+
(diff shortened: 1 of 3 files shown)


stetho/src/main/java/com/facebook/stetho/common/android/AccessibilityUtil.java
```diff
@@ -0,0 +1,252 @@
...
+  private AccessibilityUtil() {
+  }
+
+  /**
+   * Returns whether the specified node has text or a content description.
+   *
+   * @param node The node to check.
+   * @return {@code true} if the node has text.
+   */
+  public static boolean hasText(AccessibilityNodeInfoCompat node) {
```
rickbrew: Since `null` is allowed, please add `@Nullable` to the parameter. (same for any other occurrences here -- I'll spare you the repetition)

stetho/src/main/java/com/facebook/stetho/common/android/AccessibilityUtil.java
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    int important = ViewCompat.getImportantForAccessibility(view);
+    if (important == ViewCompat.IMPORTANT_FOR_ACCESSIBILITY_NO_HIDE_DESCENDANTS ||
+        (important == ViewCompat.IMPORTANT_FOR_ACCESSIBILITY_NO &&
+            node.getChildCount() <= 0)) {
+      return false;
+    }
+
+    return hasText(node) || node.isCheckable() || hasNonActionableSpeakingDescendants(node, view);
```
rickbrew: let's put the `node.isCheckable()` call first, since that is probably the cheapest call to make. `hasText` looks like it has to do a few allocations in the eventual call to `getText()` (iirc), but `isCheckable` is probably just a `return mIsCheckable;` type of call. (and feel free to verify and double check all of this)

stetho/src/main/java/com/facebook/stetho/common/android/AccessibilityUtil.java
```diff
@@ -0,0 +1,252 @@
...
+   * <p>
+   * NOTE: Accessibility services will include these children's descriptions in the closest
+   * focusable ancestor.
+   *
+   * @param view The {@link View} to evaluate
+   * @param node The {@link AccessibilityNodeInfoCompat} to evaluate
+   * @return {@code true} if it has any non-actionable speaking descendants within its subtree
+   */
+  public static boolean hasNonActionableSpeakingDescendants(
+      AccessibilityNodeInfoCompat node, View view) {
```
rickbrew: nit: when wrapping parameters to the next line, please put 1 parameter per line after that

stetho/src/main/java/com/facebook/stetho/common/android/AccessibilityUtil.java
```diff
@@ -0,0 +1,252 @@
...
+
+    if (node == null || view == null || !(view instanceof ViewGroup)) {
+      return false;
+    }
+
+    ViewGroup viewGroup = (ViewGroup) view;
+    for (int i = 0; i < viewGroup.getChildCount(); i++) {
```
rickbrew: for the loop, prefer this slightly more efficient form:

`for (int i = 0, count = viewGroup.getChildCount(); i < count; i++) {`

Then we don't have a virtual function call for every loop iteration!

stetho/src/main/java/com/facebook/stetho/common/android/AccessibilityUtil.java
```diff
@@ -0,0 +1,252 @@
...
+
+      if (childView == null) {
+        continue;
+      }
+
+      AccessibilityNodeInfoCompat childNode = AccessibilityNodeInfoCompat.obtain();
+      ViewCompat.onInitializeAccessibilityNodeInfo(childView, childNode);
+
+      if (isAccessibilityFocusable(childNode, childView)) {
+        childNode.recycle();
```
rickbrew: you can avoid the repetetive calls to `childNode.recycle()` by using try/finally, e.g.

```
AccessibilityNodeInfoCompat childNode = AccessibilityNodeInfoCompat.obtain();

try {
  ViewCompat.onInitializeAccessibilityNodeInfo(childView, childNode);
  if (isAccessibilityFocusable(childNode, childView)) {
    continue;
  }
  if (isSpeakingNode(childNode, childView)) {
    return true;
  }
} finally {
  childNode.recycle();
}
```

stetho/src/main/java/com/facebook/stetho/common/android/AccessibilityUtil.java
```diff
@@ -0,0 +1,252 @@
...
+    if (node == null || view == null) {
+      return false;
+    }
+
+    ViewParent parentView = ViewCompat.getParentForAccessibility(view);
+    if (!(parentView instanceof View)) {
+      return false;
+    }
+
+    AccessibilityNodeInfoCompat parentNode = AccessibilityNodeInfoCompat.obtain();
```
rickbrew: consider using `try { ... } finally { parentNode.recycle(); }' here as well

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AccessibilityNodeInfoWrapper.java
```diff
@@ -37,33 +38,66 @@ public static boolean getIgnored(View view, AccessibilityNodeInfoCompat nodeInfo
...
-    if (ViewCompat.getAccessibilityNodeProvider(view) != null) {
-      return false;
+    return true;
+  }
+
+  public static String getIgnoredReasons(AccessibilityNodeInfoCompat node, View view) {
+    int important = ViewCompat.getImportantForAccessibility(view);
+
+    if (important == ViewCompat.IMPORTANT_FOR_ACCESSIBILITY_NO) {
+      return "View has importantForAccessibility set to 'NO'.";
```
rickbrew: I'm not sure if we'll have to make this localizable at some point ... we can probably wait for @jasta to give us some wisdom on that.

rickbrew:
Ok there's my first round of code review feedback. Looks pretty straightforward, just some easy stuff leftover I think :)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AccessibilityNodeInfoWrapper.java
blavalla: Yeah, since most of the CSS property names will never be localized since they are derived from the method names on the view, I figure it's probably okay to also have the values not be localized.  This would be easy enough to switch out later if we want though.

rickbrew:
Okay, looks good :) I'll merge it ...

PR #414, opened 2016-08-12
Added various other properties to the accessibility inspection

blavalla:
This adds "focusable", "focusable-reasons", "focused", "description", and "actions" to the Accessibility Properties.  I'll detail below what each of these mean:

**Focusable**
This is the inverse of "ignored", and simply states whether or not a view will be able to be focused by an `AccessibilityService`.

**Focusable Reasons**
Similar to ignored-reasons, this states precisely _why_ a view is focusable.  This will help people to understand how to make a view that they don't want to be focusable, not get focus.

**Focused**
This will be true for the currently accessibility-focused view, and false for all others.  This will help determine precisely which view is receiving focus at any given time, which can often be difficult to determine when you have a deep view hierarchy.

**Description**
This is the computed description that `AccessibilityService`s will receive.  If a view has a specific description set via the `contentDescription` attribute, this will likely be it, but actionable views that do not have their own description will create one from their children that are not actionable but have descriptions.  

NOTE:  This does not include extraneous information added by`AccessibilityService`s, such as states (editable, selected, checked, etc.) or gesture hints (double-tap, double-tap and hold, etc.).

**Actions**
A list of all `AccessibilityAction`s that can be performed on the view, including custom actions if there are any defined.

screen shot 2016-08-11 at 5 32 19 pm

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AccessibilityNodeInfoWrapper.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AccessibilityNodeInfoWrapper.java
@@ -9,12 +9,20 @@
 
 package com.facebook.stetho.inspector.elements.android;
 
+import android.support.annotation.Nullable;
 import android.support.v4.view.ViewCompat;
 import android.support.v4.view.accessibility.AccessibilityNodeInfoCompat;
+import android.support.v4.view.accessibility.AccessibilityNodeInfoCompat.AccessibilityActionCompat;
+import android.text.TextUtils;
 import android.view.View;
+import android.view.ViewGroup;
 import android.view.ViewParent;
+import android.widget.EditText;
 
 import com.facebook.stetho.common.android.AccessibilityUtil;
+
+import java.util.ArrayList;
+import java.util.List;
 
 public final class AccessibilityNodeInfoWrapper {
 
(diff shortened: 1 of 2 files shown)


stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AccessibilityNodeInfoWrapper.java
```diff
@@ -100,4 +108,165 @@ public static String getIgnoredReasons(AccessibilityNodeInfoCompat node, View vi
...
+    if (hasText) {
+      return "View has a description and is not actionable, but has no actionable ancestor.";
+    }
+
+    return null;
+  }
+
+  @Nullable
+  public static String getActions(AccessibilityNodeInfoCompat node) {
+    List<CharSequence> actionLabels = new ArrayList<>();
```
rickbrew: should probably use a `StringBuilder` here. Nice thing about SB is that if you only call `toString()` once then it won't do another allocation. It transfers the internal `char[]` to the new `String`.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AccessibilityNodeInfoWrapper.java
```diff
@@ -100,4 +108,165 @@ public static String getIgnoredReasons(AccessibilityNodeInfoCompat node, View vi
...
+            actionLabels.add(action.getLabel());
+          } else {
+            actionLabels.add("unknown");
+          }
+          break;
+      }
+    }
+
+    String actions = String.valueOf(actionLabels);
+    return actions.substring(1, actions.length() -1);
```
rickbrew: why does the string start from offset 1?

rickbrew: and a nit: space between operators plz, `actions.length() -1` (there's no space before the `1`)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AccessibilityNodeInfoWrapper.java
```diff
@@ -100,4 +108,165 @@ public static String getIgnoredReasons(AccessibilityNodeInfoCompat node, View vi
...
+    }
+
+    if (hasNodeText) {
+      return nodeText;
+    }
+
+    // If there are child views and no contentDescription the text of all non-focusable children,
+    // comma separated, becomes the description.
+    if (view instanceof ViewGroup) {
+      String concatChildDescription = "";
```
rickbrew: definitely use a `StringBuilder` here. Doing `+=` with a `String` in a loop is a classic performance anti-pattern for a language/runtime with immutable strings (like Java, C#/VB, etc.)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AccessibilityNodeInfoWrapper.java
```diff
@@ -100,4 +108,165 @@ public static String getIgnoredReasons(AccessibilityNodeInfoCompat node, View vi
...
+      String separator = ", ";
```
rickbrew: should add `final` to this

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AccessibilityNodeInfoWrapper.java
```diff
@@ -100,4 +108,165 @@ public static String getIgnoredReasons(AccessibilityNodeInfoCompat node, View vi
...
+          actionLabels.add("copy");
+          break;
+        case AccessibilityNodeInfoCompat.ACTION_PASTE:
+          actionLabels.add("paste");
+          break;
+        case AccessibilityNodeInfoCompat.ACTION_SET_SELECTION:
+          actionLabels.add("set-selection");
+          break;
+        default:
+          if (action.getLabel() != null) {
```
rickbrew: not sure if `getLabel()` allocates and returns a new `String` each time. If so, it's important to call `getLabel()` only once. In principle I'd do that anyway as a defensive measure.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AccessibilityNodeInfoWrapper.java
```diff
@@ -100,4 +108,169 @@ public static String getIgnoredReasons(AccessibilityNodeInfoCompat node, View vi
...
+    return null;
+  }
+
+  @Nullable
+  public static String getActions(AccessibilityNodeInfoCompat node) {
+    final StringBuilder actionLabels = new StringBuilder();
+    final String separator = ", ";
+
+    for (AccessibilityActionCompat action : node.getActionList()) {
+      if (actionLabels.length() > 0) {
```
rickbrew: this seems like the wrong logic for adding the separator. don't you want to append it for everything except the 1st item? I think this code will result in a string like: `, first, second, third` because if `actionLabels.length() > 0` then you'll still append the separator on the first item

rickbrew: wait, I see what you did here ... you're looking at the length of `actionLabels`, not the length of `node.getActionList()` .... okay, this should work fine :)

rickbrew:
Okay, this looks good to me

cdr13:
Who is getting my facebook listings.

PR #417, opened 2016-08-17
Update build tools and Android plug-in

friederbluemle:
This updates the build tools to `24.0.1` and the Android Gradle plug-in to `2.1.3` (latest versions as of Aug 18, 2016)


friederbluemle:
Rebased and updated to `24.0.2`

PR #449, opened 2016-10-20
Make sure only unfocusable children's descriptions are being co-opted by parents

blavalla:
ViewParents that contain both focusable, and unfocusable children were incorrectly having their description set to the concatenation of all children.  The parent should have only had its description contain the descriptions of the unfocusable children, since the focusable children will retain their own descriptions and focus.


rickbrew:
lgtm 👍

PR #454, opened 2016-11-08
Update Okhttp3 to 3.4.2

kikuchy:
We are using Stetho for debugging applications. It's very useful! We thank for this community.

Sometimes, we are suffering conflicts of dependencies by the Okhttp's version that included by Stetho.

We hope that Stetho uses latest version of Okhttp. Thank you.


kikuchy:
Done! I signed.

longinoa:
Thanks @kikuchy

PR #456, opened 2016-11-12
stetho_open.py should respect ANDROID_ADB_SERVER_PORT

dezfx:
This change now uses ANDROID_ADB_SERVER_PORT when connecting to adb, if
a numeric value is specified. Otherwise the default port (5037) is used.

### DIFF
--- a/scripts/dumpapp
+++ b/scripts/dumpapp
@@ -26,8 +26,14 @@ def main():
   # transport.
   device = os.environ.get('ANDROID_SERIAL')
 
+  # Connect on the overridden port
+  portStr = os.environ.get('ANDROID_ADB_SERVER_PORT', '')
+  port = 5037
+  if portStr.isdigit():
+      port = int(portStr)
+
   try:
-    sock = stetho_open(device, process)
+    sock = stetho_open(device, process, port)
 
     # Send dumpapp hello (DUMP + version=1)
     sock.send(b'DUMP' + struct.pack('!L', 1))
--- a/scripts/dumpapp
+++ b/scripts/dumpapp
@@ -26,11 +26,8 @@ def main():
   # transport.
   device = os.environ.get('ANDROID_SERIAL')
 
-  # Connect on the overridden port
-  portStr = os.environ.get('ANDROID_ADB_SERVER_PORT', '')
-  port = 5037
-  if portStr.isdigit():
-      port = int(portStr)
+  # Connect on the overridden port if specified
+  port = get_adb_server_port()
 
   try:
     sock = stetho_open(device, process, port)
--- a/scripts/stetho_open.py
+++ b/scripts/stetho_open.py
@@ -13,11 +13,27 @@
 ##
 ###############################################################################
 
+import os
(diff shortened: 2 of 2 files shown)


longinoa:
Thanks @dezfx. This will help with non default installations!

`stetho_open` should take a port input. If the port is specified use the specified port. If it is not we should lookup the value from the environment. If that value still isn't specified I think using 5037 is a good fallback.

scripts/dumpapp
```diff
@@ -26,8 +26,14 @@ def main():
   # transport.
   device = os.environ.get('ANDROID_SERIAL')
 
+  # Connect on the overridden port
+  portStr = os.environ.get('ANDROID_ADB_SERVER_PORT', '')
+  port = 5037
+  if portStr.isdigit():
```
longinoa: If `ANDROID_ADB_SERVER_PORT` is not an integer this should throw. adb itself does something similar.

scripts/stetho_open.py
```diff
@@ -78,9 +78,9 @@ def _find_only_stetho_socket(device):
   finally:
     adb.sock.close()
 
-def _connect_to_device(device=None):
+def _connect_to_device(device=None, port=5037):
```
longinoa: does it make sense to have port default to 5037 here?

scripts/stetho_open.py
```diff
@@ -17,12 +17,12 @@
 import struct
 import re
 
-def stetho_open(device=None, process=None):
-  adb = _connect_to_device(device)
+def stetho_open(device=None, process=None, port=5037):
```
longinoa: port should probably default to `None` and perform the same check you performed in `dumpapp` here as well. This should enable scripts that are using just `stetho_open.py` to actually use the env variable as well

dezfx: will do

scripts/stetho_open.py
longinoa: yeah, the convention here is that _ functions are "private" so this would /have/ to be specified

scripts/stetho_open.py
```diff
@@ -13,16 +13,32 @@
...
 ###############################################################################
 
+import os
 import socket
 import struct
 import re
 
-def stetho_open(device=None, process=None):
-  adb = _connect_to_device(device)
+def get_adb_server_port():
```
longinoa: :+1:

scripts/stetho_open.py
```diff
@@ -78,9 +94,9 @@ def _find_only_stetho_socket(device):
   finally:
     adb.sock.close()
 
-def _connect_to_device(device=None):
+def _connect_to_device(device=None, port=None):
   adb = AdbSmartSocketClient()
-  adb.connect()
+  adb.connect(port)
```
longinoa: we should probably throw if `port is None`

dezfx: I was kinda figuring adb would. My python convention ignorance is showing, is parameter validation in internal methods typical?

Issue #381, opened 2016-05-04
NullPointerException


droidandrew:
I'm can't reproduce it every time  (version 1.3.1)

```
Process: com.adoriasoft.quietping, PID: 9303
java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.Integer.intValue()' on a null
at com.facebook.stetho.inspector.elements.Document$4.store(Document.java:388)
at com.facebook.stetho.inspector.elements.ShadowDocument$Update.getChangedElements(ShadowDocument.java:297)
at com.facebook.stetho.inspector.elements.Document.applyDocumentUpdate(Document.java:372)
at com.facebook.stetho.inspector.elements.Document.updateTree(Document.java:327)
at com.facebook.stetho.inspector.elements.Document.access$1200(Document.java:30)
at com.facebook.stetho.inspector.elements.Document$ProviderListener.onPossiblyChanged(Document.java:682)
at com.facebook.stetho.inspector.elements.android.AndroidDocumentProvider$1.run(AndroidDocumentProvider.java:67)
at android.os.Handler.handleCallback(Handler.java:739)
at android.os.Handler.dispatchMessage(Handler.java:95)
at android.os.Looper.loop(Looper.java:145)
at android.app.ActivityThread.main(ActivityThread.java:5951)
at java.lang.reflect.Method.invoke(Native Method)
at java.lang.reflect.Method.invoke(Method.java:372)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1400)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1195)
```

dabluck:
happening on 4.4 samsung tablet.

jasta:
Hmm, this looks like it could be the case that we assume the old parent wasn't also removed causing getIdForObject to yield null (but we assume this can't be and use int).  The fix should be straight forward, something like:
```java
Integer parentNodeId = mObjectIdMpapper.getIdforObject(...);
Integer nodeId = ...;
if (nodeId != null && parentNodeId != null) {
  onChildNodeRemoved(...);
}
```

Can you repro this issue?  Would be interesting to see what the value of newElementInfo and oldElementInfo was to at least confirm that it's reasonable to see this case in practice.

martinobrink:
I'm also seing this issue. Pretty consistently crashes on my end. crashes seen on various Android devices, various Android versions. 

Stetho versions:
[...]

PR #457, opened 2016-11-15
Fixes a few NPEs when unboxing null Integer

### LINKED ISSUE #443: Crashing !!!
09-29 16:46:25.638 27959 27959 E AndroidRuntime: java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.Integer.intValue()' on a null object reference
09-29 16:46:25.638 27959 27959 E AndroidRuntime:        at com.facebook.stetho.inspector.elements.Document$4.store(Document.java:394)
09-29 16:46:25.638 27959 27959 E AndroidRuntime:        at com.facebook.stetho.inspector.elements.ShadowDocument$Update.getChangedElements(ShadowDocument.java:297)
09-29 16:46:25.638 27959 27959 E AndroidRuntime:        at com.facebook.stetho.inspector.elements.Document.applyDocumentUpdate(Document.java:378)
09-29 16:46:25.638 27959 27959 E AndroidRuntime:        at com.facebook.stetho.inspector.elements.Document.updateTree(Document.java:333)
09-29 16:46:25.638 27959 27959 E AndroidRuntime:        at com.facebook.stetho.inspector.elements.Document.access$1200(Document.java:30)
09-29 16:46:25.638 27959 27959 E AndroidRuntime:        at com.facebook.stetho.inspector.elements.Document$ProviderListener.onPossiblyChanged(Document.java:688)
09-29 16:46:25.638 27959 27959 E AndroidRuntime:        at com.facebook.stetho.inspector.elements.android.AndroidDocumentProvider$1.run(AndroidDocumentProvider.java:67)
09-29 16:46:25.638 27959 27959 E AndroidRuntime:        at android.os.Handler.handleCallback(Handler.java:739)
09-29 16:46:25.638 27959 27959 E AndroidRuntime:        at android.os.Handler.dispatchMessage(Handler.java:95)
09-29 16:46:25.638 27959 27959 E AndroidRuntime:        at android.os.Looper.loop(Looper.java:148)
09-29 16:46:25.638 27959 27959 E AndroidRuntime:        at android.app.ActivityThread.main(ActivityThread.java:5417)
09-29 16:46:25.638 27959 27959 E AndroidRuntime:        at java.lang.reflect.Method.invoke(Native Method)
09-29 16:46:25.638 27959 27959 E AndroidRuntime:        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
09-29 16:46:25.638 27959 27959 E AndroidRuntime:        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)

felipecsl:
Fixes #443

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.1.3'
+        classpath 'com.android.tools.build:gradle:2.2.2'
     }
 }
 
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
@@ -10,14 +10,12 @@
 package com.facebook.stetho.inspector.elements;
 
 import android.os.SystemClock;
+
 import com.facebook.stetho.common.Accumulator;
 import com.facebook.stetho.common.ArrayListAccumulator;
 import com.facebook.stetho.common.LogUtil;
 import com.facebook.stetho.inspector.helper.ObjectIdMapper;
 import com.facebook.stetho.inspector.helper.ThreadBoundProxy;
-
-import javax.annotation.Nullable;
-import javax.annotation.concurrent.GuardedBy;
 
 import java.util.ArrayDeque;
 import java.util.ArrayList;
@@ -26,6 +24,9 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Queue;
 import java.util.regex.Pattern;
+
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.GuardedBy;
 
 public final class Document extends ThreadBoundProxy {
   private final DocumentProviderFactory mFactory;
@@ -361,9 +362,11 @@ public final class Document extends ThreadBoundProxy {
         // sub-tree is included automatically, so we don't need to send events for those.
         if (newElementInfo.parentElement == null) {
           ElementInfo oldElementInfo = mShadowDocument.getElementInfo(element);
-          int parentNodeId = mObjectIdMapper.getIdForObject(oldElementInfo.parentElement);
-          int nodeId = mObjectIdMapper.getIdForObject(element);
-          mUpdateListeners.onChildNodeRemoved(parentNodeId, nodeId);
+          Integer parentNodeId = mObjectIdMapper.getIdForObject(oldElementInfo.parentElement);
+          Integer nodeId = mObjectIdMapper.getIdForObject(element);
+          if (parentNodeId != null && nodeId != null) {
+            mUpdateListeners.onChildNodeRemoved(parentNodeId, nodeId);
+          }
         }
 
         // All garbage elements should be unhooked.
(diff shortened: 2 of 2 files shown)


longinoa:
Thanks @felipecsl!

Im not super familiar with this code, do we also need to check the ternary operators in `ChildEventingList#acquire` and `ChildEventingList#addWithEvent` It seems like we would need to.

Also the code in `ChildEventingList` seems to imply that parentId should be -1 when null should we be translating this to -1 rather than not calling `onChildNodeRemoved`?

cc @rickbrew

felipecsl:
@longinoa yeah there's a couple other places where I wasn't sure what to do. I'd be happy to update the ternary operators to handle null too

rickbrew:
Do we have a reliable (and hopefully minimal) repro for this? I'm afraid this is fixing a symptom, but not the actual cause, and will result in incorrect or incoherent state in the elements tree. This code was very carefully constructed, which leads me to believe that using `int` instead of `Integer` is intentional (a cheap way to do an assert, in other words).

Let me spend some time to (finally) investigate these code paths. A repro would be tremendously helpful though so I can catch it in the act.

build.gradle
```diff
@@ -3,7 +3,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.1.3'
+        classpath 'com.android.tools.build:gradle:2.2.2'
```
rickbrew: why the version bump here?

felipecsl: this was added automatically by Android Studio, but i'd be happy to revert

longinoa:
@felipecsl see the latest comment on #381 do you have an easy repro?

felipecsl:
@longinoa I can really easily reproduce on our app by just navigating around, but I'm not sure what exactly triggers it.

longinoa:
@felipecsl it would help @rickbrew to have a local repro, is your app open source?

felipecsl:
Unfortunately not. Maybe @pollux has a repro case?

rickbrew:
For instance, it's unclear from local code inspection whether 1) `oldElementInfo.parentElement` is `null`, or 2) it's non-`null` but missing from `mObjectIdMapper`. Is `getGarbageElements` pushing elements in the wrong order? Is the application itself violating some invariant that I incorrectly thought would hold true? etc.

Even if we have a fix, I'll still need a repro in order to really validate that the algorithm and data structures for this are good to go.

build.gradle
rickbrew: okie dokie 👍

felipecsl:
I agree that blindly fixing by doing defensive checks is bad. In our case, the `Integer` returned from `getIdForObject()` was null

felipecsl:
We're using React Native, if that matters. It seems like there could be a correlation between that and the crash

rickbrew:
So you have your own (non-minimal) repro? What is `oldElementInfo.parentElement` ? Is it a `Fragment` with `setRetainInstance(true)` ?

longinoa:
See #459 Thanks to @Axolot1 looks like a repro

felipecsl:
OK that's great!

longinoa:
Hey @felipecsl Thanks for the PR but there was a larger logic issue in play. See the commit https://github.com/facebook/stetho/commit/37e615ed293ab49073f57b37e51fdcc6334cf919

Stetho has been updated to 1.4.2 to address this issue.

PR #465, opened 2016-12-06
Fix ObjectMapper's treatment of static fields

rickbrew:
We should not be setting static fields from JSON data, or serializing them into JSON data. This prevents Stetho from working with Android Studio's Instant Run since it injects some stuff into there. This can manifest itself as a blank Elements tab, as well as no incoming data on any of the other tabs.


longinoa:
:+1: Thanks @rickbrew

xiabi:
Hi. How can I update the jar, your commit didn't  come with a new version no.

longinoa:
@xiabi we are still working on a couple of more fixes before releasing a new version. You can pull master down and build locally in the mean time.

xiabi:
thanks for your work.

but it seem that this problem still exists.

image

PR #481, opened 2017-01-14
	Introduce nextRequestId() method to ensure its uniqueness

### LINKED ISSUE #468: Multiple instance of StethoInterceptor and mNextRequestId
`StethoInterceptor` instance can be created multiple times, but it have `mNextRequestId` which should be application-wide unique value (as `NetworkEventReporterImpl` seems require it).

I have two okhttp client: picasso's one with cache and retrofit's one without cache.
Creating `StethoInterceptor` each time I create OkHttpClient instance (i.e. calling `addNetworkInterceptor()`) violates `mNextRequestId` (application-wide) uniqueness.

I think it'd be better to make mNextRequestId singleton which is shared by all kind of interceptors.

ypresto:
Fixes #468

### DIFF
--- a/stetho-okhttp/src/main/java/com/facebook/stetho/okhttp/StethoInterceptor.java
+++ b/stetho-okhttp/src/main/java/com/facebook/stetho/okhttp/StethoInterceptor.java
@@ -29,7 +29,6 @@ import javax.annotation.Nullable;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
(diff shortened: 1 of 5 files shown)


ypresto:
I created another PR #482 which is simpler (without singleton)..!

ypresto:
Adding method instead of singleton might be better.
https://github.com/facebook/stetho/pull/482#discussion_r105539390

ypresto:
> It's already a Singleton

I agree with you..!

ypresto:
Revised PR!

PR #482, opened 2017-01-14
Ensure uniqueness of requestId across interceptors

### LINKED ISSUE #468: Multiple instance of StethoInterceptor and mNextRequestId

ypresto:
Fixes #468

### DIFF
--- a/stetho-okhttp/src/main/java/com/facebook/stetho/okhttp/StethoInterceptor.java
+++ b/stetho-okhttp/src/main/java/com/facebook/stetho/okhttp/StethoInterceptor.java
@@ -46,11 +46,11 @@ import java.util.concurrent.atomic.AtomicInteger;
 public class StethoInterceptor implements Interceptor {
   private final NetworkEventReporter mEventReporter = NetworkEventReporterImpl.get();
 
-  private final AtomicInteger mNextRequestId = new AtomicInteger(0);
+  private static final AtomicInteger sNextRequestId = new AtomicInteger(0);
 
   @Override
   public Response intercept(Chain chain) throws IOException {
-    String requestId = String.valueOf(mNextRequestId.getAndIncrement());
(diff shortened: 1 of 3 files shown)


stetho-okhttp/src/main/java/com/facebook/stetho/okhttp/StethoInterceptor.java
```diff
@@ -46,11 +46,11 @@
 public class StethoInterceptor implements Interceptor {
   private final NetworkEventReporter mEventReporter = NetworkEventReporterImpl.get();
 
-  private final AtomicInteger mNextRequestId = new AtomicInteger(0);
+  private static final AtomicInteger sNextRequestId = new AtomicInteger(0);
```
jasta: This creates some unique problems of its own since now each interceptor is maintaining its own sequence counting instead of sharing the same singleton state across them all.  You use a static prefix to combat this issue but ultimately what I'd prefer to see is the interceptors centralize and all use a single global which ensures uniqueness.

For instance:

    long nextId = mEventReporter.nextRequestId();

Should solve your problem nicely if all the interceptors (including any custom ones you have) are updated.

jasta:
I think you've exposed a legitimate bug here but I would prefer a simpler, more centralized fix.

ypresto:
I firstly created #481 which adds singleton to store next request id. How about this one?

ypresto:
Deprecate this in favor of #481.

PR #487, opened 2017-02-08
Make NodeDescriptor, Descriptor, and HighlightableDescriptor generic

emilsjolander:
Previously only AbstractChainedDescriptor was generic. Anyone extending
directly from Descriptor or implementing HighlightableDescriptor had to
make sure to add casts from Object to their concrete type. This diff
solves that by making all descriptors generic.

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/AbstractChainedDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/AbstractChainedDescriptor.java
@@ -34,11 +34,13 @@ import javax.annotation.Nullable;
  * @param <E> the class that this descriptor will be describing for {@link DocumentProvider},
  * {@link Document}, and ultimately {@link DOM}.
  */
-public abstract class AbstractChainedDescriptor<E> extends Descriptor implements ChainedDescriptor {
+public abstract class AbstractChainedDescriptor<E>
+    extends Descriptor<E> implements ChainedDescriptor<E> {
 
-  private Descriptor mSuper;
+  private Descriptor<? super E> mSuper;
 
-  public void setSuper(Descriptor superDescriptor) {
+  @Override
+  public void setSuper(Descriptor<? super E> superDescriptor) {
     Util.throwIfNull(superDescriptor);
 
     if (superDescriptor != mSuper) {
@@ -49,26 +51,24 @@ public abstract class AbstractChainedDescriptor<E> extends Descriptor implements
     }
   }
 
-  final Descriptor getSuper() {
+  final Descriptor<? super E> getSuper() {
     return mSuper;
   }
 
   @Override
-  @SuppressWarnings("unchecked")
-  public final void hook(Object element) {
+  public final void hook(E element) {
     verifyThreadAccess();
     mSuper.hook(element);
-    onHook((E) element);
+    onHook(element);
   }
 
   protected void onHook(E element) {
   }
 
   @Override
-  @SuppressWarnings("unchecked")
-  public final void unhook(Object element) {
+  public final void unhook(E element) {
     verifyThreadAccess();
-    onUnhook((E) element);
+    onUnhook(element);
     mSuper.unhook(element);
   }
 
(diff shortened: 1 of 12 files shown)


stetho/src/main/java/com/facebook/stetho/inspector/elements/ChainedDescriptor.java
```diff
@@ -53,5 +53,5 @@
  * {@link AbstractChainedDescriptor}) which solves all of these issues for you.<p/>
  */
 public interface ChainedDescriptor {
-  void setSuper(Descriptor superDescriptor);
+  void setSuper(Descriptor<Object> superDescriptor);
```
rickbrew: this should probably be either `Descriptor<?>` or just `Descriptor`

stetho/src/main/java/com/facebook/stetho/inspector/elements/Descriptor.java
```diff
@@ -17,7 +17,7 @@
 import java.util.HashMap;
 import java.util.Map;
 
-public abstract class Descriptor implements NodeDescriptor {
+public abstract class Descriptor<T> implements NodeDescriptor<T> {
```
rickbrew: let's change `<T>` to `<E>`

stetho/src/main/java/com/facebook/stetho/inspector/elements/NodeDescriptor.java
```diff
@@ -14,27 +14,27 @@
 
 import javax.annotation.Nullable;
 
-public interface NodeDescriptor extends ThreadBound {
-  void hook(Object element);
+public interface NodeDescriptor<T> extends ThreadBound {
```
rickbrew: let's change `<T>` to `<E>`

stetho/src/main/java/com/facebook/stetho/inspector/elements/ChainedDescriptor.java
rickbrew: actually ... it'd be great if we could somehow make it so that this is some form like `Descriptor<? super E>`, so that `AbstractChainedDescriptor<E>` can enforce type safety better

It may be best to leave `ChainedDescriptor#setSuper` as accepting a raw `Descriptor` and do the `<? super E>` thing only in the derived class. (java generics are weird ... that might even work)

stetho/src/main/java/com/facebook/stetho/inspector/elements/AbstractChainedDescriptor.java
```diff
@@ -34,11 +34,13 @@
  * @param <E> the class that this descriptor will be describing for {@link DocumentProvider},
  * {@link Document}, and ultimately {@link DOM}.
  */
-public abstract class AbstractChainedDescriptor<E> extends Descriptor implements ChainedDescriptor {
+public abstract class AbstractChainedDescriptor<E>
+    extends Descriptor<E> implements ChainedDescriptor {
 
-  private Descriptor mSuper;
+  private Descriptor<Object> mSuper;
```
rickbrew: see my comments on `ChainedDescriptor` ... I'd love it if we can make this `Descriptor<? super E>`

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/HighlightableDescriptor.java
```diff
@@ -13,7 +13,7 @@
 
 import javax.annotation.Nullable;
 
-interface HighlightableDescriptor {
+public interface HighlightableDescriptor<T> {
```
rickbrew: `T` -> `E`

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/HighlightableDescriptor.java
```diff
@@ -13,7 +13,7 @@
 
 import javax.annotation.Nullable;
 
-interface HighlightableDescriptor {
+public interface HighlightableDescriptor<E> {
```
rickbrew: do we need to make this public? seems like this is something for a later diff, right?

PR #488, opened 2017-02-09
Add ability to register custom descriptors

emilsjolander:
This commit introduces the notion of a DescriptorProvider. A
DescriptorProvider is an object which given a DescriptorMap will
register a number of Descriptors. A list of DescriptorProviders can be
passed to the AndroidDocumentProviderFactory to add or override
descriptors for class of your choosing. This functionality enables
applications and libraries to add their own descriptors while relying
on Stetho defaults for all other settings.

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDocumentProvider.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDocumentProvider.java
@@ -28,6 +28,7 @@ import com.facebook.stetho.common.Util;
 import com.facebook.stetho.common.android.ViewUtil;
 import com.facebook.stetho.inspector.elements.DocumentProvider;
 import com.facebook.stetho.inspector.elements.Descriptor;
+import com.facebook.stetho.inspector.elements.DescriptorProvider;
 import com.facebook.stetho.inspector.elements.DescriptorMap;
 import com.facebook.stetho.inspector.elements.DocumentProviderListener;
 import com.facebook.stetho.inspector.elements.NodeDescriptor;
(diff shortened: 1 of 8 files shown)


stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDocumentProvider.java
```diff
@@ -89,8 +93,13 @@ public AndroidDocumentProvider(Application application, ThreadBound enforcer) {
         .register(TextView.class, new TextViewDescriptor())
         .register(View.class, new ViewDescriptor())
         .register(ViewGroup.class, new ViewGroupDescriptor())
-        .register(Window.class, new WindowDescriptor())
-        .setHost(this)
+        .register(Window.class, new WindowDescriptor());
+
+    for (DescriptorProvider descriptorProvider : descriptorProviders) {
```
rickbrew: Since we have a `List`, we should avoid the temporary `Iterator` allocation by looping like so:

```
for (int i = 0, size = descriptorProviders.size(); i < size; ++i) {
  final DescriptorProvider descriptorProvider = descriptorProviders.get(i);
  ...
}
```

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDocumentProvider.java
```diff
@@ -71,27 +72,34 @@ public void run() {
...
     mApplication = Util.throwIfNull(application);
     mDocumentRoot = new AndroidDocumentRoot(application);
 
-    mDescriptorMap = new DescriptorMap()
-        .beginInit()
-        .register(Activity.class, new ActivityDescriptor())
-        .register(AndroidDocumentRoot.class, mDocumentRoot)
-        .register(Application.class, new ApplicationDescriptor())
-        .register(Dialog.class, new DialogDescriptor());
+    mDescriptorMap = new DescriptorMap().beginInit();
```
rickbrew: you can still use the fluent "builder syntax" here because `beginInit()` returns `DescriptorMap` and `registerDescriptor()` returns `DescriptorRegistrar`.

and then -- IIRC! -- you can change the return type in `DescriptorMap` so that `registerDescriptor` actually returns `DescriptorMap` even though it's implementing an interface method that returns a narrower type. Java is weird like that (if I'm right, and I'm not 100% sure, but I'm pretty sure I've seen this)

PR #489, opened 2017-02-09
Add ability to highlight sub-bounds of view

emilsjolander:
A descriptor may not map directly to a View. For example if a
Descriptor was created for a Drawable it may occupy only a partial
space in the View. This diff solves this problem for descriptor
highlighting by extending HighlightableDescriptor with a method for
returning the bounds to highlight. Returning null will highlight the
full bounds of the View.

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDocumentProvider.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDocumentProvider.java
@@ -14,6 +14,7 @@ import android.app.Application;
 import android.app.Dialog;
 import android.content.Context;
 import android.graphics.Canvas;
+import android.graphics.Rect;
 import android.view.MotionEvent;
 import android.view.View;
 import android.view.ViewGroup;
@@ -43,6 +44,8 @@ final class AndroidDocumentProvider extends ThreadBoundProxy
     implements DocumentProvider, AndroidDescriptorHost {
   private static final int INSPECT_OVERLAY_COLOR = 0x40FFFFFF;
   private static final int INSPECT_HOVER_COLOR = 0x404040ff;
+
+  private final Rect mHighlightingBoundsRect = new Rect();
 
   private final Application mApplication;
   private final DescriptorMap mDescriptorMap;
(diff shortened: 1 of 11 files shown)


stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDocumentProvider.java
```diff
@@ -138,11 +141,16 @@ public NodeDescriptor getNodeDescriptor(Object element) {
...
 
-    View highlightingView = getHighlightingView(element);
-    if (highlightingView == null) {
+    HighlightableDescriptor highlightableDescriptor = getHighlightableDescriptor(element);
+    if (highlightableDescriptor == null) {
       mHighlighter.clearHighlight();
     } else {
-      mHighlighter.setHighlightedView(highlightingView, color);
+      mHighlightBoundsRect.setEmpty();
+      highlightableDescriptor.getBoundsForHighlighting(element, mHighlightBoundsRect);
```
rickbrew: I'm worried there's a bug hazard here. It looks like we're asking the descriptor for the highlighting bounds for an element, but not really specifying what that's relative to. It's apparent later, when we call `getViewForHighlighting`, what those bounds should be relative to. But these calls are disjoint; the protocol seems confusing or clumsy for the implementor.

Can we merge these methods? e.g. `@Nullable View getViewAndBoundsForHighlighting(Object element, Rect bounds)` ... so basically rename `getViewForHighlighting` and add the `Rect` parameter.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -157,11 +158,15 @@ private static String getIdAttribute(View element) {
   }
 
   @Override
+  @Nullable
   public View getViewForHighlighting(View element) {
     return element;
   }
 
   @Override
+  public void getBoundsForHighlighting(View element, Rect bounds) { }
```
rickbrew: The rest of the code places the closing brace on the next line, so let's maintain that throughout this PR as well

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDocumentProvider.java
rickbrew: Plus, it looks like `getViewForHighlighting` gets called twice -- once within `getHighlightableDescriptor` and then again when we call `setHighlightedView`. That could hurt performance depending on the implementation.

emilsjolander: Good call! I used to have the `Rect` as a return value which is why there were initially split into two methods. I'll post an update later today. Thanks.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDocumentProvider.java
```diff
@@ -319,7 +328,7 @@ public boolean onTouchEvent(MotionEvent event) {
 
           if (event.getAction() != MotionEvent.ACTION_CANCEL) {
             if (view != null) {
-              mHighlighter.setHighlightedView(view, INSPECT_HOVER_COLOR);
+              mHighlighter.setHighlightedView(view, null, INSPECT_HOVER_COLOR);
```
emilsjolander: Will file an issue once this lands

PR #490, opened 2017-02-10
Extended StethoTree with configurable options

SiebeSysmans:
I added the ability to log the tag and set a minumum log level for the StethoTree for Timber.

To not have breaking changes and keep the behaviour the same I overloaded the constructor with configuration parameter. 
If the default constructor is used, a configuration is supplied with the same behaviour as the current Stetho implementation.

I extended from Timber.DebugTree instead because it holds functionality to automatically infer the tag from the calling class. Since Stetho is mainly a debug tool, I evaluated that this change is alright.

### DIFF
--- a/stetho-timber/src/main/java/com/facebook/stetho/timber/StethoTree.java
+++ b/stetho-timber/src/main/java/com/facebook/stetho/timber/StethoTree.java
@@ -17,16 +17,37 @@ import timber.log.Timber;
 
 /**
  * Timber tree implementation which forwards logs to the Chrome Dev console.
+ * This uses a {@link Timber.DebugTree} to automatically infer the tag from the calling class.
  * Plant it using {@link Timber#plant(Timber.Tree)}
  * <pre>
  *   {@code
  *   Timber.plant(new StethoTree())
+ *   //or
+ *   Timber.plant(new StethoTree(
+ *       new StethoTree.Configuration.Builder()
+ *           .showTags(true)
+ *           .minimumPriority(Log.WARN)
+ *           .build()));
  *   }
  * </pre>
  */
-public class StethoTree extends Timber.Tree {
+public class StethoTree extends Timber.DebugTree {
+  private final Configuration mConfiguration;
+
+  public StethoTree() {
+    this.mConfiguration = new Configuration.Builder().build();
+  }
+
+  public StethoTree(Configuration configuration) {
+    this.mConfiguration = configuration;
+  }
+
   @Override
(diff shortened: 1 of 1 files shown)


stetho-timber/src/main/java/com/facebook/stetho/timber/StethoTree.java
```diff
@@ -17,17 +17,38 @@
...
+ *   Timber.plant(new StethoTree(
+ *       new StethoTree.Configuration.Builder()
+ *           .showTags(true)
+ *           .minimumPriority(Log.WARN)
+ *           .build()));
  *   }
  * </pre>
  */
-public class StethoTree extends Timber.Tree {
+public class StethoTree extends Timber.DebugTree {
```
jasta: Hmm, this looks to have pretty significant performance overhead, even if just for debug builds.  Also, changing the base class for `StethoTree` without any configuration option seems a bit heavy-handed since again it seems to have performance impact.

Prefer maybe a `StethoDebugTree` alternative which extends `Timber.DebugTree` and provides this functionality.

stetho-timber/src/main/java/com/facebook/stetho/timber/StethoTree.java
```diff
@@ -54,10 +75,63 @@ protected void log(int priority, String tag, String message, Throwable t) {
...
     CLog.writeToConsole(
         logLevel,
         Console.MessageSource.OTHER,
-        message
+        messageBuilder.toString()
     );
   }
-}
+
+  public static class Configuration {
```
jasta: nit: this appears to be two classes where 1 would do:

  StethoTree tree = new StethoTree.Builder()
    .setStuff()
    .build();

Since the configuration doesn't need to be exposed as its own separate class it can just be merged with StethoTree.

jasta:
lgtm overall but the change to extend DebugTree is worrisome

SiebeSysmans:
Alright, I understand the concern with the DebugTree. I'll look into that.

maks:
It would be preferable to still log to android log if not connected to devtools, eg.
```
if (peerManager == null) {
  Log.println(priority, tag, message);
  return;
}
```

jasta:
Closing to clean up our queue; feel free to re-open with new changes for review.

AlexTrotsenko:
@SiebeSysmans do you have any plans to update and re-open this pull request? 
Your configuration ability  looks better, then custom class in client project.

PR #491, opened 2017-02-12
Update build dependencies

friederbluemle:
This updates build tools to `25.0.2`, Gradle wrapper to `3.4.1`, and the plug-in to `2.3.0` (latest versions as of Mar 14, 2017)

<sub><sup>Please NO GitHub rebase/squash merges. If you'd like me to combine some or all of the commits, please let me know, and I will update my branch. Thank you.</sup></sub>


friederbluemle:
Amended the Gradle wrapper commit for the (now) latest version 3.4

gradle.properties
```diff
@@ -1,2 +1,20 @@
...
+# IDE (e.g. Android Studio) users:
+# Gradle settings configured through the IDE *will override*
+# any settings specified in this file.
+
+# For more details on how to configure your build environment visit
+# http://www.gradle.org/docs/current/userguide/build_environment.html
+
+# Specifies the JVM arguments used for the daemon process.
+# The setting is particularly useful for tweaking memory settings.
+org.gradle.jvmargs=-Xmx1536m
```
jasta: Hmm, what's this all about?  This could affect our Travis builds in theory.

release.gradle
```diff
@@ -5,15 +5,15 @@ def isReleaseBuild() {
     return VERSION_NAME.contains("SNAPSHOT") == false
 }
 
-def getRepositoryUrl() {
+def getMavenRepositoryUrl() {
```
jasta: What's this change about?  Seems unrelated (though not necessarily wrong, I'd prefer to see it in a separate PR for bookkeeping sake).

jasta:
Two iffy changes here, otherwise lgtm

friederbluemle:
@jasta Thanks for the review and sorry for not explaining the non-obvious changes. Actually both changes are directly related to the updates I made. I will explain them individually as a reply to your comments.

gradle.properties
friederbluemle: The reason is that with the default setting of 1GB and a more recent version of the Android build tool chain and Gradle, `dex` will run out of process since it requires a larger heap. Adding this setting allows dex to run in the same process, speeding up the build. This is now also the default setting if you create a new project using the latest version of Android Studio.

During the build, you would encounter this warning if the heap was not explicity set to at least 1.5G:

```
Running dex as a separate process.

To run dex in process, the Gradle daemon needs a larger heap.
It currently has 1024 MB.
For faster builds, increase the maximum heap size for the Gradle daemon to at least 1536 MB.
To do this set org.gradle.jvmargs=-Xmx1536M in the project gradle.properties.
For more information see https://docs.gradle.org/current/userguide/build_environment.html
```

release.gradle
friederbluemle: This change is necessary with a more recent version of Gradle and/or the Android Gradle plugin. It seems like there is a naming clash with the property `repositoryUrl` the auto-generated getter method.

Without this change the build will fail with this error message:

```
FAILURE: Build failed with an exception.

* Where:
Script '[path]/release.gradle' line: 9

* What went wrong:
A problem occurred configuring project ':stetho'.
> Could not get unknown property 'repositoryUrl' for project ':stetho' of type org.gradle.api.Project.
```

friederbluemle:
I rebased my commits, and also amended the version updates to the now-latest versions Gradle Wrapper `3.4.1` (from `3.4`), and Android Gradle plugin `2.3.0` (from `2.2.3`). Please let me know if you have any other questions.

gradle.properties
jasta: Ack'd, sounds good.

release.gradle
jasta: Ack'd, I wonder if Google added the support upstream that we had to manually hack in.  I am a little worried that `./gradlew uploadArchives` may have been broken by this change as well but wasn't invoked during your tests.  I guess we'll just have to take that pain when it comes to release time...

vaiden: This error started with gradle 3.4 and originates from hasProperty() returning true even if the property is indeed missing, but still has an overrding getter.
The simpler fix is to change
```
def getRepositoryUrl() {
    return **hasProperty('repositoryUrl')** ? property('repositoryUrl') : 'https://oss.sonatype.org/service/local/staging/deploy/maven2/'
}
```
Into:
```
def getRepositoryUrl() {
    return **findProperty('repositoryUrl') != null** ? property('repositoryUrl') : 'https://oss.sonatype.org/service/local/staging/deploy/maven2/'
}
```
findProperty() ignores overloaded getters.

More info here:  http://www.devsbedevin.com/the-pains-of-building-android-react-native-from-source/

friederbluemle: @vaiden That makes sense. Thanks for the explanation!! :+1:

PR #492, opened 2017-02-12
Add ability for descriptors to customize element IDs

emilsjolander:
Currently a node id is directly connected to the instance of an object.
This has worked well up to this point. However we are implementing
custom descriptors which describe elements which are pooled under the
hood. When the UI changes we make no guarantees that a logical UI node
is represented by the same java object as before.

The problem with the current implementation in stetho is that this
causes the chrome element inspector to collapse/move nodes unexpectedly
when updating. This commit solve this by allows descriptors to
implement custom stable ids for nodes.

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/common/android/AccessibilityUtil.java
+++ b/stetho/src/main/java/com/facebook/stetho/common/android/AccessibilityUtil.java
@@ -34,10 +34,10 @@ public final class AccessibilityUtil {
   }
 
   /**
-   * Returns whether the specified node has text or a content description.
+   * Returns whether the specified nodeID has text or a content description.
    *
-   * @param node The node to check.
-   * @return {@code true} if the node has text.
+   * @param node The nodeID to check.
+   * @return {@code true} if the nodeID has text.
    */
   public static boolean hasText(@Nullable AccessibilityNodeInfoCompat node) {
     if (node == null) {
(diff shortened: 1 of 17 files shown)


rickbrew:
While the premise of this makes perfect sense, the implementation is a bit complex and performance heavy. Using a new `NodeID` class provides a good encapsulation for what's being done here, but adds O(n) allocations with respect to the element tree. Perhaps a simpler solution is possible.

Is it possible, as well as reasonable, for the `Descriptor` to know when an element has been recycled? If so, we could poke a hole into the `Descriptor.Host` and add a method such as, `invalidateNodeIDForElement(Object element)`. The implementation of that, over in `AndroidDocumentProvider`, could then relay that through to `DocumentProviderListener`. This interface is implemented over in `Document` and could be implemented as just calling `mObjectIdMapper.remove(element)`.

Would that work?

emilsjolander:
Removing the mapping would not solve my use case. In my use case there exists a Logical node in the hierarchy, let's call it `<MyNode/>`. This node is backed by some physical object which may change at any time. I would like the id of a single logical `<MyNode/>` instance to stable across many different instances of the underlying physical memory instance of the backing object.

This is similar to how android adapters provide "stable ids" (https://developer.android.com/reference/android/support/v7/widget/RecyclerView.Adapter.html#hasStableIds()) where the underlying data model may switch memory address when refreshing data but still represents the same logical instance.

PR #493, opened 2017-02-12
Move selector naming to the descriptor

emilsjolander:
Currently two selector names are hard-coded in DOM.java.
"<this_element>" and "Accessibility Properties". With this commit
descriptors are able to themselves choose what the selector name should
be and how their various properties should be distributed among the
selectors.

This enables an element to have styles such as:

```
layout {
  width: 100px;
  height: 100px;
}

props {
  text: "Hello World";
}
```

Due to this requiring an added property in StyleAccumulator I chose to
also take this commit to split up styles from computed styles which are
two separate concepts but up until now have been handled the same. A
style may be `width: 100%` while a computed style would be `width:
320px`.

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/DialogFragmentDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/DialogFragmentDescriptor.java
@@ -20,6 +20,7 @@ import com.facebook.stetho.common.android.FragmentCompat;
 import com.facebook.stetho.inspector.elements.AbstractChainedDescriptor;
 import com.facebook.stetho.inspector.elements.AttributeAccumulator;
 import com.facebook.stetho.inspector.elements.ChainedDescriptor;
+import com.facebook.stetho.inspector.elements.ComputedStyleAccumulator;
 import com.facebook.stetho.inspector.elements.Descriptor;
 import com.facebook.stetho.inspector.elements.DescriptorMap;
 import com.facebook.stetho.inspector.elements.NodeType;
@@ -133,6 +134,6 @@ final class DialogFragmentDescriptor
   }
 
   @Override
-  public void getAccessibilityStyles(Object element, StyleAccumulator accumulator) {
+  public void getComputedStyles(Object element, ComputedStyleAccumulator styles) {
   }
 }
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
@@ -20,6 +20,7 @@ import com.facebook.stetho.common.StringUtil;
 import com.facebook.stetho.common.android.ResourcesUtil;
 import com.facebook.stetho.inspector.elements.AbstractChainedDescriptor;
 import com.facebook.stetho.inspector.elements.AttributeAccumulator;
+import com.facebook.stetho.inspector.elements.ComputedStyleAccumulator;
 import com.facebook.stetho.inspector.elements.StyleAccumulator;
 import com.facebook.stetho.inspector.helper.IntegerFormatter;
 
@@ -42,6 +43,8 @@ final class ViewDescriptor extends AbstractChainedDescriptor<View>
   private static final String ID_NAME = "id";
   private static final String NONE_VALUE = "(none)";
   private static final String NONE_MAPPING = "<no mapping>";
+  private static final String VIEW_SELECTOR_NAME = "<this_view>";
+  private static final String ACCESSIBILITY_SELECTOR_NAME = "Accessibility Properties";
 
   private final MethodInvoker mMethodInvoker;
 
@@ -163,12 +166,12 @@ final class ViewDescriptor extends AbstractChainedDescriptor<View>
 
   @Override
   protected void onGetStyles(View element, StyleAccumulator styles) {
-
     List<ViewCSSProperty> properties = getViewProperties();
     for (int i = 0, size = properties.size(); i < size; i++) {
       ViewCSSProperty property = properties.get(i);
       try {
         getStyleFromValue(
+            VIEW_SELECTOR_NAME,
             element,
             property.getCSSName(),
             property.getValue(element),
@@ -183,15 +186,13 @@ final class ViewDescriptor extends AbstractChainedDescriptor<View>
         }
       }
     }
-  }
 
-  @Override
-  protected void onGetAccessibilityStyles(View element, StyleAccumulator styles) {
     AccessibilityNodeInfoCompat nodeInfo = AccessibilityNodeInfoCompat.obtain();
     ViewCompat.onInitializeAccessibilityNodeInfo(element, nodeInfo);
 
     boolean ignored = AccessibilityNodeInfoWrapper.getIgnored(nodeInfo, element);
     getStyleFromValue(
+        ACCESSIBILITY_SELECTOR_NAME,
         element,
         "ignored",
         ignored,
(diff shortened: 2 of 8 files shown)


stetho/src/main/java/com/facebook/stetho/inspector/elements/android/DialogFragmentDescriptor.java
```diff
@@ -24,6 +24,7 @@
 import com.facebook.stetho.inspector.elements.DescriptorMap;
 import com.facebook.stetho.inspector.elements.NodeType;
 import com.facebook.stetho.inspector.elements.StyleAccumulator;
+import com.facebook.stetho.inspector.elements.ComputedStyleAccumulator;
```
rickbrew: nit: to keep things nice and alphabetical, this should move up a few lines

(and there's a few more of these in the diff, I'll just leave this 1 comment for them)

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -28,6 +29,8 @@
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.HashMap;
```
rickbrew: nit: alphabetical ordering of this too :)

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -96,30 +96,7 @@ public JsonRpcResult getMatchedStylesForNode(JsonRpcPeer peer, JSONObject params
...
-
-    rule.selectorList.selectors = ListUtil.newImmutableList(selector);
-
-    rule.style = new CSSStyle();
-    rule.style.cssProperties = new ArrayList<>();
-
-    match.rule = rule;
-
-    rule.style.shorthandEntries = Collections.emptyList();
+    final Map<String, List<CSSProperty>> selectors = new HashMap<>();
```
rickbrew: using a `HashMap` for temporary storage like this is actually pretty heavy weight ... is there an alternative? how many keys to we expect to have here? can we reuse/recycle, especially since the strings we put in here will probably be repeated over and over?

for instance: maybe we create and cache 1 `HashMap` in an `mSelectorsCache` field. during this method, we still use it as a way to keep track of the `ArrayList` for each selector. then, we `put(selector, null)` at the end instead of either throwing away the `HashMap` or `clear()`ing it. That way we don't constantly churn on the `HashMap.Entry` allocations

I think it's reasonable to do this in a follow-up PR as well, it doesn't have to block us right now

rickbrew: and, we may as well type the `Map`'s `V`alue as `ArrayList` instead of `List` -- doing this can sometimes enable certain static optimizations

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -136,57 +113,46 @@ public void run() {
...
                 if (!isDefault) {
-                  CSSProperty property = new CSSProperty();
-                  property.name = name;
-                  property.value = value;
-
-                  match.rule.style.cssProperties.add(property);
-                }
-              }
-            });
+                  List<CSSProperty> properties = selectors.get(selector);
```
rickbrew: `ArrayList` instead of `List` plz (see comment above!)

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -136,57 +113,46 @@ public void run() {
...
     });
 
-    result.inherited = Collections.emptyList();
-    result.pseudoElements = Collections.emptyList();
+    final GetMatchedStylesForNodeResult result = new GetMatchedStylesForNodeResult();
+    result.matchedCSSRules = new ArrayList<>();
 
-    return result;
-  }
+    for (String selectorName : selectors.keySet()) {
```
rickbrew: Here's another reason i'd prefer to avoid `HashMap` ... iterating over this requires a temporary `Iterator` allocation. Preferable to avoid, although not dogmatically so.

emilsjolander:
Addressed most comments (still iterating keySet()) and fixed some bugs in my implementation which I somehow did not catch last time I was testing this.

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -27,12 +28,15 @@
...
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 public class CSS implements ChromeDevtoolsDomain {
   private final ChromePeerManager mPeerManager;
   private final Document mDocument;
   private final ObjectMapper mObjectMapper;
+  private final Map<String, ArrayList<CSSProperty>> mSelectorCollector = new HashMap<>();
```
rickbrew: please put a javadoc on this explaining how this is used and the restrictions on it -- certainly don't want a future maintainer using it on another thread, or not realizing that there is a "protocol" attached with correctly using it!

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -42,6 +43,9 @@
   private static final String ID_NAME = "id";
   private static final String NONE_VALUE = "(none)";
   private static final String NONE_MAPPING = "<no mapping>";
+  private static final String VIEW_SELECTOR_NAME = "<this_view>";
+  private static final String ACCESSIBILITY_SELECTOR_NAME = "Accessibility Properties";
+
```
rickbrew: nit: an extra line of whitespace somehow appeared :)

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -136,57 +112,56 @@ public void run() {
...
 
+    final GetMatchedStylesForNodeResult result = new GetMatchedStylesForNodeResult();
+    result.matchedCSSRules = new ArrayList<>();
     result.inherited = Collections.emptyList();
     result.pseudoElements = Collections.emptyList();
 
-    return result;
-  }
+    for (String selectorName : mSelectorCollector.keySet()) {
+      final ArrayList<CSSProperty> properties = mSelectorCollector.get(selectorName);
```
rickbrew: is there any reason we're iterating over the `keySet()` and then retrieving the value via `get()`, versus just iterating over the `entrySet()` and using `Map.Entry#getKey()` and `#getValue()` ? they're functionally equivalent, although the latter (iterating over entrySet) avoids an extra hashtable lookup ... which isn't very expensive ... 

iterating the `keySet()` would make perfect sense if we're using an implementation of `Map` that creates the `Map.Entry`s on-demand, such as `ArrayMap` or (iirc) `IdentityHashMap`, but `HashMap` creates them up-front

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -136,57 +119,56 @@ public void run() {
...
     });
 
+    final GetMatchedStylesForNodeResult result = new GetMatchedStylesForNodeResult();
+    result.matchedCSSRules = new ArrayList<>();
     result.inherited = Collections.emptyList();
     result.pseudoElements = Collections.emptyList();
 
-    return result;
-  }
+    for (Map.Entry<String, ArrayList<CSSProperty>> enrty : mSelectorCollector.entrySet()) {
```
rickbrew: typo: `enrty` should be `entry`

rickbrew:
ok looks good. builds locally, and don't really want to wait for Travis-CI

PR #497, opened 2017-02-22
Use empty rect if bounds are null, avoiding NPE

emilsjolander:
The bounds passed to `setHighlightedViewImpl` is `@Nullable` but not the bounds declared in `HighlightOverlays.highlightView`. This commit makes sure `HighlightOverlays.highlightView` is always passed a non-null `Rect`.


rickbrew:
👍

PR #499, opened 2017-02-23
Made URLConnectionInspectorHeaders case insensitive

mariotaku:
URLConnection back by earlier versions of OkHttp (1.5.x) has all lowercased header names, this PR fixed empty Content-Type issue.


jasta:
Good catch, case sensitive comparison is just wrong for headers anyway (and it's not how the OkHttp version of the same method works so even better that they're now consistent)

PR #503, opened 2017-02-28
Add support for editable style properties

emilsjolander:
This commit adds support for editable style properties based on rule
name (previously called selector). Previously properties had to be
implemented as attributes to be able to be edited.

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/common/StringUtil.java
+++ b/stetho/src/main/java/com/facebook/stetho/common/StringUtil.java
@@ -29,4 +29,16 @@ public final class StringUtil {
       return string;
     }
   }
+
+  public static String removeAll(String string, char target) {
+    final int length = string.length();
+    final StringBuilder builder = new StringBuilder(length);
+    for (int i = 0; i < length; ++i) {
+      char c = string.charAt(i);
+      if (c != target) {
+        builder.append(c);
+      }
+    }
+    return builder.toString();
+  }
 }
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/AbstractChainedDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/AbstractChainedDescriptor.java
@@ -140,12 +140,30 @@ public abstract class AbstractChainedDescriptor<E>
   }
 
   @Override
-  public final void getStyles(E element, StyleAccumulator accumulator) {
-    mSuper.getStyles(element, accumulator);
-    onGetStyles(element, accumulator);
+  public final void getStyleRuleNames(E element, StyleRuleNameAccumulator accumulator) {
+    mSuper.getStyleRuleNames(element, accumulator);
+    onGetStyleRuleNames(element, accumulator);
   }
 
-  protected void onGetStyles(E element, StyleAccumulator accumulator) {
+  protected void onGetStyleRuleNames(E element, StyleRuleNameAccumulator accumulator) {
+  }
+
+  @Override
+  public final void getStyles(E element, String ruleName, StyleAccumulator accumulator) {
+    mSuper.getStyles(element, ruleName, accumulator);
+    onGetStyles(element, ruleName, accumulator);
+  }
+
+  protected void onGetStyles(E element, String ruleName, StyleAccumulator accumulator) {
+  }
+
+  @Override
+  public final void setStyle(E element, String ruleName, String name, String value) {
+    mSuper.setStyle(element, ruleName, name, value);
+    onSetStyle(element, ruleName, name, value);
+  }
+
+  protected void onSetStyle(E element, String ruleName, String name, String value) {
   }
 
   @Override
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
@@ -141,10 +141,22 @@ public final class Document extends ThreadBoundProxy {
     mDocumentProvider.setAttributesAsText(element, text);
   }
 
-  public void getElementStyles(Object element, StyleAccumulator styleAccumulator) {
+  public void getElementStyleRuleNames(Object element, StyleRuleNameAccumulator accumulator) {
     NodeDescriptor nodeDescriptor = getNodeDescriptor(element);
 
-    nodeDescriptor.getStyles(element, styleAccumulator);
+    nodeDescriptor.getStyleRuleNames(element, accumulator);
+  }
+
+  public void getElementStyles(Object element, String ruleName, StyleAccumulator accumulator) {
+    NodeDescriptor nodeDescriptor = getNodeDescriptor(element);
+
+    nodeDescriptor.getStyles(element, ruleName, accumulator);
+  }
+
+  public void setElementStyle(Object element, String ruleName, String name, String value) {
+    NodeDescriptor nodeDescriptor = getNodeDescriptor(element);
+
+    nodeDescriptor.setStyle(element, ruleName, name, value);
   }
 
   public void getElementComputedStyles(Object element, ComputedStyleAccumulator styleAccumulator) {
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/NodeDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/NodeDescriptor.java
@@ -34,7 +34,11 @@ public interface NodeDescriptor<E> extends ThreadBound {
 
   void setAttributesAsText(E element, String text);
 
-  void getStyles(E element, StyleAccumulator accumulator);
+  void getStyleRuleNames(E element, StyleRuleNameAccumulator accumulator);
+
+  void getStyles(E element, String ruleName, StyleAccumulator accumulator);
+
+  void setStyle(E element, String ruleName, String name, String value);
 
   void getComputedStyles(E element, ComputedStyleAccumulator accumulator);
 }
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/ObjectDescriptor.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/ObjectDescriptor.java
@@ -53,7 +53,15 @@ public final class ObjectDescriptor extends Descriptor<Object> {
   }
 
   @Override
-  public void getStyles(Object element, StyleAccumulator accumulator) {
+  public void getStyleRuleNames(Object element, StyleRuleNameAccumulator accumulator) {
+  }
+
+  @Override
+  public void getStyles(Object element, String ruleName, StyleAccumulator accumulator) {
+  }
+
+  @Override
+  public void setStyle(Object element, String ruleName, String name, String value) {
   }
 
   @Override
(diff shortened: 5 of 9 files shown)


stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -348,4 +381,18 @@ public PseudoIdMatches() {
     @JsonProperty
     public List<InheritedStyleEntry> inherited;
   }
+
```
rickbrew: nit: just 1 line of whitespace between class entities

stetho/src/main/java/com/facebook/stetho/inspector/elements/ObjectDescriptor.java
```diff
@@ -53,7 +53,18 @@ public void setAttributesAsText(Object element, String text) {
   }
 
   @Override
-  public void getStyles(Object element, StyleAccumulator accumulator) {
+  public void getStyleRules(Object element, StyleRuleAccumulator accumulator) {
+
```
rickbrew: nit: to match the style of the rest of the code, empty methods should have exactly 1 newline after the opening brace

stetho/src/main/java/com/facebook/stetho/inspector/elements/Document.java
```diff
@@ -141,10 +141,22 @@ public void setAttributesAsText(Object element, String text) {
...
+  public void getElementStyles(Object element, String ruleName, StyleAccumulator accumulator) {
+    NodeDescriptor nodeDescriptor = getNodeDescriptor(element);
+
+    nodeDescriptor.getStyles(element, ruleName, accumulator);
+  }
+
+  public void setElementStyle(Object element, String ruleName, String name, String value) {
+    NodeDescriptor nodeDescriptor = getNodeDescriptor(element);
+
+    nodeDescriptor.setStyle(element, ruleName, name , value);
```
rickbrew: nit: errant space after `name` and before the comma

stetho/src/main/java/com/facebook/stetho/inspector/elements/NodeDescriptor.java
```diff
@@ -34,7 +34,11 @@
 
   void setAttributesAsText(E element, String text);
 
-  void getStyles(E element, StyleAccumulator accumulator);
+  void getStyleRules(E element, StyleRuleAccumulator accumulator);
+
+  void getStyles(E element, String ruleName, StyleAccumulator accumulator);
```
rickbrew: I feel like we can improve the naming of the methods and/or interfaces to better highlight the relationship between them. I have a few ideas but I'm not convinced of them:

e.g. renaming `getStyles` to `getStyleRuleStyles` is a bit wordy and weird

renaming `getStyles` to `getRuleStyles` seems pretty good, but a dyslexic nightmare next to `getStyleRules`

renaming `getStyles` to `getStylesForRule` avoids the dyslexia but is still a bit unsatisfying

We could rename `StyleRuleAccumulator` to be `StyleRuleNameAccumulator`, then rename `getStyleRules` to `getStyleRuleNames`, and then leave `getStyles` alone. I think I like this one the best.

Thoughts?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -348,4 +381,18 @@ public PseudoIdMatches() {
...
+
+  private static class SetPropertyTextRequest implements JsonRpcResult {
+      @JsonProperty(required = true)
```
rickbrew: nit: the indentation of this class is super weird ... 4 spaces indent, and then 2 spaces un-indent at the end? pretty sure this should be 2 and 2

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -348,4 +381,18 @@ public PseudoIdMatches() {
...
+      public String styleSheetId;
+
+      @JsonProperty(required = true)
+      public String text;
+    }
+
+  private static class SetPropertyTextResult implements JsonRpcResult {
+      @JsonProperty(required = true)
```
rickbrew: nit: should use 2 space indent

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -98,75 +92,114 @@ public void store(String name, String value) {
...
+            final CSSRule rule = new CSSRule();
+            rule.origin = Origin.REGULAR;
+            rule.selectorList = new SelectorList();
+            rule.selectorList.selectors = ListUtil.newImmutableList(selector);
+            rule.style = new CSSStyle();
+            rule.style.cssProperties = properties;
+            rule.style.shorthandEntries = Collections.emptyList();
+
+            if (editable) {
+              rule.style.styleSheetId = String.format("%d.%s", request.nodeId, selector.value);
```
rickbrew: in order to save an int->Object boxing allocation, which will relieve some GC pressure, prefer this pattern:

`rule.style.styleSheetId = String.format("%s.%s", Integer.toString(request.nodeId), selector.value);`

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -167,93 +168,98 @@ public View getViewAndBoundsForHighlighting(View element, Rect bounds) {
...
+
+  @Override
+  protected void onGetStyles(View element, String ruleName, StyleAccumulator accumulator) {
+    if (VIEW_STYLE_RULE_NAME.equals(ruleName)) {
+      List<ViewCSSProperty> properties = getViewProperties();
+      for (int i = 0, size = properties.size(); i < size; i++) {
+        ViewCSSProperty property = properties.get(i);
+        try {
+          getStyleFromValue(
+                  element,
```
rickbrew: nit: I'm seeing this all over the diff: why 8-space continuation indents? should be 4 spaces

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -98,75 +92,114 @@ public void store(String name, String value) {
...
-      match.matchingSelectors = ListUtil.newImmutableList(0);
+    final String[] parts = request.styleSheetId.split("\\.", 2);
+    final int nodeId = Integer.parseInt(parts[0]);
+    final String ruleName = parts[1];
 
-      final Selector selector = new Selector();
-      selector.value = entry.getKey();
+    final String[] keyValue = request.text.split(":", 2);
+    final String key = keyValue[0].trim();
+    final String value = keyValue[1].replace(";", "").trim();
```
rickbrew: Every time you call `String#replace(String, String)`, it compiles a regular expression via `Pattern`. This can be quite expensive! 

For this super simple replacement, let's put a utility method over into `java.com.facebook.stetho.common.StringUtil` :

```
public static String remove(String string, char target) {
  int length = string.length();
  StringBuilder builder = new StringBuilder(length);
  for (int i = 0; i < length; ++i) {
    char c = string.charAt(i);
    if (c != target) {
      builder.append(c);
    }
  }
  return builder.toString();
}
```

(note that `String#split(String, int)` also interprets as a regex, but if you look at the implementation it has a fast path that covers exactly how you're using it here)

stetho/src/main/java/com/facebook/stetho/inspector/elements/AbstractChainedDescriptor.java
```diff
@@ -140,12 +140,30 @@ protected void onSetAttributesAsText(E element, String text) {
...
+    mSuper.getStyles(element, ruleName, accumulator);
+    onGetStyles(element, ruleName, accumulator);
+  }
+
+  protected void onGetStyles(E element, String ruleName, StyleAccumulator accumulator) {
+  }
+
+  @Override
+  public final void setStyle(E element, String ruleName, String name, String value) {
+    mSuper.setStyle(element, ruleName, name, value);
```
emilsjolander: I can definetly see `View` handling some style such as `width` and `height` and `View` subclasses such as `LinearLayout` handling other styles such as `orientation`.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
emilsjolander: My editor was set up wrong :/

stetho/src/main/java/com/facebook/stetho/inspector/elements/NodeDescriptor.java
emilsjolander: I agree. Implemented last recommendation.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ViewDescriptor.java
```diff
@@ -167,93 +168,98 @@ public View getViewAndBoundsForHighlighting(View element, Rect bounds) {
...
+      }
+
+      nodeInfo.recycle();
     }
+  }
 
-    nodeInfo.recycle();
+  @Override
+  protected void onSetStyle(View element, String ruleName, String name, String value) {
+    super.onSetStyle(element, ruleName, name, value);
```
rickbrew: this method can be removed?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -98,75 +92,117 @@ public void store(String name, String value) {
...
+            rule.origin = Origin.REGULAR;
+            rule.selectorList = new SelectorList();
+            rule.selectorList.selectors = ListUtil.newImmutableList(selector);
+            rule.style = new CSSStyle();
+            rule.style.cssProperties = properties;
+            rule.style.shorthandEntries = Collections.emptyList();
+
+            if (editable) {
+              rule.style.styleSheetId = String.format(
+                  "%d.%s",
```
rickbrew: does `%d` work with a `String`? should this be `"%s.%s"` ?

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/CSS.java
```diff
@@ -98,75 +92,117 @@ public void store(String name, String value) {
...
 
-    for (Map.Entry<String, ArrayList<CSSProperty>> entry : mSelectorCollector.entrySet()) {
-      final ArrayList<CSSProperty> properties = entry.getValue();
-      if (properties.isEmpty()) {
-        continue;
-      }
+  @ChromeDevtoolsMethod
+  public SetPropertyTextResult setPropertyText(JsonRpcPeer peer, JSONObject params) {
+    final SetPropertyTextRequest request =  mObjectMapper.convertValue(
+            params,
```
rickbrew: nit on the 8 space indentation

stetho/src/main/java/com/facebook/stetho/common/StringUtil.java
```diff
@@ -29,4 +29,16 @@ public static String removePrefix(String string, String prefix) {
       return string;
     }
   }
+
+  public static String remove(String string, char target) {
```
rickbrew: Let's rename this to `removeAll`. My suggestion to call it `remove` was almost right, but it implies that it behaves like a regular `Collection#remove` call which would only remove the 1st occurrence that is found. (I would've put up my own PR to change this, but I found some other things in this PR to fix up, so we can just roll it all together)

rickbrew:
Ok everything looks good to me, merging

PR #510, opened 2017-03-08
Some fixes to recent hit test improvements

emilsjolander:
1. Tell chrome to inspect the clicked element, not the view which is
highlighted.

2. Find child to return back to front to respect z ordering.

3. Disregard hidden views and views which are not focusable.


rickbrew:
Alright, let's merge this 👍

PR #511, opened 2017-03-10
Fixed unexpected dependency on support library for Accessibility properties

blavalla:
If you used Stetho in an app without the Android Support library as a dependency, the Accessibility Inspection feature would cause Stetho to crash when you tried to inspect an element.

This makes the dependency on the support library optional, and if it's not included in your app the accessibility inspection feature is simply disabled.

To do this, I had to go through all places we were accessibility AccessibilityNodeInfoCompat and make sure they were only used behind a check for the existence of this class.  To keep this check to be in as few places as possible, I moved all AccessibilityNodeInfo creation and recycling into the AccessibilityNodeInfoWrapper class, which are the bulk of the changes here.

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AccessibilityNodeInfoWrapper.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AccessibilityNodeInfoWrapper.java
@@ -21,15 +21,26 @@ import android.widget.EditText;
 
 import com.facebook.stetho.common.android.AccessibilityUtil;
 
-import java.util.ArrayList;
-import java.util.List;
-
 public final class AccessibilityNodeInfoWrapper {
 
   public AccessibilityNodeInfoWrapper() {
   }
 
-  public static boolean getIgnored(AccessibilityNodeInfoCompat node, View view) {
+  public static AccessibilityNodeInfoCompat createNodeInfoFromView(View view) {
+    AccessibilityNodeInfoCompat nodeInfo = AccessibilityNodeInfoCompat.obtain();
+    ViewCompat.onInitializeAccessibilityNodeInfo(view, nodeInfo);
+    return nodeInfo;
+  }
+
+  public static boolean getIsAccessibilityFocused(View view) {
+    AccessibilityNodeInfoCompat node = createNodeInfoFromView(view);
+    boolean isAccessibilityFocused = node.isAccessibilityFocused();
+    node.recycle();
+
+    return isAccessibilityFocused;
+  }
+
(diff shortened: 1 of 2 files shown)


blavalla:
Just a note, the way github is showing the changes here makes this look way more complex than it is.  For the most part, I simply wrapped each of the AccessibilityNodeInfoWrapper methods in a try/finally block, so that I could easily recycle the node when they were complete.

jasta:
lgtm, although this does seem to be missing code to use AccessibilityNodeInfo directly as well.  The way Stetho interacts with the support library is counter-intuitive: even if the support library contains code to interop with either the framework version of the library or some compatibility fallback it means that we can only do that if the support layer is in the classpath for the app.

For instance, consider an app that doesn't use the support lib because they run natively on newer versions of Android.  They would still have AccessibilityNodeInfo information that is useful to us, but AccessibilityNodeInfoCompat will not be in the classpath.  See an example of this in how we think about Fragments:

https://github.com/facebook/stetho/blob/36aa5bd356d9cf5893b9424b06a83dda9ec5e44f/stetho/src/main/java/com/facebook/stetho/common/android/FragmentCompat.java
https://github.com/facebook/stetho/blob/36aa5bd356d9cf5893b9424b06a83dda9ec5e44f/stetho/src/main/java/com/facebook/stetho/common/android/FragmentCompatFramework.java
https://github.com/facebook/stetho/blob/36aa5bd356d9cf5893b9424b06a83dda9ec5e44f/stetho/src/main/java/com/facebook/stetho/common/android/FragmentCompatSupportLib.java

jasta:
FWIW, this code isn't wrong per se so we could merge it as-is, but it does leave folks not using the compat lib in the dust.  Which, admittedly, is not many users, but it would be very confusing why using the support lib vs not would cause that issue in Stetho :(

blavalla:
Ah good point.  Yeah I can create our own compat class like we did for Fragments, that uses either the native lib if they are on a high enough version of Android or the support lib if it's included.  I believe most of the AccessibilityNodeInfo stuff we're doing here works in API 14 with a few pieces requiring 16.

If a user is on an older version of Android *and* they don't include the support lib, than this will still  fall back to simply being disabled.

jasta:
@blavalla yeah that sounds like the right behaviour.  Let me know if you'd like to merge as-is and follow-up separately, otherwise I'll wait for the next commits.

blavalla:
@jasta lets merge this fix in first, so that at least we stop crashing apps that don't have this library.  I'll start working on a more robust fix later this week and will submit a separate pull request for it.

PR #512, opened 2017-03-10
Fix stdin issue when the stream is read to exhaustion.

### LINKED ISSUE #509: dumpapp not accepting any stdin data
## Expected:

Execute `./dumpapp -p packageName pluginName < someLocalFile.txt` on computer, and `DumperPlugin` have contents in someLocalFile.txt to read.

## Actual:

stdin not receiving any data, attempt to `read()` throws EOFExcepion, on computer side I got
````
Traceback (most recent call last):
  File "./dumpapp", line 84, in <module>
    main()
  File "./dumpapp", line 47, in main
    read_frames(sock)
  File "./dumpapp", line 73, in read_frames
    sock.send(b'-' + struct.pack('!L', -1))
struct.error: argument out of range
````

jasta:
This PR corrects a larger signedness mismatch between dumpapp client and
server.  The dumpapp client (python) used unsigned integers where the
server (Android/Java) used signed integers due to a limitation in
DataInput/OutputStream not supporting unsigned 32 and 64-bit types
(technically this could be done with readLong & 0xffffffffff though).
The critical fix comes in when sys.stdin.buffer.read reaches EOF whereby
the client tried to pack -1 in an unsigned int which is not obviously
not possible.  Since the server was always expecting this to be signed, simply
correcting the client causes the problem to go away.

Closes #509

### DIFF
--- a/scripts/dumpapp
+++ b/scripts/dumpapp
@@ -33,9 +33,9 @@ def main():
     sock = stetho_open(device, process, port)
 
     # Send dumpapp hello (DUMP + version=1)
-    sock.send(b'DUMP' + struct.pack('!L', 1))
+    sock.send(b'DUMP' + struct.pack('!l', 1))
 
-    enter_frame = b'!' + struct.pack('!L', len(args))
+    enter_frame = b'!' + struct.pack('!l', len(args))
     for arg in args:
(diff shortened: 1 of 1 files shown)


jasta:
Self-approving due to the reporter confirming the fix

PR #516, opened 2017-03-13
Add analytics capabilities

emilsjolander:
We would like to track internal usage of stetho to figure out what to
focus on when improving our teams developer experience. This adds an
interface to do so.

I also added a single analytics log for when a method is called from
the inspector. This allows us to track which features are being used.

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/AnalyticsLogger.java
+++ b/stetho/src/main/java/com/facebook/stetho/AnalyticsLogger.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho;
+
+import java.util.Map;
+
+import javax.annotation.Nullable;
+
+/**
+ * Use to track usage of stetho. By default stetho is initialized with a null logger and will not
+ * log anything. However if you would like to track usage of stetho you may provide an
+ * implementation of this interface.
+ */
+public interface AnalyticsLogger {
+
+  /**
+   * @param event The name of the event being tracked
+   * @param data Any additional data to attach to this event
+   */
+  void track(String event, @Nullable Map<String, String> data);
+}
--- a/stetho/src/main/java/com/facebook/stetho/Stetho.java
+++ b/stetho/src/main/java/com/facebook/stetho/Stetho.java
@@ -12,7 +12,6 @@ package com.facebook.stetho;
 import android.app.Application;
 import android.content.Context;
 import android.content.SharedPreferences;
-import android.database.sqlite.SQLiteDatabase;
 import android.os.Build;
 import com.facebook.stetho.common.LogUtil;
 import com.facebook.stetho.common.Util;
@@ -26,7 +25,6 @@ import com.facebook.stetho.dumpapp.plugins.HprofDumperPlugin;
 import com.facebook.stetho.dumpapp.plugins.SharedPreferencesDumperPlugin;
 import com.facebook.stetho.inspector.DevtoolsSocketHandler;
 import com.facebook.stetho.inspector.console.RuntimeReplFactory;
-import com.facebook.stetho.inspector.database.DatabaseConnectionProvider;
 import com.facebook.stetho.inspector.database.DatabaseFilesProvider;
 import com.facebook.stetho.inspector.database.DefaultDatabaseConnectionProvider;
 import com.facebook.stetho.inspector.database.DefaultDatabaseFilesProvider;
@@ -81,6 +79,7 @@ import java.util.Set;
  * the {@code stetho-sample} for more information.
  */
 public class Stetho {
+
   private Stetho() {
   }
 
@@ -91,8 +90,14 @@ public class Stetho {
    * <p />
    * For simple use cases, consider {@link #initializeWithDefaults(Context)}.
    */
+  public static InitializerBuilder newInitializerBuilder(
+      Context context,
+      @Nullable AnalyticsLogger analyticsLogger) {
+    return new InitializerBuilder(context, analyticsLogger);
+  }
+
   public static InitializerBuilder newInitializerBuilder(Context context) {
-    return new InitializerBuilder(context);
+    return newInitializerBuilder(context, null);
   }
 
   /**
@@ -100,8 +105,10 @@ public class Stetho {
    * first socket connection is received, allowing this to be safely used for debug builds on
    * even low-end hardware without noticeably affecting performance.
    */
-  public static void initializeWithDefaults(final Context context) {
-    initialize(new Initializer(context) {
+  public static void initializeWithDefaults(
+      final Context context,
+      @Nullable AnalyticsLogger analyticsLogger) {
+    initialize(new Initializer(context, analyticsLogger) {
       @Override
       protected Iterable<DumperPlugin> getDumperPlugins() {
         return new DefaultDumperPluginsBuilder(context).finish();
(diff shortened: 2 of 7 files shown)


stetho/src/main/java/com/facebook/stetho/Stetho.java
```diff
@@ -81,9 +80,31 @@
  * the {@code stetho-sample} for more information.
  */
 public class Stetho {
+  private static final AnalyticsLogger sLocalAnalyticsLogger = new AnalyticsLogger() {
```
jasta: Hmm, prefer that the these implementation details be moved out of such a highly visible class.  Likewise, usage of statics here is counter to the normal way of configuring the `Stetho` instance.  Prefer instead to add a method to `Stetho.InitializerBuilder` or a more focused one to `InspectorModulesProvider` and/or `DumperPluginsProvider`.

stetho/src/main/java/com/facebook/stetho/inspector/MethodDispatcher.java
```diff
@@ -63,6 +64,11 @@ public JSONObject dispatch(JsonRpcPeer peer, String methodName, @Nullable JSONOb
           "Not implemented: " + methodName,
           null /* data */));
     }
+
+    Stetho.getAnalyticsLogger().track(
+        "inspector_method_called",
+        Collections.singletonMap("name", methodName));
```
jasta: `Map` (or more accurately `HashMap`) is a very slow class to reach for in Java.  It scales well for large values of n but has incredibly high fixed overhead so that a large number of instances where n is small can have a staggeringly negative effect on overall memory usage and app performance.  Google offered an alternative, `SimpleArrayMap` which uses far less memory and trades the `O(1)` lookup time for `O(log n)` (which is faster where n<5) but in your case since you actually just want to iterate a key/value pair you should reach for a simple key/value list.

I'd recommend either defining your own simple class for this, with syntax similar to how you're using `Collections.singletonMap`:

```java
analyticsLogger.track(
    "inspector_method_called", 
    KeyValues.of("name", methodName));
```

or just taking advantage of the fact that a String[] with an even number of items is essentially a key/value list of alternating keys and values:

```java
analyticsLogger.track(
    "inspector_method_called",
    "name", methodName,
    "someOtherThing", bla,
    ...);
```

I'd be happy to discuss specifics with you in FB4A and Messenger about just how much RAM we use in map overhead :)

stetho/src/main/java/com/facebook/stetho/inspector/MethodDispatcher.java
```diff
@@ -63,6 +64,11 @@ public JSONObject dispatch(JsonRpcPeer peer, String methodName, @Nullable JSONOb
           "Not implemented: " + methodName,
           null /* data */));
     }
+
+    Stetho.getAnalyticsLogger().track(
```
jasta: I'm not a fan of how this method allocates even when there is no analytics logger present.  The default strategy you went with here is strange given that it just sends the data to logcat but this would be seen as useless noise to most of our users, even internally when there is no real analytics logger hooked up.  For these reasons, prefer an if-guard to be explicit and optimal:

```java
if (Stetho.hasAnalyticsLogger()) {
  Stetho.getAnalyticsLogger().track(...);
}
```

However I'd hold off on implementing this suggestion until we settle on the API as I am generally not a fan of having this as a global singleton.

stetho/src/main/java/com/facebook/stetho/inspector/MethodDispatcher.java
```diff
@@ -63,6 +64,11 @@ public JSONObject dispatch(JsonRpcPeer peer, String methodName, @Nullable JSONOb
...
+        "inspector_method_called",
```
jasta: This I fear may expose more of an implementation detail in Stetho than actual feature usage.  For instance, Chrome is responsible for invoking these methods and it often does so eagerly.  For instance, on connect, Chrome "enables" every feature of the entire system with Stetho instead of waiting for the user to click around to various tabs and actually use them.  There are usually subtle clues after that that certain methods are responsible for usage but it could be hard to actually develop a heuristic to convert that to meaningful apples to apples usage numbers.

Likewise, this also suffers from not exposing the methods that Stetho invokes against Chrome (this tracks incoming method calls only).  However if you were to add that support it would highlight the problem even further as for example the vast majority of invokes Stetho does to Chrome would be around view tree diffing even if the tab isn't used at all.

stetho/src/main/java/com/facebook/stetho/AnalyticsLogger.java
```diff
@@ -0,0 +1,16 @@
...
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.stetho;
+
+import java.util.Map;
+
+public interface AnalyticsLogger {
```
jasta: javadocs please :)

stetho/src/main/java/com/facebook/stetho/inspector/MethodDispatcher.java
emilsjolander: Known issue. I could not find a great way to track "real" usage. This was the best proxy I could find.

stetho/src/main/java/com/facebook/stetho/inspector/MethodDispatcher.java
emilsjolander: This is not a HashMap. It is a SingletonMap. Looking at the source of java.util.Collections in Android shows what I would like to think is a near optimal implementation

stetho/src/main/java/com/facebook/stetho/Stetho.java
```diff
@@ -91,17 +90,25 @@ private Stetho() {
...
   /**
    * Start the listening server.  Most of the heavy lifting initialization is deferred until the
    * first socket connection is received, allowing this to be safely used for debug builds on
    * even low-end hardware without noticeably affecting performance.
    */
-  public static void initializeWithDefaults(final Context context) {
-    initialize(new Initializer(context) {
+  public static void initializeWithDefaults(
+      final Context context,
+      @Nullable AnalyticsLogger analyticsLogger) {
```
jasta: You shouldn't add the option here.  This conflicts with the builder strategy and is a Java antipattern which would cause all options to be entered into a giant matrix of overload possibilities (if you're curious I can show you many, many examples of this problem *grin*).  Also, this doesn't even semantically make sense as it is no longer initializing with defaults if one option is provided as a non default.

Instead, look at the InitializerBuilder itself and maybe also the webkit and dumpapp specific builders (both of which seem more appropriate as the analytics API could be customized based on which facility is being used.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/AndroidDocumentProviderFactory.java
```diff
@@ -13,6 +13,7 @@
 
 import android.os.Handler;
 import android.os.Looper;
+
```
jasta: A few spurious edits should be cleaned up...

stetho/src/main/java/com/facebook/stetho/inspector/MethodDispatcher.java
```diff
@@ -63,6 +67,13 @@ public JSONObject dispatch(JsonRpcPeer peer, String methodName, @Nullable JSONOb
           "Not implemented: " + methodName,
           null /* data */));
     }
+
+    if (mAnalyticsLogger != null) {
+      mAnalyticsLogger.track(
+          "inspector_method_called",
```
jasta: Again I'm not convinced this strategy will yield the result you want.  Can you please try to model the kind of data this will produce and explain how it will be used to evaluate usage?  To me, I think it'll just be a big awkward blob of data that will mislead as much as it informs (for instance, it would be easy to assume Network.* methods being invoked correspond to the network tab being used but they don't)

jasta:
@emilsjolander I'm assuming this is no longer valid.

PR #519, opened 2017-03-15
Expand scraper.js support for events.

jasta:
This PR adds events (instead of just commands) support for scraper.js
and fixes a crash when a type definition has no fields.


longinoa:
lgtm

PR #521, opened 2017-03-28
Allow hitting empty viewgroups

emilsjolander:
This was implemented previously to get around an issue with overlays.
This was wrong and should be implemented by making overlays
non-focusable instead.


rickbrew:
Looks good 👍

PR #522, opened 2017-03-30
Fix crash when stylesheet text is empty

emilsjolander:
When deleting the value of a style property chrome sends back an empty string instead of a key with an empty value.


rickbrew:
Looks good! 👍

PR #529, opened 2017-04-20
Fix the missing notification on WebSocket handshake response received

misakuo:
Now the session information can be displayed correctly


misakuo:
@jasta review please~

jasta:
Wow, I'm so sorry about missing this!  Thanks for the fix and we'll be pushing out v1.5.1 soon with a big rollup of dangling fixes like this.

Issue #541, opened 2017-06-30
Stetho with OkHttp 3.8.


naturalwarren:
`Interceptor.Chain.connection()` was [marked nullable](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/Interceptor.java#L38) in OkHttp 3.8.0. 

In `StethoInterceptor` we retrieve a connection and don't check for `null`. This causes Stetho to crash when calling [`OkHttpInspectorResponse.connectionId()`](https://github.com/facebook/stetho/blob/master/stetho-okhttp3/src/main/java/com/facebook/stetho/okhttp3/StethoInterceptor.java#L226) when a connection is not present. Stacktrace below:

```
java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.Object.hashCode()' on a null object reference
    at com.facebook.stetho.okhttp3.StethoInterceptor$OkHttpInspectorResponse.connectionId(StethoInterceptor.java:226)
    at com.facebook.stetho.inspector.network.NetworkEventReporterImpl.responseHeadersReceived(NetworkEventReporterImpl.java:145)
    at com.facebook.stetho.okhttp3.StethoInterceptor.intercept(StethoInterceptor.java:70)
    at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:101)
    at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:76)
```

Would you be interested in a PR that null checks `connection` and returns 0 for `connectionId` if a connection is not present? 

This is the only issue I've countered with Stetho 1.5 & OkHttp 3.8 thus far.

naturalwarren:
@jasta Could you please take a peak at this?

naturalwarren:
@rickbrew Trying for Rick :)

naturalwarren:
Fixed in #544.

Hazer:
@naturalwarren When this will be in a maven release?

naturalwarren:
When the maintainers decide to cut a release. @jasta or @rickbrew may know.

PR #544, opened 2017-07-05
Null check Interceptor.Chain.connection().

naturalwarren:
`Interceptor.Chain.connection()` was [marked nullable](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/Interceptor.java#L38) in OkHttp 3.8.0. 

In `StethoInterceptor` we retrieve a connection and don't check for `null`. This causes Stetho to crash when calling [`OkHttpInspectorResponse.connectionId()`](https://github.com/facebook/stetho/blob/master/stetho-okhttp3/src/main/java/com/facebook/stetho/okhttp3/StethoInterceptor.java#L226) when a connection is not present. 

Addresses #541

### DIFF
--- a/stetho-okhttp3/src/main/java/com/facebook/stetho/okhttp3/StethoInterceptor.java
+++ b/stetho-okhttp3/src/main/java/com/facebook/stetho/okhttp3/StethoInterceptor.java
@@ -182,13 +182,13 @@ public class StethoInterceptor implements Interceptor {
     private final String mRequestId;
     private final Request mRequest;
     private final Response mResponse;
-    private final Connection mConnection;
+    @Nullable private final Connection mConnection;
 
     public OkHttpInspectorResponse(
         String requestId,
         Request request,
         Response response,
-        Connection connection) {
+        @Nullable Connection connection) {
       mRequestId = requestId;
       mRequest = request;
       mResponse = response;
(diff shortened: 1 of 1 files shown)


stetho-okhttp3/src/main/java/com/facebook/stetho/okhttp3/StethoInterceptor.java
```diff
@@ -223,7 +223,7 @@ public boolean connectionReused() {
 
     @Override
     public int connectionId() {
-      return mConnection.hashCode();
+      return mConnection == null ? 0 : mConnection.hashCode();
```
rickbrew: Add `@Nullable` to the `mConnection` field and the `connection` constructor parameter too, please :)

naturalwarren: 68709f1

naturalwarren:
@rickbrew Would appreciate another look.

stetho-okhttp3/src/main/java/com/facebook/stetho/okhttp3/StethoInterceptor.java
```diff
@@ -182,13 +182,13 @@ public String firstHeaderValue(String name) {
     private final String mRequestId;
     private final Request mRequest;
     private final Response mResponse;
-    private final Connection mConnection;
+    @Nullable private final Connection mConnection;
```
rickbrew: `@Nullable` should go after `private` (see ElementInfo.java for an example)

naturalwarren: d0afaad

rickbrew:
looks good 👍

naturalwarren:
Thanks @rickbrew!

Issue #423, opened 2016-09-07
Stetho's OkHttp3 interceptor logs incomplete request


paour:
If an application is using other OkHttp3 interceptors besides Stetho's, the network log shows the state of the request at the point Stetho is located at within the interceptor chain, not the request as it was actually sent over the wire.

I don't have a PR to offer, but conceptually, [this block](https://github.com/facebook/stetho/blob/master/stetho-okhttp3/src/main/java/com/facebook/stetho/okhttp3/StethoInterceptor.java#L53) should use not the `request` object it's getting from the `chain` (it will be superseded by the next interceptor in the chain), but from `response.getRequest()`, which contains the final Request as it bubbles back up the chain of interceptors after the request has been submitted.

The fact that we would need to wait for the response to come back (or not) before calling `NetworkEventReporterImpl#requestWillBeSent`, so the timing might be off.

longinoa:
We ran into a similar problem with this internally, and looked at a few different solutions... 
1) provide a separate interface to add interceptors.
2) use something similar to what you propose above
3) attach the `StethoInterceptor` as the last interceptor by convention. 

Ultimately we ended up going with 3. The basic assumption is that most consumers will have some sort of utility library that provides them with an `OkHttpClient` they will just start using and dont tend to add more interceptors as this is more of an application level concept

paour:
Option 3 above would need to be documented. The [Quick-start documentation](http://facebook.github.io/stetho/#enable-network-inspection) is in need of changes, in any case: the first example (`client.networkInterceptors().add(new StethoInterceptor());`) is incorrect, since `networkInterceptors` is an immutable list…

Could you expand on option 1? I'm not sure I understand…

longinoa:
So basically 1 would be stetho providing a builder interface for okhttp rather than requiring you to call `addNetworkInterceptor` yourself. This is really not a good direction for stetho as we support multiple versions of okhttp and would have to rev the builder each time a new version came out. It looks like #426 is a good modification to the readme

longinoa:
@paour  does #426 close this for you?

paour:
Yes, thank you.

Issue #420, opened 2016-08-29
Getting a NPE crash from stetho internal library


Gopinathp:
Hi Team,

I am getting a NPE from stetho internal code. Here is the stacktrace.

```
com.facebook.stetho.okhttp3.StethoInterceptor$OkHttpInspectorResponse.connectionId (StethoInterceptor.java:229)
com.facebook.stetho.inspector.network.NetworkEventReporterImpl.responseHeadersReceived (NetworkEventReporterImpl.java:143)
com.facebook.stetho.okhttp3.StethoInterceptor.intercept (StethoInterceptor.java:73)
okhttp3.internal.http.RealInterceptorChain.proceed (RealInterceptorChain.java:92)
okhttp3.internal.http.RealInterceptorChain.proceed (RealInterceptorChain.java:67)
okhttp3.logging.HttpLoggingInterceptor.intercept (HttpLoggingInterceptor.java:212)
okhttp3.internal.http.RealInterceptorChain.proceed (RealInterceptorChain.java:92)
okhttp3.internal.http.RealInterceptorChain.proceed (RealInterceptorChain.java:67)
com.wiznsystems.android.utils.ApiClient$HostSelectionInterceptor.intercept (ApiClient.java:133)
okhttp3.internal.http.RealInterceptorChain.proceed (RealInterceptorChain.java:92)
okhttp3.internal.http.RealInterceptorChain.proceed (RealInterceptorChain.java:67)
okhttp3.RealCall.getResponseWithInterceptorChain (RealCall.java:170)
okhttp3.RealCall.execute (RealCall.java:60)
retrofit2.OkHttpCall.execute (OkHttpCall.java:174)
retrofit2.ExecutorCallAdapterFactory$ExecutorCallbackCall.execute (ExecutorCallAdapterFactory.java:89)
com.wiznsystems.android.utils.NavDrawerUtils$4.run (NavDrawerUtils.java:330)
```

longinoa:
Are you calling `addNetworkInterceptor` or `addInterceptor`? This is looking at the `Connection` that comes from the `Chain` which should be non null if the request hit the network.

longinoa:
I verified this using a local build. If it does not work for you please re-open

Issue #346, opened 2016-01-27
OkHttp3 Interceptor crash


juandiana:
I just upgraded OkHttp to version 3.0.1 and Stetho to version 1.3.0 (and its corresponding new OkHttp3 network interceptor).

I'm experiencing the following NullPointerException when inspecting HTTP traffic:

```
E/stetho: com.facebook.stetho.inspector.elements.android.DialogFragmentDescriptor.getChildren() emitted a null child at position 0 for element LayoutFragment{3a8b2dd3 #0 id=0x7f0c00dc}
E/AndroidRuntime: FATAL EXCEPTION: OkHttp Dispatcher
         Process: com.artech.androidbasictest.testgxsalto, PID: 29855
         java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.Object.hashCode()' on a null object reference
         at com.facebook.stetho.okhttp3.StethoInterceptor$OkHttpInspectorResponse.connectionId(StethoInterceptor.java:229)
         at com.facebook.stetho.inspector.network.NetworkEventReporterImpl.responseHeadersReceived(NetworkEventReporterImpl.java:143)
         at com.facebook.stetho.okhttp3.StethoInterceptor.intercept(StethoInterceptor.java:73)
         at okhttp3.internal.http.HttpEngine$NetworkInterceptorChain.proceed(HttpEngine.java:681)
         at okhttp3.internal.http.HttpEngine.readResponse(HttpEngine.java:563)
         at okhttp3.RealCall.getResponse(RealCall.java:241)
         at okhttp3.RealCall$ApplicationInterceptorChain.proceed(RealCall.java:198)
         at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:160)
         at okhttp3.RealCall.access$100(RealCall.java:30)
         at okhttp3.RealCall$AsyncCall.execute(RealCall.java:127)
         at okhttp3.internal.NamedRunnable.run(NamedRunnable.java:33)
         at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
         at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
         at java.lang.Thread.run(Thread.java:818)
```

jasta:
This looks like the same issue as with okhttp2, you may have accidentally used a regular interceptor instead of a network interceptor.  I keep forgetting to actually fix this with a proper error message :(

[...]

Issue #267, opened 2015-09-02
NullPointerException in OkHttpInspectorResponse.connectionId()


Turbo87:
I'm getting the following crash once Chrome is connected to the app and the app tries to do an HTTP request:

````java
09-02 06:00:27.044  28152-28197/... E/AndroidRuntime﹕ FATAL EXCEPTION: OkHttp Dispatcher
    Process: ..., PID: 28152
    java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.Object.hashCode()' on a null object reference
            at com.facebook.stetho.okhttp.StethoInterceptor$OkHttpInspectorResponse.connectionId(StethoInterceptor.java:228)
            at com.facebook.stetho.inspector.network.NetworkEventReporterImpl.responseHeadersReceived(NetworkEventReporterImpl.java:144)
            at com.facebook.stetho.okhttp.StethoInterceptor.intercept(StethoInterceptor.java:72)
            at com.squareup.okhttp.Call$ApplicationInterceptorChain.proceed(Call.java:221)
            at com.squareup.okhttp.Call.getResponseWithInterceptorChain(Call.java:195)
            at com.squareup.okhttp.Call.access$100(Call.java:34)
            at com.squareup.okhttp.Call$AsyncCall.execute(Call.java:162)
            at com.squareup.okhttp.internal.NamedRunnable.run(NamedRunnable.java:33)
            at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
            at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
            at java.lang.Thread.run(Thread.java:818)
````

I'm using the following lines in my `build.gradle` file:

```gradle
    debugCompile 'com.facebook.stetho:stetho:1.1.1'
    compile 'com.facebook.stetho:stetho-okhttp:1.1.1'

    compile 'com.squareup.okhttp:okhttp:2.5.0'
```

jasta:
It's possible you accidentally used `okHttpClient.interceptors().add(...)` instead of <tt>okHttp.<b>networkInterceptors()</b>.add(...)</tt>.  An NPE is probably the wrong response out of Stetho, but it can't work all the same because interceptors provide very different kinds of information than the network interceptors.

Turbo87:
That is indeed possible. I was not aware that there was more than one kind of interceptors in OkHttp...

BBlackwo:
[...]

PR #549, opened 2017-08-03
Throw RegistrationException when connection isn't present.

naturalwarren:
@rickbrew I was chatting with @swankjesse about this: Connection is guaranteed to be present in network interceptors and is only and always nullable in vanilla interceptors.

This change causes Stetho to throw a `StethoInterceptor.RegistrationException` when a connection is not present and prompts consumers to check their interceptor registration.

Helps mitigate issues like: #423, #135, #420, #267, #346, #555


naturalwarren:
cc @jasta

naturalwarren:
@rickbrew Could this get a head nod?

jasta:
Commandeered this in favor of #584 as it applied the fix back to okhttp2 which, sadly, Facebook still uses :)

Issue #480, opened 2017-01-08
No inspect button are appear


pedrohenriquerls:
I followed the tutorial [here](https://facebook.github.io/react-native/docs/debugging.html), but I cant inspect.
The device is a emulator and am using the Ubuntu 16.04.
I have the server running and everything looks ok, but the option to inspect never appears.

![screenshot from 2017-01-08 01-55-03](https://cloud.githubusercontent.com/assets/952303/21747161/9909d7d2-d545-11e6-98f6-b125e49da6a8.png)

aandis:
hitting this. I have a physical device and trying on ubuntu 14.04

aandis:
Fixed by adding my created `Application` class to manifest.

jasta:
Can you check that `stetho-sample` works properly (download the sources and run `./gradlew installDebug`, find and run the sample, then look at `chrome://inspect`)?

jasta:
Closing due to #551

PR #551, opened 2017-08-14
Fix issue #480 : Documentation to add MyApplication to AndroidManifest.xml

ozmium:
The MyApplication Java class must be added to the AndroidManifest.xml, otherwise there will be no "Inspect" button that appears in chrome://inspect/#devices

Add this explanation in the documentation of Readme.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,18 @@ public class MyApplication extends Application {
   }
 }
 ```
+Also ensure that your `MyApplication` Java class is registered in your `AndroidManifest.xml` file, otherwise you will not see an "Inspect" button in `chrome://inspect/#devices` :
+
+```xml
+<manifest
+        xmlns:android="http://schemas.android.com/apk/res/android"
+        ...>
+        <application
+                android:name="MyApplication"
+                ...>
+         </application>
+</manifest>                
+```
 
 This brings up most of the default configuration but does not enable some
 additional hooks (most notably, network inspection).  See below for specific


naturalwarren:
Feels a little verbose to me, never seen an Android project tell you to do rudimentary things like this.

ozmium:
Not all Android apps have an Application class if they don't need it. So these apps would have to explicity add it, just to use the stetho library.

Also, there is no obvious explanation of why the "Inspect" button is missing, so it's unclear to library users why this problem would happen, and hence the documentation explains this for users.

alecperkey:
Um I also don't see the Inspect button, but not sure this is why. 
I would prefer some verbosity, since I'm unfamiliar with the internals of android. I'm actually working from react-native init structure which looks like https://github.com/alecperkey/chatty/commit/f1267e1a8927446bf2d6a40e97f6b1f3746f242f#diff-07d4e3cf4609b3deea6aff38092ec1f9R14

Is ```android:name="MainApplication"```even correct? Shouldn't it be rather be:

```android:name=".MainApplication"```

with dot prefix? I don't know I am just not seeing the 'inspect' button also and noticed you don't have the dot while I do. Removing it gave me other errors

```
Running C:\Users\user\AppData\Local\Android\sdk/platform-tools/adb -s 192.168.95.101:5555 reverse tcp:8081 tcp:8081
Starting the app on 192.168.95.101:5555 (C:\Users\user\AppData\Local\Android\sdk/platform-tools/adb -s 192.168.95.101:5555 shell am start -n com.chatty/com.chatty.MainActivity)...
Starting: Intent { cmp=com.chatty/.MainActivity }
Error type 3
Error: Activity class {com.chatty/com.chatty.MainActivity} does not exist.
```

jasta:
lgtm.  It is a bit verbose as @naturalwarren points out but also reads somewhat naturally to me.  Folks that find it redundant will likely auto-skip past it and ignore it :)

Issue #329, opened 2016-01-17
OkHttp inspector hangs on big uploads


evgeny-sureev:
My app uploads photos taken from camera. If upload request appears on log, Chrome hangs completely. I think this happens because inspector tries to dump raw jpeg content to Headers tab.

jasta:
This is due to a very unfortunate limitation in the inspector protocol which appears to require that we serialize request payloads as UTF-8 strings and upload them to the UI.  I know of no way around this other than to perhaps upgrade to a newer version of the protocol (which I haven't evaluated in ~6 months) or to simply detect large payloads and omit them when reporting the request.

I tried for a while to figure out what current Chrome upstream does when uploading large images but for reasons I forget now couldn't seem to repro this issue.  If you're feeling motivated, you could raelly help speed this up by connecting Chrome for Android to Chrome on your PC/laptop (chrome://inspect just like with Stetho, find your Chrome for Android instance and click Inspect), then try to get Chrome for Android to upload a very large payload to a server and capture what the inspector does.  In either case (it does what you want or not) could you take a screenshot for us and we'll see what we can do?

evgeny-sureev:
It seems that Chrome omits large payloads:

![screenshot](https://cloud.githubusercontent.com/assets/7701455/12440402/37a60124-bf4c-11e5-827a-4028b0cc491b.png)

This is from stetho with image displayed as raw content:

![example](https://cloud.githubusercontent.com/assets/7701455/12440473/e259fad0-bf4c-11e5-84fb-25d0a1313f74.jpg)

With small sized images inspector just slow down, so I am able to capture screenshot. When image size exceed 1 Mb, inspector stop responding.

jasta:
I think the right thing to do then would be to detect large payloads and simply omit them from the payload string.  If you're interested in taking a crack at this, it should be a simple matter of applying this heuristic in `NetworkEventReporterImpl.java:readBodyAsString`.  Because this method sees the decompressed body, it should be safe to simply look for binary and/or very large payloads and replace those payloads with something else.

Also note that it's possible to do an even better job filtering this out for `StethoInterceptor` because it can detect a multi-part request body and inspect the content-type of each part individually to determine what it should report.  But if you don't do this carefully you will end up causing the "request size" field to not be computed correctly.

jasta:
[...]

PR #564, opened 2017-09-15
CheckCharset on PostData to prevent hanging or crashing

thebehera:
This should resolve the error around: https://github.com/facebook/stetho/issues/329
Use CharsetDecoder to decode the byte[]. if it reports an error, bail out and report to the chrome inspector that `Charset in POST/PUT is not UTF-8. Data (length:X) cannot be represented as a string.`

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkEventReporterImpl.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkEventReporterImpl.java
@@ -10,21 +10,30 @@
 package com.facebook.stetho.inspector.network;
 
 import android.os.SystemClock;
+
 import com.facebook.stetho.common.Utf8Charset;
 import com.facebook.stetho.inspector.console.CLog;
 import com.facebook.stetho.inspector.protocol.module.Console;
 import com.facebook.stetho.inspector.protocol.module.Network;
 import com.facebook.stetho.inspector.protocol.module.Page;
+
 import org.json.JSONException;
 import org.json.JSONObject;
 
-import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
+import java.nio.charset.CharacterCodingException;
+import java.nio.charset.Charset;
+import java.nio.charset.CharsetDecoder;
+import java.nio.charset.CodingErrorAction;
 import java.util.ArrayList;
 import java.util.concurrent.atomic.AtomicInteger;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 
 /**
  * Implementation of {@link NetworkEventReporter} which allows callers to inform the Stetho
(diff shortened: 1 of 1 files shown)


thebehera:
Showing a NPE, debugging now, and will update the PR

thebehera:
fixed NPE around getting a null body

jasta:
Awesome work, so glad someone tackled this!

stetho/build.gradle
```diff
@@ -19,6 +19,7 @@ dependencies {
     compile 'com.google.code.findbugs:jsr305:2.0.1'
 
     compile 'com.android.support:appcompat-v7:23.0.1' // optional
+    compile 'net.zetetic:android-database-sqlcipher:3.5.7@aar'
```
jasta: You appear to have a number of unrelated changes here that should be removed.  Also an aside, you don't need this SQLCipher thing directly in Stetho as you can just install your own `DatabaseDriver2` implementation now (see `SqliteDatabaseDriver` and work from there as needed).

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkEventReporterImpl.java
```diff
@@ -39,6 +48,8 @@
 
   private static NetworkEventReporter sInstance;
 
+  private static final CharsetDecoder decoder = Charset.forName(Utf8Charset.NAME).newDecoder();
```
jasta: Make this a non-static member variable and initialize it in the constructor.  No need to muck up the `get()` method below.

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkEventReporterImpl.java
```diff
@@ -115,15 +128,29 @@ private static String readBodyAsString(
       InspectorRequest request) {
     try {
       byte[] body = request.body();
-      if (body != null) {
-        return new String(body, Utf8Charset.INSTANCE);
+      if (body == null || body.length == 0) {
+        return "";
```
jasta: It used to return null if body was null before.  Should it still?

stetho/src/main/java/com/facebook/stetho/inspector/network/NetworkEventReporterImpl.java
```diff
@@ -115,15 +128,29 @@ private static String readBodyAsString(
...
+        return charBuffer.toString();
+      } catch (CharacterCodingException e) {
+        String logMessage = "Charset in POST/PUT is not UTF-8. Data (length:"+ body.length
+                +") cannot be represented as a string. ";
+        CLog.writeToConsole(
+                peerManager,
+                Console.MessageLevel.WARNING,
+                Console.MessageSource.NETWORK,
+                 logMessage + e);
+        return logMessage;
```
jasta: Hmm, I like returning the log message here, but it suggests actually that we'd be better off merging the error cases below:

````
try {
 // decode...
} catch (IOException | OutOfMemoryError e) {
  String error = determineErrorMessage(e);
  CLog.writeToConsole("Could not reproduce POST/PUT body: " + error);
  return error;
}
````

Then determineErrorMessage could peak and see if it's a CharsetCodingException and craft your better error message above.

PR #578, opened 2017-11-01
fix payloadData length error

funny9527:
fix payloadData length error when it's length is more than 0xffff


jasta:
Wow good catch.  Took me a minute (actually, several) to even realize why this was broken and how this fix works :)

Issue #595, opened 2018-05-09
Order SharedPrefences by key


mariotaku:
As of 1.5.0, SharedPreferences listed in fresco is not sorted, which makes it super difficult to find a value. Is it possible to sort those by keys?

longinoa:
closing as the PR was merged

PR #596, opened 2018-05-10
Sorts SharedPreferences entries

mariotaku:
This PR solves #595 . Currently all preferences are unsorted and hard to find. This patch will make preferences sort by key alphabetically.


mariotaku:
Please consider merging this PR, thanks!

PR #603, opened 2018-07-06
Upgrade gradle & android plugin

dant3:
I've updated gradle up to 4.4 and android plugin up to 3.1.3 and fixed all corresponding problems.
Some observations: I've tried to move to the `implementation` scope as much as it was possible without breaking stuff for other projects.
Most of dependencies are in `api` scope since it's a safe transition from end-users POV.
In reality perhaps some dependencies declarations should be further reworked to be more granular.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,8 @@ jdk:
 android:
   components:
     - tools
-    - build-tools-25.0.2
+    - build-tools-27.0.3
+    - build-tools-26.1.3
     - android-23
     - platform-tools
     - extra-android-support
--- a/build.gradle
+++ b/build.gradle
@@ -1,14 +1,18 @@
 buildscript {
     repositories {
         jcenter()
+        google()
     }
     dependencies {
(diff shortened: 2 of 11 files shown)


dant3:
I've signed it

dant3:
Seems like update of build-tools in Travis config is required for this to pass.

.travis.yml
```diff
@@ -6,7 +6,8 @@ jdk:
 android:
   components:
     - tools
-    - build-tools-25.0.2
+    - build-tools-27.0.3
+    - build-tools-26.1.3
```
rogerhu: build tools doesn't need to be specified anymore, the latest gradle will figure out what to use

dant3: it will, but it will fail to install it if license is not accepted beforehand, so you either add pre-accepted license or use travis to install it. I preferred travis as it seems simpler.

jasta:
This conflicted with another PR but the effect should now be merged.

PR #604, opened 2018-07-06
Fixes #602 - allow to change console banner

### LINKED ISSUE #602: Custom banner in debugger console
Hi.

Currently there is no way to change the banner inside of the console. I would love to customize that.

I've implemented some additional api exclusively for debugging console via JS bindings. In order to make it easy to use what I want to do is to print some sort of welcome message with help on what is exposed in the console directly and how to use it.
I've tried to just do printing into the console but this is not happening until user does not sends anything to the console. Meanwhile banner is always printed beforehand/

dant3:
As per @longinoa suggestion in #602, I've added additional constructor to Page and made a fallback to existing banner. It's still is a constant `String` so no packaging changes required.

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Page.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Page.java
@@ -29,13 +29,32 @@ import java.util.Iterator;
 import java.util.List;
 
 public class Page implements ChromeDevtoolsDomain {
+  public static final String BANNER = // Note: not using Android resources so we can maintain .jar distribution for now.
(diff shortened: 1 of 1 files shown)


dant3:
I've signed it

stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Page.java
```diff
@@ -29,13 +29,32 @@
 import java.util.List;
 
 public class Page implements ChromeDevtoolsDomain {
+  public static final String BANNER = // Note: not using Android resources so we can maintain .jar distribution for now.
```
jasta: yeah, that sucks but I agree :(

Issue #611, opened 2018-08-14
Runtime.getProperties() ignores all requests with "ownProperties":false


AlexTrotsenko:
When [`Runtime.getProperties()`](https://github.com/AlexTrotsenko/stetho/blame/00e8127a0d5d95610e8c4d20cc7592fbe7de5830/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Runtime.java#L162) is invoked from the Chrome DevTools with `"ownProperties":false` this call is ignored and no real value returned.

It's happens because of the [explicit check](https://github.com/AlexTrotsenko/stetho/commit/43585ad4e0f0e2d6ac4e0ebf2092bfdc059924e2#diff-ded90e9f295fede2805fb40d7906b789R222) for `!request.ownProperties` in Runtime.

Any ideas why this is present there? I am working on implementation of _Debugger_ for [J2V8](https://github.com/eclipsesource/J2V8), but this code prevents from retrieving values for the local variables when debugger is _paused_.

I know, that it's 3 years old code and without tests it's hard to understand why it's needed, but may be by the chance @jasta you might know why was it added there?

jasta:
Unsure if this is still relevant after the merged PRs.  Please re-open if this still needs attention.

PR #612, opened 2018-08-15
Remove private class (CallFunctionOnResponse) from signature of the public method

AlexTrotsenko:
Otherwise class can't be extended and method overridden (Currently only @ChromeDevtoolsMethod methods are detected and method needed to be overridden because look-up is done only in declared methods)

Override of Stetho's default domain classes might be need if issues like following are found: https://github.com/facebook/stetho/issues/611


jasta:
Good catch, thanks!

PR #614, opened 2018-08-17
Added static method for removing objects previously added by .mapObject()

AlexTrotsenko:
I am developing debugger for [J2V8](https://github.com/eclipsesource/J2V8), which uses Stetho for communication over Chrome  DevTools protocol. Besides implementing Debugger domain I need to add local v8 variables to `Runtime` by static `.mapObject()` in order them to be visible in the Chrome Debugger UI.

The problem, that `@ChromeDevtoolsMethod` `.releaseObject()` method is not called by Chrome DevTools and I want to remove objects stored in the session manually in order to avoid memory leaks and possible OOM errors. That's why I would like to have static method to remove them from session. Some work-around are possible already now, but it makes sense for me to have it because `static` _counter-part_ for adding object is exists - `.mapObject()` .

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Runtime.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Runtime.java
@@ -105,6 +105,13 @@ public class Runtime implements ChromeDevtoolsDomain {
     return session;
   }
 
+  /**
+   * Removes objects from peer's session previously added by {@link #mapObject}
+   */
+  public static void releaseObject(JsonRpcPeer peer, Integer id) throws JSONException {
+    getSession(peer).getObjects().removeObjectById(id);
+  }
+
   @ChromeDevtoolsMethod
   public void releaseObject(JsonRpcPeer peer, JSONObject params) throws JSONException {
     String objectId = params.getString("objectId");


jasta:
I'm fine to merge this and I think I agree with your motivation, but one thing to call out: this will be fragile and fairly hard to maintain as a real API into the future.  I'm sympathetic to your use case and want to support you, but I'm not necessarily confident that directly reaching into the Runtime module is the best contract for us to maintain.  Is your integration open sourced somewhere I can see so at least Stetho can track this as a major customer to support and make sure we don't accidentally break?

AlexTrotsenko:
@jasta  sorry for delay with the answer - I have missed your reply somehow. 
The project is available on github: https://github.com/AlexTrotsenko/j2v8-debugger

The exact place where V8 variable is added to Runtime is [following](https://github.com/AlexTrotsenko/j2v8-debugger/blob/0.1.2/j2v8-debugger/src/main/java/com/alexii/j2v8debugger/Debugger.kt#L416)  _(in Kotlin)_.

Concerning you notes about regarding fragility of such API: 
- According to what I see in the doc of the [Session](https://github.com/facebook/stetho/blob/v1.5.0/stetho/src/main/java/com/facebook/stetho/inspector/protocol/module/Runtime.java#L186) you have faced with similar leaks issues in the past for UI elements.
 - If you have something better in mind for not using the `remove()` method - I would be happy to use it as well. 
E.g. I can imaging using concept of _additional_ , but _optional_ grouping of mapped objects-   something like "sub-session" or "tag" etc.
The resulting code might looks like following then: 
```.Java

  public static int mapObject(JsonRpcPeer peer, Object object, @Nullable SubSession subSession) {
    final Session session = getSession(peer);

    if (subSession != null) {
      subSession.registerOnSubSessionEnded(new OnEndedReceiver() {
        @Override
        public void onDisconnect() {
          session.removeSubSession(subSession);
        }
      });
    }

    return session.getObjects(subSession).putObject(object);
  }


  public static int mapObject(JsonRpcPeer peer, Object object) {
    return mapObject(peer, object, null);
  }
```

Then my client code might looks like following:

```.Java
private class V8ToChromeDevToolsBreakHandler(private val currentPeerProvider: () -> JsonRpcPeer?) : BreakHandler {
   val subSession: SubSession = null
   ...
    override fun onBreak(event: DebugHandler.DebugEvent?, state: ExecutionState?, eventData: EventData?, data: V8Object?) {
                        ... 
                       val storedVariablesId = Runtime.mapObject(currentPeerProvider(), knowVariables, subSession)
                        .... 
    }
    ....
    /**
     * Resumes V8 execution. Called from Stetho thread.
     */
    fun resume() {
        subSession.onSubSessionEnded()
        resumeWith(null)
    }
}

```


P.S. Thanks taking time and merging all my pull requests - I will integrate the latest version of Stetho into my project.

PR #620, opened 2018-09-21
support window View Hierarchy (dialog, popupwindow, toast, and so on)

android-notes:
support window View Hierarchy (dialog, popupwindow, toast, and so on).

example:
click MainActivity#about button


android-notes:
@jasta   @rickbrew

android-notes:
view hierarchy
gif

stetho/src/main/java/android/view/WindowManagerGlobal.java
```diff
@@ -0,0 +1,21 @@
...
+ * A simple way to get WindowManagerGlobal
+ * <p>
+ * WindowManagerGlobal is hide,so we can not use it,but we can create a same name class.
+ * <p>
+ * <p>
+ * declare -keep class android.view.WindowManagerGlobal{*;} in your proguard-rules.pro file
+ */
+@Keep
+public class WindowManagerGlobal {
+    public static WindowManagerGlobal getInstance() {
```
jasta: I really prefer reflection here for several reasons, most notably that we're forced to have a checked exception proving that this method may not exist.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ApplicationDescriptor.java
```diff
@@ -61,8 +63,38 @@ protected void onGetChildren(Application element, Accumulator<Object> children)
         children.store(activity);
       }
     }
+    storeWindowIfNeeded(element,children,activities);
```
jasta: style nit: spaces after `,`

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ApplicationDescriptor.java
```diff
@@ -61,8 +63,38 @@ protected void onGetChildren(Application element, Accumulator<Object> children)
...
   }
+  private void storeWindowIfNeeded(Application application, Accumulator<Object> children, List<WeakReference<Activity>> activities){
```
jasta: several style nits: spaces before `{`, before variable names, between functions, etc.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ApplicationDescriptor.java
```diff
@@ -61,8 +63,38 @@ protected void onGetChildren(Application element, Accumulator<Object> children)
...
+    List<View>windows = WindowFetcher.getWindowViews(application);
+    if (windows == null){
+      return;
+    }
+    for (View view : windows) {
+      if (belongToActivity(view, activities)) {
```
jasta: nit: this logic reads better in reverse: `if (!belongToActivity(...)) { children.store(view); }`

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ApplicationDescriptor.java
```diff
@@ -61,8 +63,38 @@ protected void onGetChildren(Application element, Accumulator<Object> children)
...
+        continue;
+      }
+      children.store(view);
+    }
+  }
+
+
 
+  static boolean belongToActivity(View view, List<WeakReference<Activity>> references) {
+    if (references == null) {
```
jasta: the null checking through these edits seems excessive.  can you update the code to use `@Nullable` in all the places you genuinely expect nullness and drop the checks for the others / convert them into `Util.throwIfNull` calls.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/ApplicationDescriptor.java
```diff
@@ -61,8 +63,38 @@ protected void onGetChildren(Application element, Accumulator<Object> children)
...
+      if (belongToActivity(view, activities)) {
+        continue;
+      }
+      children.store(view);
+    }
+  }
+
+
 
+  static boolean belongToActivity(View view, List<WeakReference<Activity>> references) {
```
jasta: naming: this function appears to be `isDecorViewOf(...)`, no?

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/window/WindowFetcher.java
```diff
@@ -0,0 +1,103 @@
...
+ * 5.0.0_r2     WindowManagerGlobal   private final ArrayList<View> mViews
+ *
+ * 6.0.0_r1     WindowManagerGlobal   private final ArrayList<View> mViews
+ *
+ * 7.0.0_r1     WindowManagerGlobal   private final ArrayList<View> mViews
+ *
+ * 8.0.0_r4    WindowManagerGlobal   private final ArrayList<View> mViews
+ */
+
+public class WindowFetcher {
```
jasta: style nit: indent 2 spaces per the rest of the sources

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/window/WindowFetcher.java
```diff
@@ -0,0 +1,103 @@
...
+import com.facebook.stetho.common.Util;
+
+import java.lang.reflect.Field;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+
+/*
+ * 4.0.3_r1    WindowManagerImpl   private View[] mViews;
```
jasta: this documentation is useful, but please also try to provide a bit of context of what's going on here.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/window/WindowFetcher.java
```diff
@@ -0,0 +1,103 @@
...
+			WindowManagerGlobal managerGlobal = WindowManagerGlobal.getInstance();
+			return getWindowViews(managerGlobal);
+
+		} else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
+			return sWindow;
+		}
+		return null;
+	}
+
+	private static List<View> sWindow;
```
jasta: style nit: variables go before functions

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/window/WindowWrapper.java
```diff
@@ -0,0 +1,15 @@
+package com.facebook.stetho.inspector.elements.android.window;
+
+import android.view.View;
+
+public class WindowWrapper {
```
jasta: nit: I think you also mean `WindowDecorViewWrapper` or maybe even better, `WindowDecorViewHolder`

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/window/WindowFetcher.java
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	private static List<View> getWindowViews(final Object obj) {
+		try {
+			Class clz = obj.getClass();
+			final Field field = clz.getDeclaredField("mViews");
+			field.setAccessible(true);
+			return Collections.unmodifiableList(Arrays.asList((View[]) field.get(obj)));
+		} catch (Exception e) {
+			e.printStackTrace();
```
jasta: This is a pretty nasty anti-pattern in Java and I strongly encourage you not to use it here or in other Java projects.  Instead, consider each kind of exception you actually expect, what it means, and what the code is expected to do.  For instance, if you expect that this means that we're just simply on the wrong version of the platform, perhaps we should be testing for that separately.  If you suspect that this means the manufacturer has really screwed something up by deviating from aosp standard, bark appropriately with a meaningful log message to that effect.  Each kind of exception you might get should be handled differently tho -- some of them probably should just throw RuntimeException even...

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/window/WindowFetcher.java
```diff
@@ -0,0 +1,103 @@
...
+			Object outerWM = parentFiled.get(innerWM);
+			parentFiled.setAccessible(false);
+			return outerWM;
+		} catch (Exception e) {
+			e.printStackTrace();
+		}
+		return null;
+	}
+
+	private static List<View> getWindowViews(final Object obj) {
```
jasta: `@Nullable` here and many other places.  Using null without making it clear what null means is also a pretty problematic anti-pattern which makes code hard to read and maintain...

jasta: I think `obj` is better named as `windowManager` here, no?

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/window/WindowFetcher.java
```diff
@@ -0,0 +1,103 @@
...
+		try {
+			Class clz = obj.getClass();
+			final Field field = clz.getDeclaredField("mViews");
+			field.setAccessible(true);
+			return Collections.unmodifiableList(Arrays.asList((View[]) field.get(obj)));
+		} catch (Exception e) {
+			e.printStackTrace();
+		}
+
+		return null;
```
jasta: FWIW, I would actually prefer a checked exception over null, _especially_ because your version checking suggests that these errors should only happen in bizarre cases where the running OS doesn't match the code you pulled from aosp...

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/window/WindowFetcher.java
```diff
@@ -0,0 +1,103 @@
...
+				mViewsFiled.setAccessible(false);
+			} catch (Exception e) {
+				e.printStackTrace();
+			}
+		}
+	}
+
+	private static Object getOuter(Object innerWM) {
+		try {
+			Field parentFiled = innerWM.getClass().getDeclaredField("mWindowManager");
```
jasta: `parentField`

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/window/WindowFetcher.java
jasta: `sWindowDecorViews`, perhaps?  I'm starting to get pretty lost what all these variables and functions do :\

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/window/WindowFetcher.java
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * 7.0.0_r1     WindowManagerGlobal   private final ArrayList<View> mViews
+ *
+ * 8.0.0_r4    WindowManagerGlobal   private final ArrayList<View> mViews
+ */
+
+public class WindowFetcher {
+	public static List<View> getWindowViews(Context context) {
+		Util.throwIfNull(context);
+		if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.JELLY_BEAN) {
```
jasta: This class especially I think would benefit from a bit more of a structured API and pattern.  One thing I fine works super well is writing your compat layers explicitly to look like this:

```java
  abstract class WindowManagerDecorViewCompat {
    public ... get() {
      if (versionA) {
        sInstance = new VersionAImpl();
      } else if (versionB) {
        sInstance = new VersionBImpl();
      } else { 
        ...
      }
      return sInstance;
    } 

   public abstract List<View> getDecorViews();

    private static class VersionAImpl extends WindowManagerDecorViewCompat {
      @Override
      public List<View> getDecorViews() {
        ...versionAStrategy...
      }
    }

    ...
  }
```

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/RealWindowDescriptor.java
```diff
@@ -0,0 +1,34 @@
...
+import com.facebook.stetho.inspector.elements.android.window.WindowWrapper;
+
+import java.util.List;
+
+import static com.facebook.stetho.inspector.elements.android.ApplicationDescriptor.belongToActivity;
+
+
+public class RealWindowDescriptor extends AbstractChainedDescriptor<WindowWrapper> {
+	static {
+		WindowFetcher.hook();
```
jasta: hook is already a concept in `AbstractChainedDescriptor`.  Strongly recommend avoiding using that name for something that's got a different lifecycle/meaning.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/RealWindowDescriptor.java
```diff
@@ -0,0 +1,34 @@
...
+import android.view.View;
+
+import com.facebook.stetho.common.Accumulator;
+import com.facebook.stetho.inspector.elements.AbstractChainedDescriptor;
+import com.facebook.stetho.inspector.elements.android.window.WindowFetcher;
+import com.facebook.stetho.inspector.elements.android.window.WindowWrapper;
+
+import java.util.List;
+
+import static com.facebook.stetho.inspector.elements.android.ApplicationDescriptor.belongToActivity;
```
jasta: This strategy often makes code way harder to review/read since I have to jump up to the imports to see that belongToActivity wasn't reimplementing in this file scope.  Consider what it is like reviewing code without the IDE :)

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/RealWindowDescriptor.java
```diff
@@ -0,0 +1,34 @@
...
+
+
+public class RealWindowDescriptor extends AbstractChainedDescriptor<WindowWrapper> {
+	static {
+		WindowFetcher.hook();
+	}
+
+	@Override
+	protected void onGetChildren(WindowWrapper element, Accumulator<Object> children) {
+		List<View> windows = WindowFetcher.getWindowViews(element.getView().getContext());
```
jasta: This code block seems to be duplicated exactly as-is from ApplicationDescriptor; why?  Can it be re-used?  Why is the logic even needed to run in two separate contexts?

jasta:
First off, thank you so much for this change, it's been a long time coming!  I do think there's some quality stuff that needs attention but overall it should be easy to push this over the line and get it landed!

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/RealWindowDescriptor.java
android-notes: this class is useless, I have deleted

android-notes:
thank you for review my code, I have solved all the problems @jasta

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/window/WindowRootViewCompactV18Impl.java
```diff
@@ -0,0 +1,48 @@
...
+			throw new RuntimeException("unfortunately, you cannot view the view tree of the dialog etc.", e);
+		}
+
+	}
+
+	@NonNull
+	@Override
+	public List<View> getRootViews() {
+		try {
+			return Collections.unmodifiableList(Arrays.asList((View[]) mViewsField.get(mWindowManagerGlobal)));
```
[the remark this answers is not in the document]
android-notes: because before api 19,when add or remove window,WindowManager will create a new mViews array,and copy origin views to this new array.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/window/WindowRootViewCompactV16Impl.java
```diff
@@ -0,0 +1,57 @@
...
+	WindowRootViewCompactV16Impl(Context context) {
+		this.mContext = context;
+	}
+
+	@NonNull
+	@Override
+	public List<View> getRootViews() {
+		WindowManager windowManager = (WindowManager) mContext.getSystemService(Context.WINDOW_SERVICE);
+		if (windowManager == null) {
+			throw new RuntimeException("unfortunately, you cannot view the view tree of the dialog etc. windowManager == null !");
```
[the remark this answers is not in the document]
android-notes: should I remove this if statement?

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/window/WindowRootViewCompactV16Impl.java
```diff
@@ -0,0 +1,57 @@
...
+
+	private static Object getOuter(Object innerWM) {
+		try {
+			Field parentField = innerWM.getClass().getDeclaredField("mWindowManager");
+			parentField.setAccessible(true);
+			Object outerWM = parentField.get(innerWM);
+			parentField.setAccessible(false);
+			return outerWM;
+		} catch (NoSuchFieldException e) {
+			throw new RuntimeException("unfortunately, you cannot view the view tree of the dialog etc.", e);
```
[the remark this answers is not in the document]
android-notes: Should I change to throw new RuntimeException(e)?

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/window/WindowRootViewCompactV16Impl.java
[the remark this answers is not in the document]
jasta: Well, I can't really tell why you added it in the first place, but yes I generally would feel comfortable proceeding assuming that the windowManager is not null.  I can't fathom what it could possibly mean for that to be the case...

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/window/WindowRootViewCompactV16Impl.java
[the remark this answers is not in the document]
jasta: Yes.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/window/WindowRootViewCompactV18Impl.java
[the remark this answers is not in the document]
jasta: I don't think you understood my concern.  You're accessing the "root views" via WindowRootViewCompat.get(...).getRootViews(), but this will cause WindowRootViewCompat to be pinned in memory as a singleton.  If each call to getRootViews returns the same list as before, then for, say, API 19, the system will never be able to get a new list of root views after the first list is provided.

This should be corrected two-fold:

1. Each call to `getRootViews` should get the current list of root views.
2. The behaviour of each `...Impl` class should be updated to reflect this and ensure that it has this property to the best degree each `...Impl` can guarantee it.

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/window/WindowRootViewCompactV16Impl.java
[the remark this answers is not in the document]
android-notes: Because  there is a warning in android studio   Argument 'windowManager' might be null

stetho/src/main/java/com/facebook/stetho/inspector/elements/android/window/WindowRootViewCompactV18Impl.java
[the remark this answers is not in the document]
android-notes: API19,When adding or removing windows, WindowManager adds or removes views directly from WindowManagerGlobal's private final ArrayList <View> mViews instead of creating a new arraylist each time.


example  api23 
```java
package android.view;

public final class WindowManagerGlobal {

	...

	private final ArrayList<View> mViews = new ArrayList<View>();

       public static WindowManagerGlobal getInstance() {
          ...
               return sDefaultWindowManager;
      
       }
	public void addView(View view, ViewGroup.LayoutParams params,
           Display display, Window parentWindow) {

           ...
           mViews.add(view);
           ...

    }

}

```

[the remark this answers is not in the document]
jasta: In that case, you at least need to consistently make a copy of the array each time getRootViews is called in each impl.  Right now, at least v19 behaves differently by pretending to return an immutable list but in actuality it is edited under the hood without warning.

If you make sure a new copy is made in each call I think we're fine here.

android-notes:
Thank you very much for review my code, I have solved all the problems @jasta

jasta:
Great work thanks!

Issue #652, opened 2019-04-09
App crashes during work


Manwtein:
Hello, i'm trying to use the library and my browser shows me a list of my requests, but when i hide some fragments my app crashes.
My app takes the following steps. Open Activity and Fragment -> Open DialogFragment + send request -> Open nested DialogFragment(into childFragmentManager) -> Close nested DialogFragment -> Close DialogFragment(back to Activity) and i get an error message.
 I'm trying to figure out why this doesn't work, but I don't know why one of the children of ElementInfo is null. And I see the same problem in the past, sorry if there is a solution, but I cannot find it.
And i see the same problem in past, sorry if there is solution and but i can't find him.

java.lang.NullPointerException: Attempt to read from field 'java.lang.Object com.facebook.stetho.inspector.elements.ElementInfo.parentElement' on a null object reference
        at com.facebook.stetho.inspector.elements.ShadowDocument$Update.removeGarbageSubTree(ShadowDocument.java:378)
        at com.facebook.stetho.inspector.elements.ShadowDocument$Update.removeGarbageSubTree(ShadowDocument.java:386)
        at com.facebook.stetho.inspector.elements.ShadowDocument$Update.removeGarbageSubTree(ShadowDocument.java:386)
        at com.facebook.stetho.inspector.elements.ShadowDocument$Update.removeGarbageSubTree(ShadowDocument.java:386)
        at com.facebook.stetho.inspector.elements.ShadowDocument$Update.commit(ShadowDocument.java:360)
        at com.facebook.stetho.inspector.elements.Document.applyDocumentUpdate(Document.java:494)
        at com.facebook.stetho.inspector.elements.Document.updateTree(Document.java:346)
        at com.facebook.stetho.inspector.elements.Document.access$1200(Document.java:31)
        at com.facebook.stetho.inspector.elements.Document$ProviderListener.onPossiblyChanged(Document.java:720)
        at com.facebook.stetho.inspector.elements.android.AndroidDocumentProvider$1.run(AndroidDocumentProvider.java:71)
        at android.os.Handler.handleCallback(Handler.java:751)
        at android.os.Handler.dispatchMessage(Handler.java:95)
        at android.os.Looper.loop(Looper.java:154)
        at android.app.ActivityThread.main(ActivityThread.java:6119)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886)
[...]

PR #655, opened 2019-05-15
Add null checks before using element info

mathroule:
Add null checks before using element info to fix issue https://github.com/facebook/stetho/issues/652.

### DIFF
--- a/stetho/src/main/java/com/facebook/stetho/inspector/elements/ShadowDocument.java
+++ b/stetho/src/main/java/com/facebook/stetho/inspector/elements/ShadowDocument.java
@@ -373,15 +373,17 @@ public final class ShadowDocument implements DocumentView {
       // changes have been applied and after our caller (removeGarbageSubTree) removed another
       // element that claims this element as its child, then that means this element should not be
       // removed. It has been reparented, and recursion stops here.
-      if (elementInfo.parentElement != null &&
+      if (elementInfo != null && elementInfo.parentElement != null &&
           elementToInfoMap.containsKey(elementInfo.parentElement)) {
         return;
       }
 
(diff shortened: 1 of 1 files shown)


stetho/src/main/java/com/facebook/stetho/inspector/elements/ShadowDocument.java
```diff
@@ -373,15 +373,17 @@ private void removeGarbageSubTree(
       // changes have been applied and after our caller (removeGarbageSubTree) removed another
       // element that claims this element as its child, then that means this element should not be
       // removed. It has been reparented, and recursion stops here.
-      if (elementInfo.parentElement != null &&
+      if (elementInfo != null && elementInfo.parentElement != null &&
```
piotrek1543: What about  wrapping below code into `elementInfo?.let { }`? I see that It will reduce number of `elementInfo != null`  expressions from two to one

GaidamakUA:
Please, accept this pull request. I can use stetho at all without these checks.

growupzyg:
so,now which version should i use to avoid this issues? the newest version is 1.5.1 still has this problem

ajdevy:
Here's the fix:   
```
    implementation 'com.github.mathroule.stetho:stetho:bugfix~issue-652-SNAPSHOT'
    implementation("com.facebook.stetho:stetho-okhttp3:1.5.1") {
        exclude group: 'com.facebook.stetho'
    }
```
and add jitpack to your repositories:
```
buildscript {
    repositories {
        maven {
            url 'https://maven.fabric.io/public'
        }
    }
}
```

gengjiawen:
cc @jasta

Issue #648, opened 2019-03-20
Android P + Stetho 1.5.1 - View highlighting not working


Erkks:
Samsung S8, Android 9.

When using the same device with Android 8 and Stetho 1.5.0, whenever I selected/hovered over an element in the inspector, the corresponding element was shown on my app.

I needed to upgrade the Stetho build to 1.5.1 to prevent an app crash when expanding the elements in the inspector (https://github.com/facebook/stetho/issues/607), but unfortunately I've lost view highlighting.

Please let me know if I can provide any more information.

erkkipupil:
Confirmed that view highlighting works with Stetho 1.5.1 and Android 7 - issue seems to be Android P specific

beczesz:
For me it is still not working I'm using a Pixel3 Android 9 Stetho 1.5.1

erkkipupil:
Is this project still active? @jasta? Would be incredibly useful to have this merged in

IAmDarush:
View highlighting not working on Stetho 1.5.1. Any solutions?

PR #656, opened 2019-06-20
Fixing view highlighting on Android P

shashachu:
Fixes https://github.com/facebook/stetho/issues/648

`clipOutRect` wasn't the proper API to use in this case, so changed to use `save` and `restore`.

Based on this SO answer: https://stackoverflow.com/questions/50231950/what-is-the-best-alternative-to-canvas-cliprect-with-region-op-replace/50247323#50247323


shashachu:
@jasta is this library still being maintained? Would appreciate this PR getting merged so we can stop using a locally-built stetho version.

obs182:
Can this be merged please?

PR #665, opened 2019-08-21
Use AndroidX

cristan:
This enables people turning of jetifier when using Stetho.

### DIFF
--- a/stetho/build.gradle
+++ b/stetho/build.gradle
@@ -16,9 +16,9 @@ android {
 dependencies {
     implementation 'commons-cli:commons-cli:1.2'
     implementation 'com.google.code.findbugs:jsr305:2.0.1'
-    implementation 'com.android.support:support-annotations:28.0.0'
+    implementation 'androidx.annotation:annotation:1.1.0'
 
-    implementation 'com.android.support:appcompat-v7:23.0.1' // optional
+    implementation 'androidx.appcompat:appcompat:1.0.2' // optional
 
     testImplementation 'junit:junit:4.12'
     testImplementation('org.robolectric:robolectric:2.4') {
--- a/stetho/build.gradle
+++ b/stetho/build.gradle
@@ -15,7 +15,7 @@ android {
 
 dependencies {
     implementation 'commons-cli:commons-cli:1.2'
-    implementation 'com.google.code.findbugs:jsr305:2.0.1'
+    implementation 'com.google.code.findbugs:jsr305:2.0.3'
     implementation 'androidx.annotation:annotation:1.1.0'
 
     implementation 'androidx.appcompat:appcompat:1.0.2' // optional
@@ -25,8 +25,8 @@ dependencies {
         exclude module: 'commons-logging'
         exclude module: 'httpclient'
     }
-    testImplementation 'org.powermock:powermock-api-mockito:1.6.1'
-    testImplementation 'org.powermock:powermock-module-junit4:1.6.1'
+    testImplementation 'org.powermock:powermock-api-mockito:1.6.6'
+    testImplementation 'org.powermock:powermock-module-junit4:1.6.6'
 }
 
 apply from: rootProject.file('release.gradle')
--- a/stetho/build.gradle
+++ b/stetho/build.gradle
@@ -4,7 +4,7 @@ android {
     compileSdkVersion rootProject.ext.compileSdkVersion
 
     defaultConfig {
-        minSdkVersion 9
+        minSdkVersion 14
         targetSdkVersion rootProject.ext.targetSdkVersion
         versionCode 1
         versionName "1.0"
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.1.4'
+        classpath 'com.android.tools.build:gradle:3.2.1'
     }
 }
 
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.4-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.2-all.zip
--- a/stetho-js-rhino/README.md
+++ b/stetho-js-rhino/README.md
@@ -7,7 +7,7 @@ This [Stetho](https://facebook.github.io/stetho) plugin adds a JavaScript consol
 ### Download
 Download [the latest JARs](https://github.com/facebook/stetho/releases/latest) or grab via Gradle:
 ```groovy
-compile 'com.facebook.stetho:stetho-js-rhino:1.4.2'
+implementation 'com.facebook.stetho:stetho-js-rhino:1.4.2'
 ```
 or Maven:
 ```xml
--- a/stetho-js-rhino/build.gradle
+++ b/stetho-js-rhino/build.gradle
@@ -19,7 +19,7 @@ android {
 
 dependencies {
     implementation project(':stetho')
-    implementation 'com.google.code.findbugs:jsr305:2.0.1'
+    implementation 'com.google.code.findbugs:jsr305:2.0.3'
     implementation 'org.mozilla:rhino:1.7.6'
     implementation 'com.android.support:support-annotations:28.0.0'
 
--- a/stetho-js-rhino/build.gradle
+++ b/stetho-js-rhino/build.gradle
@@ -4,7 +4,7 @@ android {
     compileSdkVersion rootProject.ext.compileSdkVersion
 
     defaultConfig {
-        minSdkVersion 9
+        minSdkVersion 14
         targetSdkVersion rootProject.ext.targetSdkVersion
         versionCode 1
         versionName "1.0"
--- a/stetho-okhttp/build.gradle
+++ b/stetho-okhttp/build.gradle
@@ -21,7 +21,7 @@ android {
 
 dependencies {
     implementation project(':stetho')
-    implementation 'com.google.code.findbugs:jsr305:2.0.1'
+    implementation 'com.google.code.findbugs:jsr305:2.0.3'
     implementation 'com.squareup.okhttp:okhttp:2.7.2'
 
     testImplementation 'junit:junit:4.12'
@@ -29,12 +29,12 @@ dependencies {
         exclude module: 'commons-logging'
         exclude module: 'httpclient'
     }
-    testImplementation 'org.powermock:powermock-api-mockito:1.6.1'
-    testImplementation 'org.powermock:powermock-module-junit4:1.6.1'
+    testImplementation 'org.powermock:powermock-api-mockito:1.6.6'
+    testImplementation 'org.powermock:powermock-module-junit4:1.6.6'
 
     // Needed for Robolectric and PowerMock to be combined in a single test.
-    testImplementation 'org.powermock:powermock-module-junit4-rule:1.6.1'
-    testImplementation 'org.powermock:powermock-classloading-xstream:1.6.1'
+    testImplementation 'org.powermock:powermock-module-junit4-rule:1.6.6'
+    testImplementation 'org.powermock:powermock-classloading-xstream:1.6.6'
 
     testImplementation 'com.squareup.okhttp:mockwebserver:2.7.2'
 }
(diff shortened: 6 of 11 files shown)


stetho/build.gradle
```diff
@@ -15,18 +15,19 @@ android {
 
 dependencies {
     implementation 'commons-cli:commons-cli:1.2'
-    implementation 'com.google.code.findbugs:jsr305:2.0.1'
-    implementation 'com.android.support:support-annotations:28.0.0'
+    implementation 'com.google.code.findbugs:jsr305:2.0.3'
+    implementation 'androidx.annotation:annotation:1.1.0'
 
-    implementation 'com.android.support:appcompat-v7:23.0.1' // optional
+    implementation 'androidx.appcompat:appcompat:1.0.2' // optional
```
consp1racy: If it's optional it should be `compileOnly`.
AppCompat is actually used nowhere in the code. Only Core and Fragment.

trevjonez: compileOnly is probably a no go? https://issuetracker.google.com/issues/109894265

consp1racy: That should only be an issue if you're trying to work with Android resources from the library, which stetho doesn't, at least as far as I can see. As long as the error doesn't pop up during compilation we're golden.

trevjonez: the resources still contribute to the `R.class` file generation which "could" lead to issues for consumers maybe at build time maybe at runtime maybe in a lint run? 

I agree it is probably not anything to worry about, but I am more making the argument that if we know it isn't correct it would be best to just avoid it and star the aosp issue to hopefully move it up in the priority list.

consp1racy: Oh. That. Yeah, that could be an issue for some unfortunate souls.

Defensively, call it a breaking change and raise major version. If in exchange I get rid of hard dependency of AppCompat, I'd consider it an acceptable price.

cristan: I'm really not a fan of a dependency which isn't actually used, `compileOnly` or otherwise. That's why I've removed the AppCompat dependency: the Fragment dependency is enough for everything AndroidX/AppCompat related.

The Fragment dependency used to have `// optional` next to it. AndroidX's Fragment is actually used (in `FragmentCompatFramework` and in `FragmentCompatSupportLib`), so I've removed this notice.

Let me know if this is to your liking :)

BTW: you might want to bump the major version in any case: because the `minSdkVersion` has been bumped to 14.

consp1racy: Nice, but! If we're raising the major version, we're good to go all the way.

We can replace the fragment dependency with `compileOnly` (on v1.0.0 to be safe). The compat fragment accessor is only useful if we already have compat/AndroidX fragments in classpath. We don't need to have an implementation dependency on it. But we still want the references at compile time, hence `compileOnly`. This is what it's for.

I didn't see this line in changes, looks important:
https://github.com/facebook/stetho/blob/98e2b6680161c58334b0d69396aeb17c7c144371/stetho/src/main/java/com/facebook/stetho/common/android/FragmentCompat.java#L45-L48

Since we have a compile-time dependency on support fragments which may not actually be there at run-time, we don't need reflection. Just a try-catch:

    static {
        boolean hasSupportFragment = false;
        try {
            androidx.fragment.app.Fragment.class;
            hasSupportFragment = true;
        } catch (NoClassDefFoundError ignore) {
            // AndroidX Fragment is not in classpath.
        }
        sHasSupportFragment = hasSupportFragment;
    }

All other places with similar calls would follow suit.

There are more reflection calls, some of them may even be obsolete with AndroidX. Worth checking.

cristan: There was actually quite a bit of code which was obsolete because of `minSdkVersion` of 14, so I've removed that :)

There were also some functions/parameters overridden which now have nullability annotations. I don't want to add even more lint warnings, so I've added the missing annotations to the best of my ability. However, I don't see when to use `javax.annotation.Nullable` and when to use `androidx.annotation.Nullable` (and their non-null variants), so if there's a reason when to choose one over the other, I'd like to know.

About the optional dependencies: you're right, I totally missed the reflection part. I've added a comment to explain the optional dependency. I've also removed `ReflectionUtil.tryGetClassForName`: we now use consp1racy's code.

I guess we'll have to wait for support for `compileOnly` dependencies, but this also  issue applies to your master, so I don't think this PR has to wait for that.

consp1racy: javax annotations for Java libraries (jar only) that may be used outside of Android ecosystem.
androidx annotations for Android libraries (jar or aar).
Considering this project is aimed exclusively at Android I'd go with androidx annotations.

Hold on, my code will only work if Fragment dependency is `compileOnly`. Well, it will always return true unless the consumer manually excludes our Fragment `implementation` dependency.

Remind me, why can't we use `compileOnly` with Fragment dependency?

consp1racy: I seem to recall that `compileOnly` worked with non-transitive dependencies on Android libraries. On Android libraries that don't depend on other Android libraries.

This should work:

    compileOnly 'androidx.fragment:fragment:1.0.0@aar'

`@aar` resolves to just the Fragment aar artifact, dropping its own dependencies. That way they don't clash with our other `implementation` dependencies e.g. on AndroidX Core.

cristan: I've tried `compileOnly 'androidx.fragment:fragment:1.0.0@aar'`, but when I try to actually run Stetho, I get `error: cannot find symbol` everywhere. I guess the reason it doesn't work is the aforementioned https://issuetracker.google.com/issues/109894265

consp1racy: Can be, we need AndroidX Core, which until now was pulled in automatically as a dependency of Fragment. This should work:

    implementation 'androidx.core:core:1.1.0'
    compileOnly 'androidx.fragment:fragment:1.1.0'

> I get `error: cannot find symbol` everywhere

The "everywhere" part wasn't helpful at all. Knowing which imports didn't resolve (and maybe seeing that all were from androidx.core package) would be.

Stop assuming failure, this is going to work ;)

cristan: Do we actually need AndroidX core? The only places with `import androidx.core` are `AccessibilityUtil` and `AccessibilityNodeInfoWrapper`. `AccessibilityUtil` is only used in `AccessibilityNodeInfoWrapper` is only used from `ViewDescriptor` where we check whether we can access `androidx.core.view.accessibility.AccessibilityNodeInfoCompat.class.getClass()`. So, as far as I can see, core is just as optional as the Fragment dependency. Is there anywhere else we use core that I missed?

> The "everywhere" part wasn't helpful at all
Good point: it actually isn't everywhere. It's everywhere where we use any transient dependencies like core.

Adding all the transient dependencies works:
```compileOnly 'androidx.core:core:1.1.0@aar'
compileOnly 'androidx.lifecycle:lifecycle-common:2.1.0'
compileOnly 'androidx.lifecycle:lifecycle-viewmodel:2.1.0@aar'
compileOnly 'androidx.savedstate:savedstate:1.0.0@aar'
compileOnly 'androidx.activity:activity:1.0.0@aar'
compileOnly 'androidx.fragment:fragment:1.1.0@aar'```

Not using `@aar` also gets rid of the errors:
`compileOnly 'androidx.fragment:fragment:1.1.0`

What shall we use?

consp1racy: I don't think we need all the dependencies of Fragment. We only need those that we need to compile, whose API we use in our code. Fragment checks out. Core, as you mentioned, too.

If the accessibility stuff is optional, `compileOnly` is fine (with its own `ReflectionUtil` check). Otherwise `implementation` is the way to go. So this is it?

    compileOnly 'androidx.core:core:1.0.0@aar'
    compileOnly 'androidx.fragment:fragment:1.0.0@aar'

Mind the `1.0.0` as we don't want to accidentally access something that's not in v1. That way the consumer is free to choose anything from [1.0.0, 2.0.0).

Plus implementation annotations, of course.

Awesome!

---

I'd add proguard rules to ignore missing Fragment and Core.

    -dontwarn androidx.fragment.**
    -dontnote androidx.fragment.**
    -dontwarn androidx.core.**
    -dontnote androidx.core.**

Otherwise consumers using Proguard would get fails, if they don't use Fragment or Core, about missing classes.

It would be better to name concrete classes or entry points that are referenced in code, but this should be fine.

Or the other way around, specify the accessing classes in this project, that should work. So many possibilities!

---

A potential issue is that the whole module is free to reference classes from Core and Fragment but it's no longer guaranteed to be in runtime classpath. These "optional" accessors should be in separate modules, which won't leak their `compileOnly` dependency to the main Stetho library.

That touches publishing, which may be a problem.

cristan: Just using Core and Fragment as `@aar` dependencies doesn't work:
```compileOnly 'androidx.core:core:1.0.0@aar'
compileOnly 'androidx.fragment:fragment:1.0.0@aar'```

You'll get the following errors:

```/Users/Cristan/repos/stetho/stetho/src/main/java/com/facebook/stetho/common/android/FragmentCompatSupportLib.java:72: error: cannot access LifecycleOwner
      return fragment.getFragmentManager();
                     ^
  class file for androidx.lifecycle.LifecycleOwner not found
/Users/Cristan/repos/stetho/stetho/src/main/java/com/facebook/stetho/common/android/FragmentCompatSupportLib.java:77: error: cannot access ViewModelStoreOwner
      return fragment.getResources();```

If you want AndroidX Fragment 1.0.0 as an AAR, you'll need to define the following to solve this:

```compileOnly 'androidx.lifecycle:lifecycle-common:2.0.0'
compileOnly 'androidx.lifecycle:lifecycle-viewmodel:2.0.0@aar'
compileOnly 'androidx.core:core:1.0.0@aar'
compileOnly 'androidx.fragment:fragment:1.0.0@aar'```

Anyway: making the optional dependencies actually optional hasn't got anything to do with AndroidX: your master branch has the same issues. I've therefore just updated the comment.

I suggest to merge this PR and create a separate issue to make the dependency actually optional (or go the non-optional route and remove the realtime checks).

cristan: I've just removed the last code which still didn't assume a `minSdkLevel` of 14. I've removed the documentation about `ActivityTracker.add` and `remove`: the AutomaticTracker now works for everyone, so I don't think we should even mention that you can fiddle with the added / removed activities. You could even argue for making `add` and `remove` private methods.

cristan:
> Compile SDK version should be raised to 29 if it's not already. That may result in even more nullability annotations.

- `compileSdkVersion` and `targetSdkVersion` are now at 29
- Missing non-null annotations are added
- Redundant casts at findViewById removed (which were no longer needed after targeting 26)

consp1racy:
I didn't find the `compileOnly` dependencies in changes. As you mentioned this should be it:

    compileOnly 'androidx.lifecycle:lifecycle-common:2.0.0'
    compileOnly 'androidx.lifecycle:lifecycle-viewmodel:2.0.0@aar'
    compileOnly 'androidx.core:core:1.0.0@aar'
    compileOnly 'androidx.fragment:fragment:1.0.0@aar'

I found this instead:

    implementation 'androidx.fragment:fragment:1.0.0@aar'

cristan:
> I didn't find the `compileOnly` dependencies in changes. As you mentioned this should be it:
> 
> ```
> compileOnly 'androidx.lifecycle:lifecycle-common:2.0.0'
> compileOnly 'androidx.lifecycle:lifecycle-viewmodel:2.0.0@aar'
> compileOnly 'androidx.core:core:1.0.0@aar'
> compileOnly 'androidx.fragment:fragment:1.0.0@aar'
> ```
> 
> I found this instead:
> 
> ```
> implementation 'androidx.fragment:fragment:1.0.0@aar'
> ```
@consp1racy That's on purpose: I don't think this should be part of this PR anymore. That's because making the optional dependencies actually optional hasn't got anything to do with AndroidX. Therefore, let's just merge this PR and create a separate issue to make the dependency actually optional (or go the non-optional route and remove the realtime checks).

pandelisgreen13:
There is any update for next release with AndroidX?

erikghonyan:
@cristan Thanks for getting the ball rolling! Are you still looking into this?

cristan:
> @cristan Thanks for getting the ball rolling! Are you still looking into this?
I don't think there's anything for me to do: all AndroidX related suggestions are implemented.

Unfortunately, this project seems to be dead: only 1 commit has been done on this project since March 2019. I'm now using another solution, I suggest y'all do the same.

lmj0011:
>I'm now using another solution, I suggest y'all do the same.

I assume that solution is [Flipper](https://github.com/facebook/flipper#--flipper-)?

cristan:
> > I'm now using another solution, I suggest y'all do the same.
> 
> I assume that solution is [Flipper](https://github.com/facebook/flipper#--flipper-)?

It isn't. I've tried it, ran into [this issue](https://github.com/facebook/flipper/issues/326) and almost 2 years later it still isn't fixed. I now use [chucker](https://github.com/ChuckerTeam/chucker).

scottyab:
Thanks for the work here @cristan.

PR #721, opened 2024-10-25
Fix the issue that CDT contents are disordered

yinyinnie:
1. Fix the issue that CDT contents are disordered, new WEBKIT_REV is from https://chromedevtools.github.io/devtools-protocol/
2. Fix the issue that latest chrome(119 or 120) can not dicover Stetho


starlitya:
Hi @yinyinnie 

I tried this PR—the UI works fine in the latest Chrome version, but the JS console appears to be broken and I can’t evaluate any scripts. Have you encountered the same problem?

